PR #52, opened 2014-09-08
Documentation: frontend state.

mf59816:
(This is just a basic email import.)


xi:
Closing now. It is still in the github histroy in case we need it.

PR #538, opened 2015-01-22
2015 01 cc sass text mixins overhaul

local-girl:
Replaced text mixin with bunch of individual text mixins than can be overwritten in the theme


local-girl:
I am going to try this again in a different way

PR #580, opened 2015-01-26
2015 01 jk enable cache mode without proxy

joka:
Enable 'without_proxy_cache' caching mode:

- Resources: cache in browser 0, must revalidate based on etags:

       changed_backrefs, changed_descendants, modification_date, userid, hidden/deleted

- Assets: cache in browser for 30 Min, must revalidate


nidico:
The `rest_api.rst` doctest fails.

PR #623, opened 2015-01-31
Lock rate button during post

nidico:
This avoids errors on "double-click". Also, post errors should show error message and trigger a reload of the current rates.


xi:
replaced by #644.

PR #642, opened 2015-02-03
2015 01 jk refactore autoupdate subscriber

joka:


nidico:
This works, I haven't done an extensive code review. From a quick glance, it looks good though I think the autoupdate mechanism won't work anymore for forkables - see comment in 4afb505. Am I wrong?

joka:
> This works, I haven't done an extensive code review. From a quick glance, it looks good though I think the autoupdate mechanism won't work anymore for forkables - see comment in 4afb505. Am I wrong?

yes, but we never had tests for autoupdate with forkables anyway. We can merge this.

nidico:
Okay. I'm not too glad removing functionality which was already there, but we can look at the old code once we need it.

PR #643, opened 2015-02-04
No hover for disabled buttons

local-girl:


xi:
The normal (not `m-call-to-action`) button still has the problem.

local-girl:
Fixedup in  5b8b659

xi:
LGTM.

PR #644, opened 2015-02-04
refactor rate

xi:
this refactors the frontend rate module and contains a reimplementation of #623. It also adds the like button at the bottom of the mercator proposal detail page.


nidico:
LGTM. However, I don't receive updates from the websocket server on cast, although the item is registered.

nidico:
For the MercatorProposal two rate directives synchronisation, we also need a soution for the situation in which the current user hasn't posted a rate *item* yet. In this case there are are no websocket registered; if an item is finally posted through one rate directive is pressed the other rate directive doesn't notice. We probably want some local event handling.

xi:
I solved the problem by adding a local `adhRateEventHandler` service that syncs the directives locally.

nidico:
Works nicely!

PR #646, opened 2015-02-04
2015 02 cs singleton rate

### LINKED ISSUE #632: Backend should refuse multiple IRates with same subject and object
The backend doesn't prevent clients from creating multiple `IRate` resources targetting the same `objects` and `subjects`. This allows to rate the same object multiple times.

ChristianSi:
Fixes #632: we now ensure (a) that there is just one rate per subject/object combination and (b) that that nobody can submit rates for others.

### DIFF
--- a/src/adhocracy_core/docs/rest_api.rst
+++ b/src/adhocracy_core/docs/rest_api.rst
@@ -993,6 +993,77 @@ this comment version::
     True
 
 
+Rates
+-----
+
+We can rate objects that provide the `adhocracy_core.sheets.rate.IRateable`
+sheet (or a subclass of it), e.g. comment versions. Rateables have their own
+post pool, so we ask the comment where to send rates about it::
+
+    >>> resp = testapp.get(snd_commvers_path)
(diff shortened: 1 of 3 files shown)


src/adhocracy_core/docs/rest_api.rst
```diff
@@ -993,6 +993,77 @@ this comment version::
...
+    >>> ratevers['data']['adhocracy_core.sheets.versions.IVersionable']['follows'] = [third_ratevers_path]
+    >>> ratevers['root_versions'] = [third_ratevers_path]
+    >>> resp_data = testapp.post_json(rate_path, ratevers, headers=god_header,
+    ...                               status=400).json
+    >>> resp_data['errors'][0]['name']
+    'data.adhocracy_core.sheets.rate.IRate.subject'
+    >>> resp_data['errors'][0]['description']
+    'Must be the currently logged-in user'
+
+
```
joka: This big doctest file is almost impossible to debug.
We should start using "normal" python  tests for functional backend testing and use the doctest only for generic concepts how to work with rest api.

PR #648, opened 2015-02-05
2015 01 jk make acceptance tests more stable

joka:
The acceptance tests run stable now on my old laptop.
But there are many failing tests in test_comments. Is sombody working on this?


nidico:
I haven't had much issues lately - apart from `test_comments`, which regularly fails (though non-deterministically). I'll look into that tomorrow. The jasmine unit tests also occasionally fail when run through splinter.

So I think this PR doesn't change much, but it might catch some webdriver issues.

joka:
@nidico comments tests should be stable now: a1efa9323cf76a6133d4b212880ae5808bc21ce6

nidico:
Neat! The last two tests still fail for me though, both in phantomjs as well as in firefox, also with `--splinter-speed=1`.

PR #654, opened 2015-02-07
Fix: 'hasattr' might swallow exceptions

joka:
Python build in hasattr might hide Attribute and zodb exceptions:

http://stackoverflow.com/questions/24971061/python-hasattr-vs-getattr
https://dev.plone.org/ticket/3285
http://mail.dzug.org/mailman/archives/zope/2006-October/002010.html

To be on the safe side we should just not use it.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/authentication/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/authentication/__init__.py
@@ -35,9 +35,11 @@ class TokenMangerAnnotationStorage:
 
     @property
     def token_to_user_id_timestamp(self):
-        if not hasattr(self.context, self.annotation_key):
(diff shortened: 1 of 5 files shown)


ChristianSi:
I think the bad behavior of `hasattr` was fixed by Python 3. In [Py2](https://docs.python.org/2/library/functions.html?highlight=hasattr#hasattr), it catches *all* exceptions thrown by `getattr` which is of course pretty bad. But in [Py 3](https://docs.python.org/3/library/functions.html?highlight=hasattr#hasattr), it only catches `AttributeError ` and thus should be safe to use.

So `hasattr` should be safe for us to use, but of course it cannot hurt using `getattr` with a default value instead -- in fact it's a bit cleaner, since it avoids raising and catching an exception.

PR #655, opened 2015-02-08
Add protractor, port most acceptance tests

nidico:
As we don't get our acceptance tests fully stabilized, I started playing with [protractor](https://angular.github.io/protractor/#/), the Angular "end to end" testing framework and ported most of the splinter acceptance tests. I'm pretty positive about the results - things mostly "just work", hardly a need for `wait` calls. Also, performance is much better locally, however it's still using a browser, so it's still slow....

This isn't fully complete yet, because:

- [x] Needs to be better integrated with our build setup / package structure.
- [x] Needs some test structuring concept.
- [x] The splinter acceptance tests still need to be removed.
- [ ] The MercatorProposal acceptance tests do not work on travis/saucelabs until next version of protractor (see commit "Skip Mercator tests..."). We could still merge this though.
- [ ] The MercatorProposal filter tests aren't ported yet, as I didn't port the MercatorProposal fixture to node yet. We can still merge, as the splinter mercator tests are still there.

Notes:

- Backend and frontend test ports are now `9080` and `9090` instead of `6542` and `6552`.
- This now builds on #686 which should be merged first
- This now uses chromedriver and needs an installed chrome or chromium browser - the chromedriver is installed through buildout.
    - PhantomJS cannot be used in `directConnect` mode and would need a running selenium server. The protractor folks advise [against using PhantomJS](https://angular.github.io/protractor/#/browser-setup).
    - Firefox webdriver is currently broken (with Firefox 35) - don't try it.
- This isn't run together with `py.test_run_all` as it's not integrated in pytest.
- Acceptance tests don't use TypeScript at the moment. I doubt it would be worth it. If we would do so, this would work fine - protractor understands `.map.js` files.
- I don't fully understand protractor locators (or they have a bug) - this is why the test suite spits out *more than one element found for locator* warnings and selects the wrong elements (but still works).

Try it:

    bin/buildout
    bin/protractor etc/protractorConf.js


nidico:
Ready for review. I'd be glad if this could be merged very soon, even if not everything is polished and final.

etc/protractorConf.js
```diff
@@ -0,0 +1,27 @@
...
+        "browserName": "chrome"
+    },
+    beforeLaunch: function() {
+        exec("bin/supervisord");
+        exec("bin/supervisorctl restart adhocracy_test:test_zeo test_backend_with_ws adhocracy_test:test_autobahn adhocracy_test:test_frontend");
+    },
+    afterLaunch: function() {
+        exec("bin/supervisorctl stop adhocracy_test:test_zeo test_backend_with_ws adhocracy_test:test_autobahn adhocracy_test:test_frontend");
+        exec("rm -rf var/test_zeodata/Data.fs* var/test_zeodata/blobs");
+    },
```
xi: This means that the database will not be purged between tests, right?

etc/protractorConfSauce.js
```diff
@@ -0,0 +1,33 @@
+var exec = require("sync-exec");
+var pr = process.env.TRAVIS_PULL_REQUEST;
+var name = ((pr === "false") ? "" : "#" + pr + " ") + process.env.TRAVIS_COMMIT;
+
+exports.config = {
+    suites: {
+        core: "../src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/*Spec.js"
```
xi: Why is mercator not included here?

xi: Oh, I see:

> The MercatorProposal acceptance tests do not work on travis/saucelabs until next version of protractor (see commit "Skip Mercator tests..."). We could still merge this though.

etc/protractorConf.js
nidico: Yes, we need to take care of that in the test construction. Should be okay though in my opinion.

xi:
This needs chrome >= 38.0.2125.0. The latest version in debian wheezy is 37.0.2062.120. But as debian jessie is about to be released, this should not be a reason not to merge this.

PR #659, opened 2015-02-09
Make alert messages translateable

nidico:
Introduce technical strings with a `TR__` prefix in order to be able to extract them automatically. Extraction isn't done in that PR though.


xi:
LGTM.

Issue #577, opened 2015-01-26
Close all overlays etc on column close or mode change


nidico:
If a column is closed, all overlays of that particular column shall be closed as well. All other behavior would be confusing.

local-girl:
Not sure if I want the overlays to close. Say I am in the middle of reporting something and I want to reference another proposal. I open the proposal list and my report disappears. We'd need a "Navigate away" warning also ?

xi:
> Say I am in the middle of reporting something and I want to reference another proposal.

I think collapsing is meant for that case. But we should probably ask the designers about that.

xi:
Fixed in #660

PR #660, opened 2015-02-09
Close overlays on navigation

xi:
this is my attempt on fixing #577. I added ways to `clear()` a moving column, i.e. hide all overlays and alerts as well as the sidebar.

I am not too happy with the interface because it exposes this functionality 3 times: As a function on the controller, via DOM event and via angular event.

What is worse, it does not even work in all cases. While I tried to cover most of what we currently use, there are still exceptions. E.g. show the sidebar on the first column and then click "add proposal". The sidebar will still be visible on the form.

The first question that I had to think about was *when exactly* we want the moving columns to be cleared. I found that this is when they are hidden, but not when they are collapsed. We do know if column got closed via the `movingColumn` key in `adhTopLevelState` and we also know which directive represents which column in `adhMovingColumns`. So this case is covered.

Unfortunately, there is a second, much harder case. We also want to clear a column when we exchange its contents. This happends for example when switching from listing to create form (first column) or from detail to edit (second column). We keep the same column element in order to get smooth transitions.

In the case of the second column, I mitigated the problem by registering a `click` handler on the "edit" button. This is not possible in the case of the first column because the "add proposal" button does not have any relation to the column where it triggers a change.

I think this pull request covers many cases and is a big improvement. So I propose merging it. But it is not sufficient to call #577 "fixed".

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/MovingColumns/MovingColumns.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/MovingColumns/MovingColumns.ts
@@ -119,12 +119,22 @@ export var movingColumns = (
                 }, 1);
             };
 
+            var triggerClearEvents = (newCls : string, oldCls : string) : void => {
+                var columns = element.find(".moving-column");
+                for (var i = 0; i < columns.length; i++) {
+                    if (newCls.split("-")[i + 1] === "hide" && oldCls.split("-")[i + 1] !== "hide") {
+                        columns.eq(i).trigger("adhMovingColumn.clear");
+                    }
+                }
+            };
+
             $($window).resize(resizeNoTransition);
 
             var move = (newCls) => {
                 if (newCls !== cls) {
                     element.removeClass(cls);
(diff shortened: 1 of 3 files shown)


nidico:
Entering the user space throws an error:

    Cannot read property 'split' of undefined
        at exports.movingColumns.link.triggerClearEvents (MovingColumns.ts:125)

nidico:
And it's 2015 :)

nidico:
Some cases aren't covered yet:

- Filter list, then "add proposal"
- Open report / share, then change proposal

The former could be tackled by adding `clear()` to the constructor of the abstract column class. The latter can be handled manually locally.

xi:
I came up with a new idea on how to do this that covers all currently known cases. I reverted all previous commits except for the first one and implemented the new idea after that.

The idea is to trigger a `clear()` whenever a column directive (the one inside of `adhMovingColumn`) is linked or an important variable changes. For example, if the `proposalUrl` changes, the proposal column should be cleared.

This works quite well and is not that ugly code.

nidico:
Reasonable approach, works well.

PR #661, opened 2015-02-09
Extract TR__ strings

xi:
this continues #659 in that it uses technical "TR__"-strings everywhere we have translation in typescript code. It also adds these strings to the extract_messages script.


nidico:
The `TR__GUEST` is missing in the translation files.

PR #664, opened 2015-02-09
Add mercator_budget index view

ChristianSi:
This adds a new mercator_budget index to the backend which allows filtering by the budget field. Values are same as with the mercator_requested_funding index, except that there is an additional value "above_50000" since this field has no upper limit.

CC @local-girl who wants to add the frontend part.


joka:
looks ok and tests pass

local-girl:
Seems to still validate 50000 or less on the server ?

PR #673, opened 2015-02-11
2015 02 cs purge varnish

ChristianSi:
Add after-commit hook to send PURGE requests to Varnish whenever resources have been changed by a transaction. The hook will only be added if an 'adhocracy.varnish_url' is configured in the ini-file.

I haven't tested this with a real Varnish yet so it's possible that further changes will be needed.

The Varnish config needs to be changed as documented at https://www.varnish-cache.org/docs/3.0/tutorial/purging.html#http-purges to allow PURGE requests.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/__init__.py
@@ -31,7 +31,7 @@ def root_factory(request):
 
 
 def add_after_commit_hooks(request):
-    """Add after commit hook to notify the websocket server."""
+    """Add after commit hooks."""
     # FIXME this is a quick hack
     from adhocracy_core.websockets.client import \
         send_messages_after_commit_hook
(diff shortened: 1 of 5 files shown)


src/adhocracy_core/adhocracy_core/__init__.py
```diff
@@ -51,8 +56,6 @@ def includeme(config):
     config.include('pyramid_zodbconn')
     config.include('pyramid_exclog')
     config.include('pyramid_mako')
-    config.hook_zca()  # enable global adapter lookup (used by adhocracy.utils)
-    authz_policy = RoleACLAuthorizationPolicy()
     config.hook_zca()  # global adapter lookup (used by adhocracy_core.utils)
```
joka: Why did you remove this?

src/adhocracy_core/adhocracy_core/__init__.py
```diff
@@ -39,6 +39,11 @@ def add_after_commit_hooks(request):
         clear_transaction_changelog_after_commit_hook
     current_transaction = transaction.get()
     registry = request.registry
+    # Order matters here
+    if 'adhocracy.varnish_url' in registry.settings:
+        from adhocracy_core.caching import purge_varnish_after_commit_hook
```
joka: If you do the  the registry  check here instead inside the commit hook, 
you need to look at two different packages to understand whats going on.

Referening to our styleguide we should prefere separating exception handling from actual programm code. 
But i guess single responsibility is more important.

src/adhocracy_core/adhocracy_core/__init__.py
ChristianSi: It's duplicate (the next two lines are the same).

src/adhocracy_core/adhocracy_core/__init__.py
ChristianSi: Good point, I changed that.

PR #675, opened 2015-02-11
implement is-selected state on commentable sections

xi:
this implements downlighting compatible states on mercator proposal commentable sections. Note that CSS is not part of this pull request.

@local-girl: CSS seems to require some restructuring. Can you please take care of that?

image


local-girl:
Continued in #683

PR #676, opened 2015-02-11
rename eventHandler to eventManager

xi:
The generic eventManager was introduced a long time ago to allow service to expose an interface similar to jquery's on/off/trigger.  For some reason, it was called "eventHandler". But this term is already used for callbacks that are registered fir an event. So I propose to rename it to "eventManager".


local-girl:
Looks good except the acceptance tests ? Not sure if they are relevant.

xi:
acceptance tests look like they are not related to this.

PR #677, opened 2015-02-11
2015 02 cc filters translations

local-girl:
Added filter translation strings. Realised I had copied typo MERACTOR so fixed it site wide


xi:
I made some additions. Could you please review and merge?

PR #678, opened 2015-02-11
http busy draft

xi:
This implements an idea @nidico and me had: A service that hooks into `$http` and counts how many requests are currently in progress. It might be useful for debugging. This is a request for comments: Do you think this might be useful, and if so, how should I procede?


nidico:
In my opinion this is very useful - not only for development, but also for usage. We should ask the designers on whether / how that should be shown to the users (I propose a rotating icon in the top header...).

nidico:
Rebased due to conflict with master.

nidico:
As I really want to have that information in master during development, I decided to merge it as it is and just not show the information in debug mode. We can improve from here and make something for normal users if we want.

PR #684, opened 2015-02-12
2015 02 cc busy ui

local-girl:
Added UI for the busy bar. Sorry the commits are a bit messy I dont know when lint failures commit and when then don't !


nidico:
Looks nice. Some remarks:

- We should not only show this in debug mode.
- I'd propose to still show the counter (in addition) in debug mode, as it's useful debugging information for us.
- A smaller animated gif should be sufficient, i.e. `12px x 4px`, no?

local-girl:
Decreasing the size of the animated gif will hardly decrease its filesize, they aren't much fun to edit but that would be a designers job. I didn't realise about debug mode I will update this now.

local-girl:
Done

xi:
LGTM.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_busy.scss
```diff
@@ -0,0 +1,18 @@
...
+    height: 4px;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+
+    &.ng-hide-add.ng-hide-add-active,
+    &.ng-hide-remove.ng-hide-remove-active {
+        transition: 0.5s linear all;
+    }
```
xi: I think you don't need the extra selectors here. Just add the transition to `.busy-bar`.

PR #686, opened 2015-02-15
Cache eggs and node_modules in travis

nidico:
This is possible by using container-based builds (without sudo). We
might be able to cache even more, but this already speeds up the travis
setup time significantly (from 8 to 2 minutes until the first test
runs).

See:

- http://docs.travis-ci.com/user/workers/container-based-infrastructure/
- http://docs.travis-ci.com/user/caching/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python:
   - "3.4"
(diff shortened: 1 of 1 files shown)


ChristianSi:
Great!

PR #689, opened 2015-02-15
Drop support for IE9

### LINKED ISSUE #486: On IE9 redirect from startpage to /#/r/mercator is not working
instead of redirecting to /#/r/mercator is not working it redirects to /#/undefined which gives an 404 error

### LINKED ISSUE #487: Content on https://a3-dev.liqd.net is not loaded when opened with IE9 - I just see a blank page
Error in Dveloper tools says: SCRIPT5: Access is denied 

text.js, line 269 character 13

Same Error is described here: 
http://stackoverflow.com/questions/20081872/cannot-get-the-text-plugin-to-work-with-requirejs-in-ie

Something related could be find here:
http://stackoverflow.com/questions/20624476/xdomainrequest-cors-for-xml-causing-access-is-denied-error-in-ie8-ie9

### LINKED ISSUE #592: Embed fails in IE9
This might be related to #486 (redirect to the `/#/` history manipulation fallback doesn't work) and to #66, as I'm also getting the following errors when trying to access `embed.html#!/#/r/mercator`:

    [$rootScope:infdig] 10 $digest() iterations reached. Aborting.

Furthermore, we might want to clean up the link of the embedding page (i.e. skip the `/#` from the embedded page).

### LINKED ISSUE #607: On IE9 I can't edit a proposal that has no image
Some kind of server error is shown, but it has no text in it.

nidico:
IE9 has multiple open issues (see below, but also no support for websockets and more). It has already caused a lot of hassle in the past, and it will do so in the future.

We don't have hard support requirements for IE9, so let's drop it.

- closes #487 (won't fix)
- closes #592 (won't fix)
- closes #486 (won't fix)
- closes #607 (won't fix)

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/root.html.mako
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/root.html.mako
@@ -9,7 +9,7 @@
         <base href="/" />
     </head>
     <body>
-        <!--[if lt IE 9]>
+        <!--[if lt IE 10]>
         <div class="unsupported-browser">
             <img src="/static/icons/png/attention.png"/>
             <div>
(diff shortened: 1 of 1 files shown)


local-girl:
I can update the banner ?

PR #690, opened 2015-02-15
Add custom MercatorProposal listitem widget

nidico:
This improves initial load significantly (from 38s to 11s with 66 proposals and a couple of comments on my development setup).

We still do 7 HTTP requests per item:

- 3 can be loaded immediately (main resource, comments & rate queries)
- 3 subresources can be loaded once the main resource is there
- the image even needs to wait for the introduction subresource, i.e. requires 3 roundtrips after the proposal version path is known.

This doesn't reuse code from the main `MercatorProposal` resource widget as this would create effort.

Possible further improvements:

- Move `title` from `introduction` subresource to the main resource in order to display title earlier. The current modeling is an error anyway IMO.
- Loading all proposals in one query instead of separately could be done with the query API, but wouldn't help too much.
- A big improvement would be an extension of the GET API, which allows to specify which referenced resources shall also be retrieved. This would make the first suggestion obsolete.

### DIFF
--- a/src/mercator/static/js/Packages/MercatorProposal/Listing.html
+++ b/src/mercator/static/js/Packages/MercatorProposal/Listing.html
@@ -8,13 +8,8 @@
     data-no-create-form="true"
     data-empty-text="{{ 'TR__PROPOSAL_EMPTY_TEXT' | translate }}">
     <div data-ng-switch="transclusionId">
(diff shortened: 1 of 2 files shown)


xi:
LGTM. Test failure looks unrelated.

PR #691, opened 2015-02-15
Avoid websocket subscription of non-resources (fixes #682)

### LINKED ISSUE #682: Cache tries to subscribe websocket to non-resource paths
The frontend cache caches all `GET` and `OPTIONS` requests and tries to subscribe a websocket to the given path.

However this fails if the request doesn't refer to a resource, but to an API endpoint such as `/message_user` and leads to an error message.

I think it's okay to cache such responses, but simply don't register add a websocket in that case.

nidico:


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Cache.ts
```diff
@@ -90,9 +97,12 @@ export class Service {
     private getOrSetCached(path : string) : IHttpCacheItem {
         var cached = this.cache.get(path);
         if (typeof cached === "undefined") {
-            var wshandle = this.adhWebSocket.register(path, (msg) => {
-                this.invalidate(path);
-            });
+            var wshandle;
+            if (!(_.contains(this.nonResourceUrls, path))) {
```
xi: To many parantheses.

nidico: Fixed in 0601a26.

xi:
LGTM

PR #698, opened 2015-02-16
Hide password in log file

### LINKED ISSUE #681: Backend logs full failed login attempt
We should definitely not log the password in plain text in failed password attempts. Logging the login name is okay in my opinion, as it allows to pursue abuse.

ChristianSi:
Fixes #681.

A "password" field at the top level of invalid JSON requests will be rendered as  `<hidden>`.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/rest/views.py
+++ b/src/adhocracy_core/adhocracy_core/rest/views.py
@@ -237,26 +237,27 @@ def _show_request_body(request: Request) -> str:
     """
     Show the request body.
 
-    In case of JSON requests with a "password" field at the top level,
-    the contents of the password field will be hidden.
-
     In case of multipart/form-data requests (file upload), only the 120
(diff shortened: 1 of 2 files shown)


src/adhocracy_core/adhocracy_core/rest/views.py
```diff
@@ -236,12 +237,26 @@ def _show_request_body(request: Request) -> str:
...
+    the contents of the password field will be hidden.
+
     In case of multipart/form-data requests (file upload), only the 120
     first characters of the body are shown.
     """
     result = request.body
     if request.content_type == 'multipart/form-data' and len(result) > 120:
         result = '{}...'.format(result[:120])
+    elif request.content_type == 'application/json':
+        # Hide password, if it occurs as a top-level field
```
joka: This comment just mirros the docstring nearby, so we dont need it

PR #700, opened 2015-02-16
Use ng-submit instead of ng-click on submit button (fixes #503)

### LINKED ISSUE #503: form is submitted twice on return key
I noticed while working on the messaging frontend that the ng-click handler seems to be executed twice if I submit a form with the return key rather than clicking. I am not yet sure what the actual symptoms are.

xi:
According to https://docs.angularjs.org/api/ng/directive/form, both methods should work the same. For me, `ng-click` results in a double-post when using the return key to submit.


nidico:
This breaks some protractor tests.

PR #701, opened 2015-02-16
Comment deletion

local-girl:
Hi there, this doesn't seem to have the desired effect for me, can you take a look please?

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/CommentDetail.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/CommentDetail.html
@@ -28,6 +28,10 @@
                         <i class="icon-ban-circle"></i>
                         {{ "TR__REPORT" | translate }}
                     </a>
+                    <a href="#" data-ng-click="delete()" data-ng-if="report">
+                        <i class="icon-x"></i>
+                        {{ "TR__DELETE" | translate }}
+                    </a>
                 </span>
                 <adh-rate data-refers-to="{{data.path}}"></adh-rate>
             </div>
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/CommentDetail.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/CommentDetail.html
@@ -71,7 +71,8 @@
                 data-path="{{comment}}"
                 data-mode="display"
                 data-frontend-order-predicate="frontendOrderPredicate"
-                data-frontend-order-reverse="frontendOrderReverse">
+                data-frontend-order-reverse="frontendOrderReverse"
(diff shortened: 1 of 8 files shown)


ChristianSi:
@local-girl You think it's a backend issue?

ChristianSi:
@local-girl There is a problem with the PUT request, it looks like this:

{"content_type":"adhocracy_core.resources.comment.IComment","data":{"adhocracy_core.sheets.metadata.IMetadata":{}}}

So the "hidden": "true" attribute is missing.

ChristianSi:
CC @xi who has also worked on the deletion frontend.

local-girl:
this is my put request:
```
return this.adhHttp.put(path, {
            content_type: "adhocracy_core.resources.comment.IComment",
            data: {
                "adhocracy_core.sheets.metadata.IMetadata": {
                    hidden: true
                }
            }
        });
```

xi:
I found the issue: Before we actually send the request, we sanitize the data (`exportContent()`). In that function, there is the following code:

```
// workaround, as normal users can not set `hidden` field
// FIXME: use more appropriate place, e.g. expose in meta api
if (sheetName === "adhocracy_core.sheets.metadata.IMetadata" && fieldName === "hidden") {
    delete sheet[fieldName];
}
```

I am not sure whether this is still needed. An alternative would be to use `adhHttp.postRaw()` which does not do the sanitation. @nidico what do you think?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/CommentDetail.html
```diff
@@ -24,6 +24,10 @@
                         <i class="icon-ban-circle"></i>
                         {{ "TR__REPORT" | translate }}
                     </a>
+                    <a href="#" data-ng-click="delete()" data-ng-if="report">
```
nidico: This should not be `ng-if="report"`, but a proper check whether we are allowed to delete.

nidico:
Yeah, well, this is still needed in order to make edit work. I regret that we ever introduced that, knowing that delete would break!

The proper solution would be getting the information from some OPTIONS response and remove the ugly sanitation hack.

ChristianSi:
@nidico:

> The proper solution would be getting the information from some OPTIONS response and remove the ugly sanitation hack.

I would assume that that information is used already to decide whether to show the delete button in the first place. No need to check it twice, so allowing the delete button to turn off the sanitation seems reasonable enough.

local-girl:
All yours @nidico  !

nidico:
Deleting comments works, but still requires a reload after deletion. Also, an error is thrown if another user replies to the deleted comment in the meantime.

Deleting proposals does not work yet.

nidico:
Ready for review.

MagdaN:
Works for me!

PR #708, opened 2015-02-17
Warmup cache for listings

nidico:
This extends `adhHttp.get` by an optional `warmupCache` parameter.

The optional parameter `warmupPoolCache` can be used on pool queries
in order to requests the IPool sheet to contain a list of resources
instead of only paths. The returned promise however has the same API
as if `warmupPoolCache` would be skipped.

This is then used in the listing widget. It saves one request per listitem,
but more importantly reduces the number of roundtrips by one for all
subresources and images.

On my local development setup, there isn't a huge impact (4s load instead
of 5s, which is at least something). However the impact may be more
relevant if there is actual network latency.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Cache.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Cache.ts
@@ -32,6 +32,7 @@ export class Service {
     private nonResourceUrls;
 
     constructor(
+        private $q : ng.IQService,
         private adhConfig : AdhConfig.IService,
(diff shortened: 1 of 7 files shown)


xi:
I did no understand all details, but looks ok to me. Shouldn't we activate this for *all* poll requests? If it is cheap on the backend, it sound like a good idea to me. At the least we should also activate this for comment trees and user listings.

nidico:
User listings are already warmed up, because they're just listings. Comments don't use pool queries at the moment, so we should treat them seperately.

xi:
Looks ok. But did no real testing.

PR #711, opened 2015-02-17
Fix lodash regression (fixes #705)

### LINKED ISSUE #705: Subresource comment pages empty
MercatorProposal subresource comment pages such as http://localhost:6551/r/mercator/ie8715/VERSION_0000004/@comments:introduction yield an empty page.

xi:
since lodash 3.0.0, `forEach()` is lazy. That means that the function is not executed until you call `value()` on the return value.

Apart from fixing #705, this should also fix yet unreported issues in batchRequests and cache invalidation.


nidico:
This doesn't compile :(

nidico:
The change in lodash only applies when using chaining (see commit messages). `_.forEach()` is still evaluated eagerly.

PR #712, opened 2015-02-17
Work on copy so that list does not get reversed twice (fixes #627)

### LINKED ISSUE #627: Strange sorting in proposal list
When I support/unsupport a proposal, the proposal list is resorted. But I do not understand the algorithm by which it is sorted. Further investigation required.

MagdaN:
Sometimes the list is loaded (for example if support button is clicked) from cache and should not reversed. So the final sorting is done on a copy.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
@@ -148,9 +148,9 @@ export class Listing<Container extends ResourcesBase.Resource> {
                         $scope.container = container;
                         $scope.poolPath = _self.containerAdapter.poolPath($scope.container);
                         // FIXME: Sorting direction should be implemented in backend, working on a copy is used,
-                        // because otherwise sometimes the already reversed sorted list (from cache) would be 
+                        // because otherwise sometimes the already reversed sorted list (from cache) would be
                         // reversed again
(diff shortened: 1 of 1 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
```diff
@@ -147,7 +147,10 @@ export class Listing<Container extends ResourcesBase.Resource> {
                     return adhHttp.get($scope.path, params).then((container) => {
                         $scope.container = container;
                         $scope.poolPath = _self.containerAdapter.poolPath($scope.container);
-                        $scope.elements = _self.containerAdapter.elemRefs($scope.container);
+                        // FIXME: Sorting direction should be implemented in backend, working on a copy is used,
+                        // because otherwise sometimes the already reversed sorted list (from cache) would be 
+                        // reversed again
+                        $scope.elements = angular.copy(_self.containerAdapter.elemRefs($scope.container));
```
xi: IMHO angular should not be used directly (because it is not available in unit tests). Either use dependency injection or lodash.

nidico: I'd use lodash.

nidico:
LGTM. I just added the lodash import. (This also works without this, because _ is also in the global scope, but we should not rely on this)

PR #713, opened 2015-02-17
2015 02 cs checks session tokens

### LINKED ISSUE #693: Allow to check session token validity
Frontends or other services (e.g. other policycompass services) want a simple way to figure out whether a user path and token combination is valid.

We currently do an `OPTIONS` request to `user_path`, which is pretty much a hack.

I'd suggest to

- either add an endpoint `check_token` or alike,
- or reject any request with incorrect headers.

If the headers are checked on every request anyway, I'd go for the second option, as this seems to be the correct way.

That way we could really just do any request, e.g. `GET user_path` for session validation.

ChristianSi:
Fixes #693.

If a request contains "X-User-..." headers, they will now always be validated, even if they aren't rquired for authorization.


nidico:
LGTM!

PR #717, opened 2015-02-17
Added user name to user message form. Closes #702

### LINKED ISSUE #702: Replace "Username" in UserMessage.html
In `src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserMessage.html` the string "Username" needs to be replaced by the actual username. It should probably be included in the translatable string via interpolsation.

```
<!-- FIXME: replace "Username" -->
<p>{{ "TR__MESSAGE_TO" | translate }} Username</p>
```

local-girl:
I dont know if this is the right way to pass this data, I fear not !


xi:
I changed to so that we get the recipient resource locally and extract the name from there.

PR #728, opened 2015-02-18
Rafactor rate adapter

xi:
In #644 I noticed that there was only a single adapter for rate and like. I split it into one adapter for each. This is a fixup which splits a second function I did not notice before.


nidico:
LGTM.

PR #731, opened 2015-02-18
add some trivial type signatures to http cache

xi:


nidico:
LGTM. Test failure seems unrelated.

PR #732, opened 2015-02-18
Fix middle button (fixes #726)

### LINKED ISSUE #726: Middle-click broken
Users are used to work with tabs. There's no technical problem preventing users from doing that (apart from the need of loading each page separately).

However, middle click is hijacked by custom click events, which set the location to the local (iframe) URL instead of the canonical URL, which is in `href`.

xi:
See https://developer.mozilla.org/en-US/docs/Web/Events/click


nidico:
LGTM. Test failures look unrelated.

PR #736, opened 2015-02-18
2015 02 cs fix rates index

### LINKED ISSUE #714: Sorting proposals by rate does not work properly
Check the listing in https://plone-a3-stage.liqd.net/en/project-space#!/r/mercator/ - there are quite a view proposals with ratings. All the frontend does is reverting the sorting direction, so I consider this a backend  issue?

ChristianSi:
Fixes #714.

The "rates" index now only count rates with the LAST tag.


nidico:
LGTM.

Note that old data still has the wrong index set. Using `bin/sd_reindex etc/development.ini` hasn't resolved this for me.

PR #740, opened 2015-02-18
fancier 404 page

local-girl:
Possibly need a german graphic ?


nidico:
I think the image should be explicit `` instead of background and the `{{code}} - {{message|translate}}` part in the `alt` tag.

nidico:
I could live without a German version of the graphics now, as this is only shown exceptionally. We can improve that later on.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/templates/Error.html
```diff
@@ -1 +1,4 @@
-<h1>{{code}} - {{message | translate}}</h1>
+<div class="error-page error-page-{{code}}" data-ng-switch="code">
+    <img data-ng-switch-when="404" src="/static/images/404.png" style="margin: 0 auto" alt="{{code}} - {{message | translate}}" />
```
xi: style should not be in the template

src/mercator/static/stylesheets/scss/a3.scss
```diff
@@ -4,6 +4,7 @@
 
 @import "compass/css3";
 @import "compass/css3/inline-block";
+@import "compass/typography/text/replacement";
```
xi: No longer needed.

xi:
I also added some default HTTP error messages.

PR #746, opened 2015-02-18
Link to terms and conditions from registration

nidico:


local-girl:
Link goes nowhere ?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Register.html
```diff
@@ -50,7 +50,7 @@
         </span>
         <div class="register-check">
             <input type="checkbox" name="registerCheck" data-ng-model="input.registerCheck" required="required">
-            {{ "TR__I_ACCEPT_TERMS_AND_CONDITIONS" | translate }}
+            {{ "TR__I_ACCEPT_THE" | translate }} <a href="{{ termsUrl }}" target="_blank">{{ "TR__TERMS_AND_CONDITIONS" | translate }}</a>
```
ChristianSi: That seems a very brittle way to handle translations.

nidico: I thought it's better than putting html in the translation string, which then needs to be marked as `safe` and thus allow translaters to break the site. I don't have a better idea how to do this at the moment.

nidico:
Well, link goes to whatever is defined in `etc/frontend_development.ini` in the `adhocracy.frontend.terms_url` setting. I don't think it makes sense to include some link by default. Well, we could do example.org, but that's no better than what we have now...

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Register.html
xi: This is a genaral problem which we do not have a solution for yet.

xi: I created #752 for this.

xi:
LGTM

PR #747, opened 2015-02-18
2015 02 jk enable varnish purging and with proxy cache header

joka:
Add varnish to adhocracy_core buildout

Force gzip compression for json files with varnish

Make cache mode "with_proxy_cache" work (backend behind varnish reverse proxy) 
   
    To activate the cache mode behind varnish proyx server set the following
    config options:
    
    * adhocracy_core.caching.http.mode = with_proxy_cache
    * adhocracy.varnish_url = <varnish url>
    * adhocracy.frontend.rest_url = <varnish url>
    
    Then start varnish and restart the backend server (with supervisor).

NEEDS TESTING on Staging. Especially regarding possible security issues (stale content somehow?, login/logout? different user with different permissions?)


joka:
@slomo Varnish does the gzip compression now.  In contrast to nginx it makes the etag "weak" instead of removing it.

nidico:
> In contrast to nginx it makes the etag "weak" instead of removing it.

Great. Then we should also use it in front of the frontend resources.

Btw, according to the [README](https://github.com/nulab/nginx-gzip-filter-allow-weak-etag-patch/blob/master/README.md) of a related project, downgrading to weak etags is in nginx 1.7.x since [this commit](http://hg.nginx.org/nginx/rev/e491b26fa5a1).

joka:
> Btw, according to the README of a related project (a nginx plugin which also etag with gzip), downgrading to weak etags is in nginx 1.7.x since this commit

If we enable gzip we also have to add "accepts" (encoding") to the "vary" header. 
Varnish does this, nginx also?

PR #750, opened 2015-02-19
Fix: template mail variable missing with tests

pallix:
- Add adhocracy.message_user_subject variable to test.ini


pallix:
Please review

PR #753, opened 2015-02-20
Minify more javascript

nidico:


xi:
templates.js is already minified by `r.js`, the rest looks good. I also updated the minified file in my fork of sticky-kit, so we might want to use that, too.

PR #755, opened 2015-02-20
Fix session token check on load

nidico:
This does two things:

- It only enables the token *after* the validity check. Previously there
was a race condition due to the temporary token activation.
- It uses a better validity check: it calls HEAD on rest_url:/ instead
of calling POST on the user path. This is possible due to #713.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserSpec.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserSpec.ts
@@ -12,6 +12,7 @@ export var register = () => {
     describe("User", () => {
         describe("Service", () => {
             var adhUser;
+            var adhConfigMock;
             var adhHttpMock;
             var adhCacheMock;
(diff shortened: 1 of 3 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserSpec.ts
```diff
@@ -82,7 +88,9 @@ export var register = () => {
                 it("calls 'enableToken' if 'user-token' and 'user-path' exist in storage", () => {
                     windowMock.localStorage.getItem.and.returnValue("huhu");
                     fn();
-                    expect(adhUser.enableToken).toHaveBeenCalledWith("huhu", "huhu", true);
+                    _.defer(() => {
+                        expect(adhUser.enableToken).toHaveBeenCalledWith("huhu", "huhu");
+                    });
```
xi: async, so it needs a `done`

xi:
LGTM.

PR #756, opened 2015-02-20
eventManager priority

xi:
This adds an optional `priority` argument to `eventManager.on()`. Handlers with a higher priority are executed first.

This is meant to fix race conditions between HTTP cache invalidation and other webSocket message handlers. Unfortunately, it did not fix any visible issues.

I think this might be useful in some way, but does not fix anything yet. In addition, unittests are missing. So there might be significant bugs, e.g. the execution order might be reversed.

Another very relevant issue is that race conditions might still exist between different messages. For example, the server sends a `modified` message to a resource and a `modified_child` message to its parent. The order in which these are executed is not touched by this pull request.


xi:
Doesn't seem like this will be merged.

Issue #757, opened 2015-02-20
Normalize URLs


xi:
We use URLs to identify resources and we expect them to unambiguous. But we do not normalize them. So for example we might have `//example.com/adhocracy` in one place and `//example.com/adhocracy/` in another and `//example.com//adhocracy` in a third.

PR #754, opened 2015-02-20
Local Socket

xi:
this implements an eventManager that is compatible to adhWebSocket. It is meant to solve problems like #410. I originally implemented it to fix #663, but that turned out to be more or less unrelated.

This branch also contains some minor fixes to related code. So I would like to get this merged fast.

PR #758, opened 2015-02-20
Normalize URL in http cache (fixes #663)

### LINKED ISSUE #663: Proposal Listing not updated on proposal edit
After I edit a mercator proposal, the listing is not updated and still links to the old version.

xi:
after attempting fixes for #663 in #754 and #756 I finally found something that actually works: I noticed that there were multiple event handlers for the mercator platform pool because the URL was not normalized. This is merely a workaround. I created #757 for a more general solution.


nidico:
Unfortunately, this doesn't work for me.

nidico:
Can't be bad though.

PR #764, opened 2015-02-21
Hide registration form after success

nidico:
The previous behavior was very confusing. In particular, on small screens users didn't see the email hint at all.


xi:
Can not test because of some strange local bug. Did you talk to the designers about this?

nidico:
I talked to the designers. They're fine with it as long there still is a cancel link. I did that in ce864b8.

Issue #762, opened 2015-02-20
Unintended redirect to login on protected page


xi:
-   steps to reproduce
   -   log in
   -   go to proposal create page
   -   reload
-   expected result: stay on that page
-   actual result: redirect to login (without being logged out) (since #765 only sometimes)
-   possible solution: routing waits for token validation (see #755)

nidico:
This causes the mercator proposal splinter tests to fail.

PR #765, opened 2015-02-21
Atomar session loading from localstorage

nidico:
See commits.

This is all good stuff, and should fix the "logged out immediately after login" issue, but unfortunately doesn't remedy #762 - this is stilly racy.


nidico:
Merging myself, because urgent.

@xi: Please review later.

xi:
LGTM

PR #771, opened 2015-02-22
Check for session readiness before proposal routing

nidico:
This is a workaround for #762.

It's not very intrusive, as it only causes waiting for the session on protected resources. At the moment, this is only done for the MercatorProposal create resource manually. This should be done in a more generic way, e.g. in `TopLevelState`.


nidico:
Merging myself, as this stops the acceptance tests from (always) failing.

xi:
LGTM.

PR #772, opened 2015-02-22
Disable varnish by default in development setup

nidico:
#747 added varnish and builds it by default. However, most developers don't need varnish during development. Also, it slows down travis considerably.


nidico:
Merging myself.

@joka / @ChristianSi : please have a look.

ChristianSi:
Looks OK to me.

PR #775, opened 2015-02-22
Sort comments - newest first

nidico:
Above all, this orders comments in *any* useful way instead of leaving them unsorted.

Furthermore, it changes the comment metadata that way that it always shows the creation time, not the modification time. If a comment has been modified, a `(edited)` label is added.

This is all subject to change, but shall make the comment thread as it is a bit more coherent and make the sorting understandable to the users.


nidico:
I talked to the designers - they will come up with a cleaner solution, but we can merge this now, as it is already a big improvement.

PR #776, opened 2015-02-23
Unify supporters and comment count in proposal main widget and listitem

nidico:
This fixes incorrect supporters count in the listing item if a user has
more then one rate item per object item. (This may happen in rare cases,
e.g. if a user rates an old version after rating for a newer version)

Also this saves two requests and some code.


src/mercator/static/js/Packages/MercatorProposal/MercatorProposal.ts
```diff
@@ -234,6 +234,53 @@ export var uploadImageFile = (
...
+ * promise recursive comments count.
+ */
+var countComments = (adhHttp : AdhHttp.Service<any>, postPoolPath : string) : ng.IPromise<number> => {
+
+    var query : any = {};
+    query.content_type = RICommentVersion.content_type;
+    query.depth = "all";
+    query.tag = "LAST";
+    query.count = "true";
+    query.elements = "omit";
```
xi: Gardening: This should have the same format as above.

xi:
LGTM.

PR #778, opened 2015-02-23
2015 02 mn fix countrylist

MagdaN:
This could be merged to have the reduced 'European Country' list for mercator - but I would keep on working on a more general approach, which also enables translation for A3.


nidico:
Works for me. Let's merge that for now.

Still open:

- use global list by default; only overwrite country list in mercator package
- use country list from trusted source (i.e. UNO, I'd say) - http://www.countryareacode.net/ doesn't look very official...

PR #782, opened 2015-02-23
2015 02 mn update granted info

MagdaN:
The python accecptance tests seem to fail, but I cannot figure out why


MagdaN:
Oh I forgot: This fixes the problem that the  yes / no checkboxes for financial support granted in the proposal should only be forced to select something here if the user enters a value in the "other source of income" box

nidico:
The acceptance tests failed, because the granted field only shows up when something is written in `finance.other_sources`, which it is not in the acceptance test. Fixed in bf05d2c.

nidico:
LGTM!

PR #784, opened 2015-02-23
Fix warnings in the acceptance tests

pallix:


src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/EmbeddedCommentsPage.js
```diff
@@ -22,18 +22,7 @@ var EmbeddedCommentsPage = function(referer) {
...
-        /*
-        return all.reduce(function(acc, elem) {
-            return protractor.promise.all(
-                acc.getAttribute("data-path"),
-                elem.getAttribute("data-path")
-            ).then(function(paths) {
-                return (path[0] > path[1] ? acc : elem);
-            })
-        });*/
+        return this.listing.element(by.xpath("(//adh-comment-resource)[1]"));
```
nidico: I'd like to keep the FIXME, as this is assumed to be working in `RateSpec`. By the way, the `reduce` was my try to get the just created version, because we currently don't have that information visible in the frontend. But that can be removed.

Issue #770, opened 2015-02-21
All resources created in one transaction should have same creation_date


nidico:
If multiple resources are created in one transaction (e.g. using batch), they should alle have the same `creation_date`.

nidico:
Not fixed, see comment in #787.

nidico:
Really fixed through #807.

PR #787, opened 2015-02-23
set same creation/modification date for modified resources in transaction

joka:


joka:
@local-girl needs this

joka:
ups, one failing tests

ChristianSi:
The failing test is unrelated. I don't now how it came into master, but I fixed it in #790.

nidico:
This doesn't work for me: A batch created comment version still has a newer `creation_date` than the corresponding item, which has been created during the same request. Reopening #770.

Issue #752, opened 2015-02-20
translatable strings with HTML inside


xi:
Sometimes we want a translatable text with some HTML inside, e.g.

```
Please login <a href="/login">here</a>.
```

On the other hand, we do not want to have HTML in the translation for security reasons. The current workaround is to split the translatable string like this:

```
{{ "TR__PLEASE_LOGIN" | translate }} <a href="/login">{{ "TR__HERE" | translate }}</a>.
```

A more generic way could be

```
{{ "TR__PLEASE_LOGIN_HERE_BEFORE" | translate }}
<a href="/login">{{ "TR__PLEASE_LOGIN_HERE_INSIDE" | translate }}</a>
{{ "TR__PLEASE_LOGIN_HERE_AFTER" | translate }}.
```

Still this is a pain to translate. So we need a better way to handle these cases.

xi:
I searched stackoverflow and found two related (currently unanswered) questions:

-  https://stackoverflow.com/questions/25528332/angular-translate-how-to-translate-a-string-with-html-formatting
-  https://stackoverflow.com/questions/22668805/how-to-handle-genitive-in-angular-translate

and one answer listing some approaches:

-  https://stackoverflow.com/questions/15273465/coding-style-for-html-i18n-l10n-with-variables

xi:
This has been implemented in #789.

PR #789, opened 2015-02-23
translate strings with HTML

xi:
An attempt at fixing #752. Translations look like this:

```
I already have an account. Log in [[link:here]].
```

NOTE: I only put this into the Locale package because I did not want to create a new package. Feel free to nag.


nidico:
I like it, let's see how long it lasts...

PR #793, opened 2015-02-24
Order user list: newest first

nidico:
As long as we don't have other sorting mechanisms, this is at least some
sorting. It also makes sure that it's not always the same users on top
of the list.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
@@ -228,6 +228,8 @@ export var userListDirective = (adhUser : AdhUser.Service, adhConfig : AdhConfig
         templateUrl: adhConfig.pkg_path + pkgLocation + "/UserList.html",
         link: (scope) => {
             scope.user = adhUser;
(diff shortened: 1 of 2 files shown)


xi:
I think I would expect alphabetical sorting.

nidico:
Indeed. But we don't have the title at hand currently. This is just about sorting in *any* meaningful way.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
```diff
@@ -228,6 +228,8 @@ export var userListDirective = (adhUser : AdhUser.Service, adhConfig : AdhConfig
         templateUrl: adhConfig.pkg_path + pkgLocation + "/UserList.html",
         link: (scope) => {
             scope.user = adhUser;
+            scope["frontendOrderPredicate"] = (id) => id;
+            scope["frontendOrderReverse"] = true;
```
xi: I'd prefere `scope.frontendOrderPredicate` syntax.

nidico: done.

PR #795, opened 2015-02-24
Handle uploaded assets represented as BytesIO correctly

### LINKED ISSUE #687: Uploading images < 1000 bytes fails
Trying to upload an image with `< 1000 bytes` fails.

It seems that `value.file` passed to `FileStorageType.deserialize` seems to be a `_io.BytesIO` instead of a `_io.BufferedRandom` in that case. This doesn't have a `fileno()` method and breaks.

Example: 

good (1001 bytes):

good-1001-bytes

bad (997 bytes):

bad-997-bytes

ChristianSi:
Fixes #687.


nidico:
Works!

PR #799, opened 2015-02-24
Add an acceptance test for the spaces navigation

pallix:


xi:
I don't think that matching by directive name or by click handlers is a good idea in acceptance tests as those are implementation details. I would like it better to match by classes and HTML tags.

pallix:
It should be better now

xi:
LGTM.

PR #800, opened 2015-02-24
Fix event manager related memory leaks (fixes #674)

### LINKED ISSUE #674: Many callbacks are not unregistered
We register callbacks in many places and only sometimes unregister them later on. As an example, `adhTopLevelState` doesn't even expose the functionality to unregister callbacks registered with `on()`.

I am not sure how big this problem is. It may either cause bugs (callback is triggered but referenced context does no longer exist) or prevent large amounts of data from being garbage collected.

Some contexts in which this might be relevant:

-  `adhTopLevelState` (`on()` and ` bind()`)
-  `adhWebSocket`
-  any instance of `EventHandler`

Basically a directive should unregister all callbacks on `$destroy`.

Further investigation needed.

xi:
**This contains breaking changes to eventManager and all derived services!**


nidico:
This doesn't build currently.

nidico:
Very good!

PR #801, opened 2015-02-24
Add acceptance test for commenting on a proposal

pallix:


src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/EmbeddedCommentsPage.js
```diff
@@ -9,7 +9,7 @@ var EmbeddedCommentsPage = function(referer) {
     this.url = "/embed/create-or-show-comment-listing"
         + "?key=" + referer + "&pool-path=" + this.poolPath;
 
-    this.listing = element(by.css(".listing"));
+    this.listing = element(by.tagName("adh-comment-listing"));
     this.listingCreateForm = this.listing.element(by.css(".listing-create-form"));
```
xi: I don't think that matching by directive name or by click handlers is a good idea in acceptance tests as those are implementation details. I would like it better to match by classes and HTML tags.

xi:
test passes for me.

PR #802, opened 2015-02-24
different country labels for planned or already non-profit organizations

### LINKED ISSUE #798: Wrong field in form
There is a wrong field in the proposal form. When you select "I am part of a registered non profit e.g. NGO" you have to select a country where your registration is planned. This is wrong.

country_of_registration

xi:
fixes #798


MagdaN:
Worked for me - tests also did not fail (except for two that also failed in master)

Issue #804, opened 2015-02-24
Comment rate + edit broken


xi:
Steps to reproduce:

-  create a comment
-  rate that comment
-  edit that comment

Expected result:

-  rate counter shows 1

Actual result:

-  rate counter shows 0

xi:
might be connected to #803

nidico:
I think this is caused by two issues - one in the frontend and one in the backend:

- The frontend doesn't update a comment resource after editing - i.e. we still show the old comment path (and thus rating). This is because the comment resource (one item in the listing) is an *item*, not a *version*, and calling `$scope.update` on the listing doesn't cause the comment to update.

- The backend doesn't seem to invalidate `LAST` tags, i.e. after creating a new version, the backend returns a 304 when asking for `LAST` and thus still points to the old version.

xi:
> This is because the comment resource (one item in the listing) is an item, not a version

See #192 on why it is implemented like that.

nidico:
This isn't fixed yet - 53e8c8f fixed #820, a backend issue with similar symptoms.

joka:
@nidico should in investigate if this is a backend issue?

nidico:
> @nidico should I investigate if there is still a backend issue?

@joka: No, #804 is a frontend issue, which is supposed to be fixed with #838.

#820 is the backend issue, which you have already fixed.

Issue #803, opened 2015-02-24
Comment editing + reply brokenComment editing + reply


xi:
Steps to reproduce:

-  Create a comment
-  edit that comment
-  reply to that comment

Expected result:

-  Reply is displayed

Actual result:

-  Reply is not displayed

nidico:
Cannot reproduce on my dev machine.

nidico:
I think this is the frontend issue of #804.

pallix:
@nidico try reproducing manually on the dev version, not on the testing version. For some reasons I could also not reproduce it on the test version.

PR #805, opened 2015-02-24
Add acceptance test for #803 et #804

pallix:


src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/RateSpec.js
```diff
@@ -34,4 +34,19 @@ describe("ratings", function() {
...
+        var page = new EmbeddedCommentsPage("c2").get();
+        var comment = page.createComment("c4");
+        var rate = page.getRateWidget(comment);
+
+        rate.element(by.css(".rate-pro")).click();
+
+        var changedComment = page.editComment(comment, ["c4 - edited"]);
+        var rate2 = page.getRateWidget(changedComment);
+        expect(rate2.element(by.css(".rate-difference")).getText()).toEqual("+1");
+
```
xi: Gardening: I would prefer no newline here. But that's not really important.

xi:
Both tests fail as expected. They should be marked as `xit` (pending) so we can merge.

pallix:
Is the test for issue #804 failing for you? It works for me locally

xi:
I will check again

xi:
In any case tests on travis should not fail. So in either case it must be marked `xit`.

xi:
The test for #804 is failing for me when running `bin/protractor etc/protractorConf.js`

PR #807, opened 2015-02-24
fix: same creation date in transaction

joka:
* mv shared modifcation date for one transaction to registry
* renamed get_request_date to get_modification_date
* add after commit hook to reset shared modification_date


ChristianSi:
Looks good.

PR #816, opened 2015-02-25
2015 02  pa deterministic sharespec acceptance test

pallix:


nidico:
I haven't had non-deterministic behavior, but you certainly know better.

PR #817, opened 2015-02-25
Fix emails not being queued in the test config

pallix:
This works to send email/message in the test version (tested manually on port 9090). I'm not sure of the implication for the other tests  (unit tests using the pyramid_mailer.testing setup?)


nidico:
Doesn't seem to be a problem. Merging.

PR #818, opened 2015-02-25
Set locale if embedded

nidico:


nidico:
Merging myself, review pending.

PR #821, opened 2015-02-26
Better check for comment edited state

nidico:
This is possible due to the fix of #770.

Note that subcomments still show up as edited as soon as parent comments have been edited due to our updating model. This is not related to this though.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Adapter.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Adapter.ts
@@ -90,7 +90,9 @@ export class CommentAdapter extends ListingCommentableAdapter implements AdhComm
     }
 
     edited(resource : RICommentVersion) : boolean {
+        // NOTE: this is lexicographic comparison. Might break if the datetime
+        // encoding changes.
         var meta = resource.data[SIMetadata.nick];
-        return (meta.modification_date > meta.item_creation_date);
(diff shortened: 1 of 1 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Adapter.ts
```diff
@@ -94,6 +90,7 @@ export class CommentAdapter extends ListingCommentableAdapter implements AdhComm
     }
 
     edited(resource : RICommentVersion) : boolean {
-        return !(<any>_).endsWith(resource.path, "/VERSION_0000001/");
+        var meta = resource.data[SIMetadata.nick];
+        return (meta.modification_date > meta.item_creation_date);
```
xi: No braces necessary.

xi:
You are comparing the two dates as strings. I am not sure if you are aware. This should work ([JavaScript compares strings lexicographically](https://stackoverflow.com/questions/10863092/why-is-string-11-less-than-string-3) and properly parsing the dates would take some effort (we had issues with that before, see d6844c8f [also note that just wrapping the string with `new Date(...)` as was done before will yield wrong results]). So I suggest leaving it as is and adding a FIXME.

nidico:
I'm aware of the lexicographic comparison.

xi:
> Tried to edit a reply and got this error:

This is most likely not related to this pull request. My guess would be #803.

PR #822, opened 2015-02-26
filter for requested funding (fixes #813)

### LINKED ISSUE #813: Switch to requested funding in mercator proposal listing
This is kind of the opposite of #666: We now want the requested funding in both the mercator proposal list item and the mercator proposal listing column filter list.

MagdaN:
show requested funding in list and filter for it


nidico:
LGTM.

xi:
The acceptance tests need to be adapted.

PR #823, opened 2015-02-26
Fix backend-based list sorting

nidico:
This got accidently removed completely in
e4c72fb170ee6a23f70572140e8108fe521b54b9, instead of just removing the
minus-sign.


nidico:
Merging myself.

PR #831, opened 2015-03-02
Fix user unit tests

xi:
currently broken on master


MagdaN:
Can't reproduce failing test in master but in this branch also no tests are failing

PR #834, opened 2015-03-02
2015 03 cc fix 832

local-girl:


xi:
This conflicts with #833. Should be a simple merge however.

xi:
LGTM.

PR #837, opened 2015-03-02
2015 03 cc unsupported browser

local-girl:
Added german text for unsupported browser


MagdaN:
This works - I was just wondering if we should really 'duzen' the users in the german message, because I could imagine it's mostly older people that use IE9 and don't want to be 'duzt'. Also, in order to have less support requests, should we rather link to Chrome instead of Chromium?

PR #838, opened 2015-03-02
Fix comments

### LINKED ISSUE #803: Comment editing + reply brokenComment editing + reply

xi:
This fixes #803 and #804. The fixes are not very nice but should work for now. I really think we should refactor resourceWidget sometime soon in order to have cleaner solutions to problems like these.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/AngularHelpers/AngularHelpers.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/AngularHelpers/AngularHelpers.ts
@@ -86,16 +86,12 @@ export var recompileOnChange = ($compile : ng.ICompileService) => {
                 link = {post: link};
             }
 
-            var contents = element.contents().remove();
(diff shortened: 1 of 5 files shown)


nidico:
This doesn't work for me:

- The original comment is still shown after edit (also on 5f88e6a - unrelated to the errors below).
- Any mercator view (e.g. proposal listing) throws `Error: [$compile:ctreq]` for me.

xi:
I split this up into #845 and #846 in order to not mix up problems.

> The original comment is still shown after edit (also on 5f88e6a - unrelated to the error below).

Can not reproduce.

> Any mercator view (e.g. proposal listing) throws Error: [$compile:ctreq] for me.

How could I miss this. This was an issue with my "fix" for `adhRecompileOnChange`. I tried to do a better version in 9f1bd97. But I am not really sure what I am doing there.

PR #839, opened 2015-03-02
avoid multiple calls to adhPermissions.bindScope()

xi:
this is similar to #800. `bindScope()` does not expose the unregister function returned by `$watch`.

The solution here is different though: Instead of exposing the unregister function, we pass a function to `$watch`. This way, we do not need to keep track of the unregister function in the calling scope. The watch will automatically update and unregister when the scope is destroyed.


nidico:
This causes the following error on entering content in the comment create field:

```
TypeError: Cannot read property 'length' of undefined
    at Object.parentPath (Util.ts:39)
```

nidico:
LGTM.

PR #842, opened 2015-03-02
Attempt at fixing user page acceptance tests

pallix:
Do not merge yet. Testing with Travis.


pallix:
Can be merged.

xi:
I guess this works because it works around the one-column layout on small screens. However, I had some trouble understanding what was going on. I would suggest to either include the `get()` in `getUserName()` or at least to add a comment explaining why it is done like this.

xi:
LGTM.

PR #845, opened 2015-03-03
Fix comments (fixes #803)

### LINKED ISSUE #803: Comment editing + reply brokenComment editing + reply

xi:
continued from #838.

@nidico reported that this does not work, but for me it does. Probably a race condition in cache invalidation?


nidico:
This time it works for me. Let's investigate possible cache invalidation race conditions if the problem appears again.

PR #849, opened 2015-03-03
2015 03 jk refactor mv index views to catalog, subscribers to subscriber module

joka:


nidico:
Why do you prefer to have the indexes in a central `catalog` module instead of locally in the respective sheets?

joka:
> Why do you prefer to have the indexes in a central catalog module instead of locally in the respective sheets?

This is part of the refactoring to have a proper Single Responsibility on module level:
http://adhocracy3mercator.readthedocs.org/en/latest/style_guide/backend_structure.html

Moving the index views to the catalog modules solves the following problems:

  - multiple places you have to know and look after to understand how catalog indexing works (sheet..., catalog.__init__, substanced.catalog...)

  - different places than in substanced

 - multiple places to change if you want to replace or modfiy the catalog implementation

PR #850, opened 2015-03-03
Reset password story

local-girl:
updates relating to reset password should be merged here before they are merged to master


xi:
Continued in #873

PR #851, opened 2015-03-03
Test that messaging an user sends an email

pallix:


pallix:
I will add one more test to this branch.

pallix:
I'm done with the branch.

src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/shared.js
```diff
@@ -11,6 +11,13 @@ var hasClass = function (element, cls) {
     });
 };
 
+var diffArray = function(a, b) {
```
nidico: Can we use `_.difference` instead?

nidico:
Nice, LGTM!

PR #852, opened 2015-03-03
Improve delete

xi:
this implements delete for mercator proposals and also adds a confirm dialog before actual deletion.


nidico:
Works nicely, after adding a small fixup.

PR #853, opened 2015-03-03
reset password backend api & acceptance test

joka:


src/adhocracy_core/docs/authentication_api.rst
```diff
@@ -385,3 +385,67 @@ the user to be logged in from different devices at the same time. ::
...
+If the user clicks on this link, the frontend has to send a post request with the
+new password to the reset password end point::
+
+    >>> newest_reset_path = getfixture('newest_reset_path')
+    >>> prop = {'path': newest_reset_path,
+                'password': new_password}
+    >>> resp = testapp.post_json('/reset_password', prop)
+    >>> resp.status_code
+    200
+
```
joka: Regarding  version A  or B,  I would prefere the solution that is les work for backend/frontend.

MagdaN:
I guess I would also prefer Version A. @nidico: what do you think?

src/adhocracy_core/docs/authentication_api.rst
```diff
@@ -385,3 +385,67 @@ the user to be logged in from different devices at the same time. ::
...
+``/principals/passwordresets`` pool. The user is identified with the given
+email adress::
+
+    >>> prop = {'content_type': 'adhocracy_core.resources.principal.IPaswordReset',
+    ...         'data': {
+    ...              'adhocracy_core.sheets.principal.IPasswordReset': {
+    ...                  'email': 'anna@example.org'}}}
+    >>> resp_data = testapp.post_json(rest_url + "/principals/passwordresets", prop).json
+
+the url for the created password reset object is protected by obscurity::
```
mf59816: i think i'm missing a point here.  why does that link have to be sent to the user in the first place?  shouldn't it be in the email and on the server and nowhere else?

src/adhocracy_core/docs/authentication_api.rst
```diff
@@ -385,3 +385,67 @@ the user to be logged in from different devices at the same time. ::
...
+new password::
+
+  >>> prop = {'data': {'adhocracy_core.sheets.principal.IPasswordAuthentication': {
+    ...                     'password': 'new_password'}}}
+    >>> resp = testapp.put_json(password_reset_path, prop)
+    >>> resp.status_code
+    200
+
+Alternative B: User Password reset with API extension (more thentos friendly) 
+---------------------------------------------------------------------------------
```
mf59816: i don't think either alternative is significantly costlier to implement in thentos.  you should choose based on what's easiest for you!  (-:

src/adhocracy_core/docs/authentication_api.rst
joka: > i think i'm missing a point here. why does that link have to be sent to the user in the first place? shouldn't it be in the email and on the server and nowhere else?

backen sends link to the user, user clicks link, the link is pointing to the fronted, the fronten shows a nice UI and talks with the backend to set the new password

nidico: This confuses me as well - if the password reset resource ID *is* the token, it shouldn't be returned in the response - otherwise I could reset anybody's password...

Did we understand something wrong?

nidico:
Alternative A is more inline with our API and more RESTish. However I'm confused by the response (see comment above). I'm also unsure if you could PUT new passwords to the password reset resource multiple times.

Alternative B looks more straight-forward and allows for shorter reset trigger URLs, so I'm more for that option. Also, we're not very RESTish with `message_user` (no message resource), `login` (no "POST session" metaphor) either, so this isn't absolutely necessary to be inline with the rest of the API.

joka:
> Alternative A is more inline with our API and more RESTish. However I'm confused by the response (see comment above). I'm also unsure if you could PUT new passwords to the password reset resource multiple times.

after the first put, the reset password resource is removed

src/adhocracy_core/docs/authentication_api.rst
joka: > This confuses me as well - if the password reset resource ID is the token, it shouldn't be returned in the response - otherwise I could reset anybody's password...

yes, thats true

nidico:
@joka: Do you know which alternative to do?

joka:
> @joka: Do you know which alternative to do?

Version B, the API extension, but internally version A (password reset is a normal resources, but not accessible)

src/adhocracy_core/docs/authentication_api.rst
```diff
@@ -385,3 +385,29 @@ the user to be logged in from different devices at the same time. ::
...
+
+    >>> prop = {'email': 'anna@example.org'}
+    >>> resp = testapp.post_json(rest_url + "/create_password_reset", prop)
+    >>> resp.status_code
+    200
+
+On Success the backend sends an email with the link to reset the password to
+the user. The link contains the path to identify the password reset request::
+
+    http://frontend_url/resetpassword/?path=/1318...
```
xi: Can we change this to `password_reset`?

MagdaN:
@joka: please return POST when OPTION request is to /create_password_reset

MagdaN:
@joka : the email is now sent but when I use the link in the browser I get 404 - The resource could not be found. /resetpassword/

joka:
@pallix  can you review?

joka:
> @joka : the email is now sent but when I use the link from the email in the browser I get 404 - The resource could not be found. /resetpassword/

This looks like problem with the pyramid view configuration in adhocracy_frontend.

Note: the link is now called ../password_reset/..

src/adhocracy_core/adhocracy_core/resources/principal.py
```diff
@@ -245,6 +253,35 @@ class IGroupsService(IServicePool):
...
+    """ Resource to do one user password reset. """
+
+
+@implementer(IPasswordReset)
+class PasswordReset(Base):
+
+    """ Password reset implementation."""
+
+    def reset_password(self, password):
+        """ Set `password` for creator user and commit suicide."""
```
pallix: maybe we should use a term less metaphoric?!

joka: > maybe we should use a term less metaphoric?!

mmh, 'delete resource'?

joka:
> @joka : the email is now sent but when I use the link from the email in the browser I get 404 - The resource could not be found. /resetpassword/
> This looks like problem with the pyramid view configuration in adhocracy_frontend.
> 
> Note: the link is now called ../password_reset/..

@MagdaN  did you manage to fix this problem?

src/adhocracy_core/adhocracy_core/resources/test_subscriber.py
```diff
@@ -441,6 +441,39 @@ def test_with_request(self, context, registry, mock_sheet, monkeypatch):
...
+        registry.settings['adhocracy.frontend_url'] = 'http://front.end'
+        return registry
+
+    @fixture
+    def event(self, context, registry):
+        event.object = context
+        event.registry = registry
+        return event
+
+    def _call_fut(self, event):
```
pallix: fut?

src/adhocracy_core/adhocracy_core/rest/schemas.py
```diff
@@ -488,3 +496,80 @@ def _add_arbitrary_filter_node(name, index: SDIndex, schema):
...
+    context = kw['context']
+    request = kw['request']
+
+    def validate_email(node, value):
+        locator = request.registry.getMultiAdapter((context, request),
+                                                   IUserLocator)
+        user = locator.get_user_by_email(value)
+        if user is None:
+            msg = 'No user exists with this email: {0}'.format(value)
+            raise colander.Invalid(node, msg)
```
pallix: Could this be used to brute-force Adhocracy with a list of email and attempting to deanonymize users?

src/adhocracy_core/adhocracy_core/rest/test_schemas.py
```diff
@@ -801,3 +801,165 @@ def test_deserialize_recipient_is_no_user(self, request, context):
                   'text': 'text'}
         with raises(colander.Invalid):
             inst.deserialize(cstrut)
+
+
+class TestPOSTCreateResetPasswordRequestSchema:
+
+    def make_one(self):
```
pallix: should it be private?

src/adhocracy_core/adhocracy_core/testing.py
```diff
@@ -665,6 +673,17 @@ def newest_activation_path(app):
 
 
 @fixture(scope='class')
+def newest_reset_path(app):
+    """Return the newest reset password path generated by app."""
+    from urllib.request import unquote
+    mailer = app.registry.messenger._get_mailer()
+    last_message_body = mailer.outbox[-1].body
```
pallix: Not sure how this plays with the change to the mailer in the .ini files that I did when I wrote the Protractor. I changed the mailer to receive the mails on the disk.

src/adhocracy_core/adhocracy_core/resources/principal.py
pallix: remove itself? delete itself?

pallix:
Looks good. See comments above.

src/adhocracy_core/adhocracy_core/rest/schemas.py
nidico: Yes, but this can already be done through `register` as well.

If we want to avoid that, this has a rather high impact on UX, i.e. allow to register, but maybe without success - you need to check your emails first. Same for "forget password" - we would then need to write "if you have an account, you'll now get an email".

We can improve this later if we have a good idea.

pallix: A workaround is to have a delay before returning an answer, to avoid bruteforcing.

MagdaN:
added acceptance tests, please review

pallix:
Looks good to me.

nidico:
The error message in case of an expired / invalid token is too technical ("This resource path does not exist.") and should be translated into something which user understand (e.g. 

Also, this should be shown *instead* of the form with 2 password fields.

MagdaN:
This works nicely except for the fact that all the test after the password-reset-test fail due to the changed password. I will see that I can fix that

MagdaN:
@nidico : I tried some stuff, but they all don't seem to work- the only way I could fix this, is to change the running order of the tests, mercator first and core after that. I guess, the cleanest way would be to create an extra test user whose password gets changed and nothing else. What do you think?

nidico:
Let's merge!

pallix:
How is it guaranteed that UserLoginSpec will run last in the core suite? Would it be possible to fix the problem with setting the new password to the old one at the end of the test?

PR #854, opened 2015-03-03
update some js and ruby libs

xi:
Mostly bugfix releases. Nothing important as far as I could see.


xi:
There is a sass deprecation arning that got even longer with the updates. See https://github.com/ericam/susy/issues/425 for a discussion of that.

nidico:
Everything still seems alright.

PR #857, opened 2015-03-04
2015 02 mn translate countries

MagdaN:
This works for now, but the problem is, that the translated countries are not sorted properly. This will need a different implementation, which I would work on. It's up to you, if you want to use this for now - or if you want to wait for the sorted list.


MagdaN:
Sorting is now working - except that 'Österreich' is the last item in the german list, but for that we would need to add localized alphabets and I guess that is a bit too much, right?

nidico:
The branch contains multiple duplicate commits.

PR #862, opened 2015-03-05
Translate country select an countries

MagdaN:


nidico:
Could you please state in a comment where list and translations come from?

nidico:
Works nicely!

MagdaN:
The list comes from wikipedia: http://de.wikipedia.org/wiki/Liste_der_Staaten_der_Erde

PR #866, opened 2015-03-06
Enable piwik-based frontend tracking

nidico:
This allows to 

- track users (browser settings; session click paths with login state)
- specify whether cookies shall be used
- specify whether user ids shall be tracked (I'm very much against that in non-research projects, but I'm in favor of leaving this feature in in order to learn more about tracking).

This is just a beginning. We should further evaluate what we actually want to track, communicate that properly to the users and allow more fine-grained opt-out (e.g. track browser settings and click paths, but independently).


MagdaN:
Works nicely!

PR #870, opened 2015-03-06
Revert "Pin chrome to version 40" (see #868)

nidico:
Please don't merge before March 9.


nidico:
This doesn't seem to be done yet from sauce labs side. On hold.

nidico:
Without pinning, saucelabs now uses Chrome 41. Merging.

PR #875, opened 2015-03-09
2015 03 pa fix autodoc for members not being generated

pallix:
- Fix generation of members function with Sphinx
- Remove one warning during the code generation


joka:
prima

PR #879, opened 2015-03-10
refactor login ui

xi:
*based on #873*

When doing #873 I realized that the four login related pages (login, register, password_reset, create_password_reset) were very similar but also had some inconsistencies. I talked to the designers and did some refactoring.


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_login.scss
```diff
@@ -134,32 +90,38 @@ Register Dialog.
...
+    line-height: inherit;
+    margin: 0 0 2em;
+    text-align: left;
 }
 
-.register-support,
-.login-support,
-.register-cancel {
-    font-size: $font-size-small;
+// FIXME overspecifc selector
```
local-girl: Why not place this in the .login-form block ?

PR #881, opened 2015-03-10
embed provider

xi:
This changes the embed module so that `embeddableDirectives` can be configured in a decentralized manner.


nidico:
LGTM.

Travis tests randomly fail with the message `angular never provided resumeBootstrap`, but we should resolve this outside this PR.

PR #884, opened 2015-03-11
Allow different templates for different processes

xi:
This allows to define different templates for different process types.

It does not yet include a system to conigure topLevelState differently for different processes. That will be done in a separate pull request.


nidico:
LGTM.

PR #885, opened 2015-03-11
move spaces template to resourceArea (in core)

xi:
This reverses the order of `adh-process-view` and `adh-space` in the DOM.

It also reduces `adh-mercator-workbench` to the content space.


xi:
This should also fix the error space as it is moved out of `adh-process-view`

nidico:
LGTM!

PR #888, opened 2015-03-11
Allow to do processType specific configuration on resourceArea

xi:
Note that I did not move the configuration yet. I just added a new parameter to the config functions. We probably want on frontend package for each processType. In that, we depend on all resource specific packages, register a process template and configure resourceArea.


nidico:
LGTM.

PR #891, opened 2015-03-11
Fix deps2dot

xi:
this used filename instead of the moduleName variable in one place.


nidico:
This creates an empty graph due to the move.

Also due to the move, this isn't linked to the `build` directory anymore, so can't be used in customization packages.

xi:
> This creates an empty graph due to the move.
>
> Also due to the move, this isn't linked to the build directory anymore, so can't be used in customization packages.

For me the result depends on the current working directory (because it uses `git grep` internally). So if I want to get a graph of only a specific build, I `cd` to the respective build directory.

nidico:
Well, this didn't work due to trying to run it from `build` (as it's not in git).

PR #893, opened 2015-03-12
Improve documentation

pallix:
- restructure the folders
- rename style_guide to coding_guides
- create an architecture folder
- put Git sections into their own file
- put Python guidlines into their own file
- various minor improvements


joka:
fine

PR #895, opened 2015-03-15
Fix: Don't change creator on autoupdate

nidico:
Otherwise this leads to setting the creator of the updated resource to
be the creator of the resource triggering the update.

Missing:

- [ ] adapt tests
- [ ] write migration for broken resources


nidico:
@joka and I just discussed this matter, and it's more complicated than what I described: It isn't always clear that the creator of an auto-updated resource should be the same as the creator of the previous version (think of documents which get autoupdates by a section update).

Taking that into account, we might want to solve that on the frontend side instead by displaying the creator of the item instead of the creator of the version. (Unfortunately, we only have the version at hand at the moment...)

PR #897, opened 2015-03-15
Show comment item creator instead of version creator

nidico:
Reason: On autoupdate, the creator of the updated resource is set to the
user which triggered the update.

This is an alternative solution to #895, which solves the problem
differently in the backend (closes #895).


xi:
On `IMetadata` we already have `creation_date`, `item_creation_date`. On simple resource and items, these are always the same. But on versions they can be different. This provides a uniform interface for different kinds of resources.

We could do the same with creator, right? Then we could decide in the frontend which one to display.

joka:
> On IMetadata we already have creation_date, item_creation_date. On simple resource and items, these are always the same. But on versions they can be different. This provides a uniform interface for different kinds of resources.
> 
> We could do the same with creator, right? Then we could decide in the frontend which one to display.

We also considered adding "item_creator" to the IMetadata sheet. But this adds additional data noise. The frontend has the item and  the item creator anyway and could just display this instead of the item version creator, or?

nidico:
> We could do the same with creator, right?

Basically yes, but we can do that at another time.

xi:
LGTM.

PR #899, opened 2015-03-16
Do only show toplevel outdated npms

xi:
I updated the docs so that the command for outdated gems now only shows
one level. This way we do not see the outdated dependencies of the gems
we depend on.


nidico:
Good.

PR #902, opened 2015-03-16
add documentation on frontend routing

xi:
With the addition of processes (#884, 885, #888) routing has become even more complex.  This is why I created some overview over the routing system in general.

I am not sure whether the level of detail I chose is the best one. One the one hand, the text is already quite complex. On the other, more concrete examples of how to actually *do* things might be helpful (e.g. add a resource route, add a login-like route, make a directive embedabble).

I would like to get feedback from at least 2 frontend devs before merging this to make sure that it is comprehensible and useful.

### DIFF
--- a/docs/architecture/frontend_routing.rst
+++ b/docs/architecture/frontend_routing.rst
@@ -1,0 +1,141 @@
+Frontend Routing
+================
+
+As Adhocracy 3 is a single page application, the routing is done in the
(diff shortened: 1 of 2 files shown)


nidico:
Nice writeup. Assigning to @MagdaN to check for understandability.

xi:
@rukala do you want to have a look, too?

MagdaN:
Thanks, @xi , that will help a lot!

PR #903, opened 2015-03-16
Template etc/(frontend)development.ini

nidico:


xi:
Seems to work. Some open questions

-   What about `src/adhocracy_sample/`?
-   What about all the other `etc/*.ini` files?
-   AFAIU the *.ini files need to exist before the `meta_api` job is executed. Is this ensured?

nidico:
> What about src/adhocracy_sample/?

Don't know :/

> What about all the other etc/*.ini files?

Indeed we should also generate the `*test*.ini` and `noserver.ini` files...

> AFAIU the *.ini files need to exist before the meta_api job is executed. Is this ensured?

We can ensure this by pointing this to the output of the (yet to be created) `noserver.ini` part.

xi:
LGTM.

PR #904, opened 2015-03-16
Fix frontend core

xi:
this improves on #892 and #901 by fixing more small bugs in the mostly unmaintained adhocracy core frontend.

Note that I could not get proposal creation to work because the sample resources were missing from the backend (the sheets were there). @joka could you please have a look?


nidico:
LGTM.

PR #906, opened 2015-03-16
Use unminified angular in development (fixes #890)

### LINKED ISSUE #890: Use unminified angular in development
I tried to debug an angular module issue. With minified angular I got a not very helpful error message. When I switched to the unminified version, I got a much more useful message.

So I propose to always use uniminified angular in development.

xi:


nidico:
LGTM.

PR #908, opened 2015-03-17
added comment highlighting fixes #905

### LINKED ISSUE #905: highlighting of selected comment
Comments may be reported as abusive. There are also routes for individual comments and the `is-selected` and `is-not-selected` class are properly set. But there is not CSS code for high-/downlighting the selected comment. This way moderators can not see which comment was reported.

local-girl:
It's not pretty but it works.


xi:
LGTM.

PR #909, opened 2015-03-17
Create frontend and backend packages for meinberlin

nidico:
Note that building meinberlin frontend fails due to `Adhocracy.ts` from `adhocracy_frontend` depending on `DocumentWorkbench`, depending on `IProposal`, which is only in `adhocracy_sample`.

In order to use the frontend package, `Adhocracy.ts` needs to be overwritten.


xi:
I am in favor of merging this. It still does not build, probably due to #916.

PR #912, opened 2015-03-17
Make URL clickable where available fixes #911

### LINKED ISSUE #911: Org URL not clickable in detail view

local-girl:


xi:
AFAIK, website is optional. So there should only be a link if it has been set.

Additionally, the link should also be in the "planned_nonprofit" and "planned_nonprofit" cases.

xi:
> AFAIK, website is optional. So there should only be a link if it has been set.

To clearify: It seems like `ngHref` does not create a `href` attribute if the value is empty. But the `<a>` tag is still there resulting in wrong styling.

local-girl:
Fixed up in 7775fa6

xi:
I think we should still use `ngHref` in this case (see also #914)

local-girl:
fixup 0a72f7d

xi:
LGTM

PR #913, opened 2015-03-17
separate translations for packages

xi:
this split the mercator specific translations from core and also updates the merge_messages script so it can handle multiple packages.


nidico:
LGTM and seems to work; tests also seem to pass (after merging master).

PR #922, opened 2015-03-17
fix meinberlin build

xi:
adhocracy_core (and therefore meinberlin) depends on adhocracy_sample (although it arguably should not.  The only dependency AFAIK is in two JavaScript tests that are mostly unmaintained anyway)

This reflects that dependency in the code to fix the build. Buildout now executes without an error for me.


MagdaN:
This builds. Only problem - frontend redirects to /r/adhocracy/

MagdaN:
@nidico: can you also review, please?

nidico:
LGTM.

PR #925, opened 2015-03-18
2015 03 cc fix sass color variables

### LINKED ISSUE #921: scss-lint ColorVariable issues
scss-lint add a [ColorVariable](https://github.com/causes/scss-lint/blob/master/lib/scss_lint/linter/README.md#colorvariable) linter in 0.34.0. I think this is a good idea, but our code violates that rules in several cases.

So we either have to disable that rules or (preferred) fix our scss code.

local-girl:
Changed literal colors to variables to please linting fixes #921

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
@@ -32,7 +32,7 @@ $arrow-width: 1em;
     text-transform: uppercase;
     line-height: $line-height;
     background-color: $color-brand-one-normal;
-    color: white;
+    color: $color-text-inverted;
     text-align: center;
     text-decoration: none;
     min-width: 100px / $font-size-small * 1em;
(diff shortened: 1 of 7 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
```diff
@@ -43,15 +43,15 @@ $arrow-width: 1em;
         left: -$arrow-width;
         width: 0;
         height: 0;
-        border-top: $line-height/2 * 1em solid transparent;
-        border-bottom: $line-height/2 * 1em solid transparent;
+        border-top: $line-height/2 * 1em solid $color-structure-transparent;
+        border-bottom: $line-height/2 * 1em solid $color-structure-transparent;
```
xi: AFAIK this is a but in scss-lint and is already fixed in the latest master (See https://github.com/causes/scss-lint/blob/master/CHANGELOG.md)

xi:
LGTM. We should remove `$color-structure-transparent` as soon as the next scss-lint has been released.

PR #927, opened 2015-03-19
2015 03 jk workflows

joka:


src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+    >>> data = {'adhocracy_core.sheets.name.IName': {'name': 'proposals'}}
+    >>> resp = app_god.post_resource('/', IBasicPool, data)
+    >>> data = {'adhocracy_core.sheets.name.IName': {'name': 'proposal_item'}}
+    >>> resp = app_god.post_resource('/proposals', IProposal, data)
+
+
+Workflows
+---------
+
+Workflows are fine state machines Assigned to a Resource.
```
xi: finite?

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+
+a local ACL (see doc:`authorization`) that is set when entering this state, ::
+
+    >>> state['acl']
+    [('Disallow, 'role:reader', 'view')]
+
+a hint for the frontend if displaying this state in listing should be restricted::
+
+    >>> state['display_only_to_roles']
+    ['Manager']
```
xi: Currently, the concept of roles exists only in the backend. The Frontend only uses `OPTIONS` requests to check permissions.

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+excecute arbitrary tasks ::
+
+     >>> transition = workflow['transitions']['to_announced']
+     >>> pprint(transition)
+     {
+      'callback': 'adhocracy_core.workflows.samplecallback',
+      'from_state': 'draft',
+      'to_state': 'announced',
+      'permission': 'review',
+      }
```
xi: What does "callable" mean in this context? Is this an API endpoint? If it is an actual python object, why should it be exposed to the frontend?

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+
+Workflow Assigment
+------------------
+
+Resources have a WorkflowAssigenment sheet to assign the wanted workflow::
+
+    >>> resp_data = app_god.get('/proposals/proposal_item').json
+    >>> workflow_data = resp_data['data']['adhocracy_core.sheets.workflow.ISampleWorkflowAssignment']
+    >>> workflow_data['workflow_type']
+    'sampleworkflow'
```
xi: All resources have that? I thought only process resources can have workflows.

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+
+in addition we can add custom metadata for specific workflow states::
+
+    >>> workflow_data['announced']['start_date']
+    '2015-02-14...,
+    >>> workflow_data['announced']['description]
+    'Soon you can participate...
```
xi: Would  be better to have a separate namespace for that, e.g.

```
>>> workflow_data['phase']['announced']['start_date']
'2015-02-14...,
>>> workflow_data['phase']['announced']['description]
'Soon you can participate...

File: src/adhocracy_core/docs/workflows.rst [hunk 6]
```diff
@@ -0,0 +1,119 @@
...
+
+Workflow transition to states
+-----------------------------
+
+The Metadata sheets shows the current workflow state::
+
+    >>> resp_data = app_god.get('/proposals/proposal_item').json
+    >>> metadata_data = resp_data['data']['adhocracy_core.sheets.metadata.IMetadataSheet']
+    >>> workflow_data['workflow_state']
+    'announced'
```
xi: As discussed [above](#r26924024) I do not yet understand why every resource should have a workflow. And putting workflow related data in the metadata sheet works only in that case.

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+
+We can also modify the state if the workflow has a suitable transition ::
+
+    >>> resp_data = app_god.options('/proposals/proposal_item')
+    >>> available_next_states = resp_data['PUT']['adhocracy_core.sheets.metadata.IMetadataSheet'][workflow_state]
+    ['published']
+
+NOTE: The available next states depend on the workflow transitions and user permissions.
+NOTE: To make this work every state may have only one transition to another state.
```
xi: That is a serious limitation. We should really make sure we do not need this.

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+
+TODO: How to hide resource and subresources if needed for state?
+      We can set/remove the local view permission for the resource.
+      Then you cannot do get requests anymore. But we never implementent this
+      properly, you can still view the resource in listings and the client
+      excpect that its possible to do get requests.
```
xi: I do not understand this.

xi:
I had a quick look and created some questions. (Note that one or two may be collapsed because I later pushed some Gardening on top of that)

Note that I do neither feel competent to review this nor did I spend a lot of time. So someone else should definitely have a look.

src/adhocracy_core/docs/workflows.rst
joka: > Currently, the concept of roles exists only in the backend. The Frontend only uses OPTIONS requests to check permissions.

yip I know, but how should we do this with a options request instead?

src/adhocracy_core/docs/workflows.rst
joka: > What does "callable" mean in this context? Is this an API endpoint? If it is an actual python object, why should it be exposed to the frontend?

Exposing the callable makes sense if the frontend wants to create new workflows in the future.
Maybe for know its more usefull to display the callable docstring instead. Then the frontend developer knows better whats happening when entering this state.

src/adhocracy_core/docs/workflows.rst
joka: > All resources have that? I thought only process resources can have workflows.

no, we also have use cases with workflows for proposals and maybe other resource types.
For example an continous workflow for one proposal item: initial -> Veränderbar <->  Abstimmbar -> Frozen

https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie#section-14

src/adhocracy_core/docs/workflows.rst
joka: > As discussed above I do not yet understand why every resource should have a workflow. And putting workflow related data in the metadata sheet works only in that case.

Workflow could be usefull for differen resource types, resources without it just display None, thats not much overhead.

I proposed to add this information to the metadata sheet because:

* it is some kind of metadata
* we need special permission checks on field level, thats why we need a special sheet, so why not use the metadata sheet thats already there.
* the possible values depend on the workflow and user permission, thats why we need to exposes these with the options request response, the metadata sheet alread has a similar behavior.

src/adhocracy_core/docs/workflows.rst
joka: > Would be better to have a separate namespace for that, e.g.
> 
> >>> workflow_data['phase']['announced']['start_date']
> '2015-02-14...,
> >>> workflow_data['phase']['announced']['description]
> 'Soon you can participate...

ok, I would only prefer the phrase 'state' over 'phase'. 
'phase' is not a common word for finite state machines and may imply a sequential workflow, but thats not always the case

src/adhocracy_core/docs/workflows.rst
joka: > That is a serious limitation. We should really make sure we do not need this.

yip, but this woud imply a more complex api that allows to trigger specific transitions. 
I suggest we do this only when we know that we need it. 
Up to then we can add some sanity checks in the backend.

joka:
> I had a quick look and created some questions. (Note that one or two may be collapsed because I later pushed some Gardening on top of that)

Thank you for gardening, I cannot find you hidden comments, are they still important?


> Note that I do neither feel competent to review this nor did I spend a lot of time. So someone else should definitely have a look.

Do you have a good feeling and do you think this is what the frontend needs?

joka:
@nidico can you also review?

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,119 @@
...
+    >>> available_next_states = resp_data['PUT']['adhocracy_core.sheets.metadata.IMetadataSheet'][workflow_state]
+    ['published']
+
+NOTE: The available next states depend on the workflow transitions and user permissions.
+NOTE: To make this work every state may have only one transition to another state.
+
+TODO: How to hide resources and subresources if needed for state?
+      We can set/remove the local view permission for the resource.
+      Then you cannot do GET requests anymore. But we never implemented this
+      properly, you can still view the resource in listings and the client
```
joka: @xi How does the frontend handle links to resources that do not allow GET requests for the current user?
      Do we need a proper "Not Authorized" error message? (we talk about this and I don't have a technical solution at the moment)

src/adhocracy_core/docs/workflows.rst
joka: > What does "callable" mean in this context? Is this an API endpoint? If it is an actual python object, why should it be exposed to the frontend?
> 
> Exposing the callable makes sense if the frontend wants to create new workflows in the future.
> Maybe for now its more usefull to display the callable docstring instead. Then the frontend developer knows better whats happening when entering this state.

ok, we remove callable and permission because this is not usefull for the frontend at the moment.
In future we may add the docstring explain what this transition does.

src/adhocracy_core/docs/workflows.rst
joka: > @xi How does the frontend handle links to resources that do not allow GET requests for the current user?
> Do we need a proper "Not Authorized" error message? (we talk about this and I don't have a technical solution at the moment)

To make it work with the frontend we need to filter resources without view permission from pool listings and backreferences. 
If the frontend still gets resources without view permission the backend return 'Not Found', the frontend already handles this.

src/adhocracy_core/docs/workflows.rst
joka: > Would be better to have a separate namespace for that, e.g.

ups, I have objections against this:

- a namespace is not realy needed because we only have states here, nothing else
- a namespace makes the datamodelling in the backend more complex

An alternative solution would be to have a list of state dictionaries.

xi: If there is nothing else here, no namespaces is fine with me. But in  this document, at least `workflow_state` and `workflow_type` are mentioned.

joka:
@xi  or @MagdaN 
bin/buildout install resources fails:

    throw "mkFieldType: unknown value " + field.valuetype;
                                                ^
    mkFieldType: unknown value adhocracy_core.sheets.workflow.StateAssignment

Can you add StateAssignment to mkResources.ts? 
This schema type describes the following dictionary:

     {'start_date': 'DateTime',
       'description: 'Text'}

xi:
> Can you add StateAssignment to mkResources.ts?

Done in 341334c. Note that this only works by accident. We need to fix #962 to actually make dict fields work consistently.

MagdaN:
Builds!

MagdaN:
Looks good.

joka:
I merged master, @xi  I did not find any 'ConstructurType' variables in the frontend code, but the frontend is building fine

xi:
> @xi I did not find any 'ConstructurType' variables in the frontend code, but the frontend is building fine

Fixed in d68acda.

joka:
all python tests pass
mercator acceptance and login tests are failing from time to time, but this problem is also on master

src/adhocracy_core/adhocracy_core/content/__init__.py
```diff
@@ -36,6 +37,16 @@ def __init__(self, registry):
         Dictionary with key isheet (`sheet type` interface) and value
         :class:`adhocracy_core.interfaces.SheetMetadata`.
         """
+        self.workflows_meta = dict()
```
pallix: why do you prefer dict() over {} ?

src/adhocracy_core/adhocracy_core/workflows/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+@implementer(IAdhocracyWorkflow)
+class AdhocracyACLWorkflow(ACLWorkflow):
+
+    """Workflow that sets the :term:`acl` when entering a State."""
+
+    def get_next_states(self, context, request: IRequest) -> list:
+        """Get states you can trigger a transition to."""
+        state = self.state_of(context)
+        transitions = self.get_transitions(context, request, from_state=state)
+        states = [t['to_state'] for t in transitions]
```
pallix: if I understand correctly if two or more transitions lead to the same state, the state will appear multiple times in the returned list. Is this wanted?

src/adhocracy_core/docs/workflows.rst
```diff
@@ -0,0 +1,108 @@
...
+-----------------------------
+
+We can also modify the state if the workflow has a suitable transition::
+
+    >>> resp_data = app_god.options('/proposals/proposal_item').json
+    >>> resp_data['PUT']['request_body']['data']['adhocracy_core.sheets.workflow.ISample']
+    {'workflow_state': ['announced']}
+
+NOTE: The available next states depend on the workflow transitions and user permissions.
+NOTE: To make this work every state may have only one transition to another state.
```
pallix: guess this answer my previous question :-)

src/adhocracy_core/adhocracy_core/sheets/workflow.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if 'default' not in kwargs:  # pragma: no branch
+            self.default = {}
+
+    description = Text(missing='',
+                       default='Soon you can participate')
+    start_date = DateTime(missing=None,
+                          default=datetime(2015, 2, 14))
```
pallix: valentine's day as a default?

src/adhocracy_core/adhocracy_core/workflows/schemas.py
```diff
@@ -0,0 +1,103 @@
...
+
+class StateMeta(MappingSchema):
+
+    """Workflow state."""
+
+    title = SingleLine(missing='')
+    description = Text(missing='')
+    acl = ACL()
+    display_only_to_roles = Roles(missing=[])
+    """Hint for the fronted, this not security related."""
```
pallix: typo

src/adhocracy_core/adhocracy_core/workflows/schemas.py
```diff
@@ -0,0 +1,103 @@
...
+class TransitionMeta(MappingSchema):
+
+    """Workflow transition to state."""
+
+    callback = WorkflowCallback()
+    from_state = SingleLine(missing=required)
+    to_state = SingleLine(missing=required)
+    permission = SingleLine(missing='')
+
+    # FIXME add default permission to do transitions
```
pallix: still relevant?

pallix:
Looks good to me after an overview of  the code (i'm not understanding everything). A few minor comments above.

joka:
Can we merge this now?

PR #928, opened 2015-03-19
Sort proposals by date

nidico:
This adds an `item_creation_date` index in the backend and allows to use it in the mercator proposal listing.

In order to make this work, the data has to be reindexed (`bin/sd_reindex etc/development.ini`).

Some peculiarity: It looks like there is a different number of items when sorting this or that way, at least in my testdata - haven't investigated that any further yet.

Missing:

- [x] backend unit tests
- [ ] some way to avoid triggering `update` in the frontend twice (can also be done later)


MagdaN:
I tried it but could not really test it, as all my proposals are from the same day. Also, even after updating the index I got an 400 error - but maybe there is still something wrong with my database.

nidico:
Rebased.

xi:
I also got 400 errors when I tested this with an existing database. It worked fine on a fresh database. But I fear that #936 was caused by this. @nidico could you please check?

nidico:
> I also got 400 errors when I tested this with an existing database.

Did you reindex as described above?

> But I fear that #936 was caused by this.

Don't understand how.

> @nidico could you please check?

Can do.

xi:
>> I also got 400 errors when I tested this with an existing database.
> Did you reindex as described above?
Yes.

>> But I fear that #936 was caused by this.
> Don't understand how.

According to @slomo, #936 was not an issue a week ago. Since the deployment of this branch is AFAIK the only thing that changed I assumed it might be a reason.

nidico:
This branch has not been deployed on production.

PR #930, opened 2015-03-20
2015 03 mn add maps

MagdaN:


MagdaN:
This should work when using the buildout-meinberlin.cfg for buildout and open Link: http://localhost:6551/embed/map-input?id=mymap&lng=13.3592486393&lat=52.5011278698&height=500

MagdaN:
I am still working on this

xi:
I did some refactoring. Some notable changes:

-  `l.Map` can also take a `DOMElement` as first argument. I think this is preferable because ids need to be unique.
-  I added types. That required some small changes, e.g. I had to rename `L` to `leaflet`.

xi:
LGTM.

PR #935, opened 2015-03-23
API Change: password reset works also for not activated users

joka:


joka:
Our coverage ist broken, I don't know if this branch was the reason.

pallix:
Just for reference: coverage not by 100% since this build: https://coveralls.io/builds/2164412 when the export script was included. Master is now by 100% thanks to your lastest changes.

PR #939, opened 2015-03-24
2015 03 cc mein berlin create form

local-girl:
Simple skeleton for the create form, I don't have any info for validation.


local-girl:
http://localhost:6551/embed/mein-berlin-create :D

MagdaN:
Works

PR #942, opened 2015-03-25
mv sglclick hack into its own service

xi:
#930 used a hack to detect single clicks (click events that were not part of a double click). This refactors that by moving the hack into its own service.

Note: I am not too happy about the service name (`adhClickContext`). I am open for proposals.


MagdaN:
Works!

PR #944, opened 2015-03-25
Add polygon to map input

xi:
this adds basic polygon support to `adhMapInput`. Passing a polygon is currently required. I am not sure whther that is a good idea.


MagdaN:
Works!

PR #946, opened 2015-03-26
Export users script with dates for rates

pallix:


nidico:
This could still be improved (in particular better performance), but is already useful as it is.

pallix:
Talked with @joka and performances can be improved by using the catalog queries when looking for which user has voted for which proposal.

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
```diff
@@ -7,16 +7,61 @@
 import sys
 import textwrap
 import time
-
 from pyramid.paster import bootstrap
```
joka: style guide: import groups should be sepearted by empty line

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
```diff
@@ -7,16 +7,61 @@
...
-from adhocracy_core.sheets.metadata import IMetadata
+from adhocracy_core.sheets.rate import IRateable
+from adhocracy_core.sheets.rate import IRate
 from adhocracy_core.utils import get_sheet_field
+from adhocracy_mercator.resources.mercator import IMercatorProposalVersion
+from adhocracy_mercator.sheets.mercator import ITitle
+from adhocracy_core.sheets.tags import filter_by_tag
+
+
+def _get_most_rated_proposals(root, min_rate=100):
```
joka: style guide: 

- place helper functions below function that uses it 
- one abstrcation level per function
- short  (5 lines if possible), split in helper functions with lower abstraction level
- not more than 2 indention levels
- seperate error handling from actually doing something (not usefull for very simple functions)

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
```diff
@@ -7,16 +7,61 @@
...
+
+    if len(rated) == 1:
+        rate = rated[0]
+        creation_date = get_sheet_field(rate, IMetadata, 'item_creation_date')
+        return (get_sheet_field(rate, IRate, 'rate'),
+                creation_date.strftime('%Y-%m-%d_%H:%M:%S'))
+    return (0, '')
 
 
 def export_users():
```
joka: To  export all user it would be much simpler to just use bin/sd_dump   ../principals/users
But in addition this function add most  rated proposals. This is not clear when reading the function name or docstring.

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
```diff
@@ -43,27 +88,39 @@ def export_users():
 
     users = query.execute()
 
+    # FIXME: set 100 instead of 0 on prod
```
joka: style guide:

- one abstrcation level per function
- short  (5 lines if possible), split in helper functions with lower abstraction level instead
- not more than 2 indention levels
- seperate error handling / command line options handling from actually doing something (not usefull for very simple functions)

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
joka: The code would be simpler if you use the pool sheet to search the catalog. 
You can add all filter prameters describes in rest_api.rst  to the pool sheet get method.

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
```diff
@@ -7,16 +7,61 @@
...
+    last_rates = filter_by_tag(rates, 'LAST')
+    rated = [rate for rate in last_rates
+             if _get_rate_subject_name(rate) == user_name]
+
+    if len(rated) == 1:
+        rate = rated[0]
+        creation_date = get_sheet_field(rate, IMetadata, 'item_creation_date')
+        return (get_sheet_field(rate, IRate, 'rate'),
+                creation_date.strftime('%Y-%m-%d_%H:%M:%S'))
+    return (0, '')
```
joka: same as above, just use the pool sheet for searching and aggregation

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
joka: @pallix @nico I added some styleguide comments

joka: You don't need the catalog to find all users. Instead get the users service:

    from substanced.util import find_service
    users = find_service(context, 'principals', 'users')


and iterate all children.

joka: > @pallix @nico I added some styleguide comments

We don't need to refactor this script, but it would be good to make it better next time. 
Also maybe we need to improve the backend documentation about  catalog searches with the pool sheet.

src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py
```diff
@@ -7,16 +7,61 @@
...
+    interfaces = catalog['interfaces']
+    tag = adhocracy_catalog['tag']
+
+    query = path.eq('/mercator') \
+        & interfaces.eq(IMercatorProposalVersion) \
+        & tag.eq('LAST')
+
+    proposals = query.execute()
+    proposals = [p for p in proposals if index_rates(p, None) > min_rate]
+    proposals = list(reversed(sorted(proposals,
```
joka: performance: What we need here is a search for proposals aggregatedby  `rates` index. 
The current filtered pool sheet implementation does this already, but only returns the count number for each `rate` index value

nico:
@joka: Wrong nico

PR #948, opened 2015-03-29
Fix column assignment in export proposal script (fixes #931)

### LINKED ISSUE #931: Budget and request funding inverted in the export script
It seems to me that the budget and requested funding values are interverted when exporting proposals with the export script.

Here https://github.com/liqd/adhocracy3.mercator/blob/ae2b0673b2afc95ba04405e58dd3ff5c867c6c9e/src/adhocracy_mercator/adhocracy_mercator/scripts/export.py#L107 the budget is exported first, but here the requested fundind is declared first:

https://github.com/liqd/adhocracy3.mercator/blob/ae2b0673b2afc95ba04405e58dd3ff5c867c6c9e/src/adhocracy_mercator/adhocracy_mercator/scripts/export.py#L63

nidico:


xi:
LGTM.

PR #949, opened 2015-03-30
automatically redirect on login/register when loggin in

xi:
this is usefull if you log in in another tab/window.


xi:
there are some open UX questions, so "on hold".

joka:
We have a similar problem with the activation link.
If the user clicks this link, but he is already logged in, he gets an error message 
and does not now how to proceed..

xi:
> We have a similar problem with the activation link.

I think that's different. Imagine someone is already logged in on a computer and then a second person tries to use an activation link in the same browser. For me it is not clear whether we should assume the activating user to be the same as the already logged in one or not.

Also, IMHO the bigger problem with activation links is that we can not distinguish between invalid, expired and already used activation links. So people who use their activation links twice will get a confusing error message.

I suggest opening a separate issue for activation as it has distinctive issues of its own.

xi:
This is continued in #1055.

PR #952, opened 2015-03-31
Add current/total numbers to load-more button

nidico:


local-girl:
Does {{currentLimit}}/{{totalCount}} need to be in the translation variable as it's just numbers. Might confuse the translators

nidico:
Yes, well, in German it's in the middle of the string. It could by simplified to just one variable, but that's not a big difference, and the way it's done now allows for localization.

xi:
LGTM.

PR #958, opened 2015-04-01
Enable travis building for meinberlin and mercator

slomo:


slomo:
@nidico or @xi ready to be review

src/adhocracy_core/base.cfg
```diff
@@ -65,6 +66,11 @@ recipe = collective.recipe.omelette
 eggs =
     ${adhocracy:eggs}
 
+[coveragerc]
+recipe = collective.recipe.template
+input = ${buildout:directory}/etc/coveragerc.in
+output = ${buildout:directory}/etc/coveragerc
+
 [test_run_unit]
```
joka: I don't see the point in having a custom coveragerc for every buildout config.
What coverage report should we send to coverall.us?
Also it would be much nicer to get a coverage for all packages together.

nidico:
This is already partially implemented with #1353. The remaining changes should be split in separate PRs.

slomo:
I think the remaining issue are not relevant any more

PR #959, opened 2015-04-01
Allow to pass non-string arguments to sheet constructors

xi:
Before this, sheet constructors expected all arguments to be strings
because that is what we receive from the backend. The constructors would
then apply parsers to convert the incoming strings to native types.

When you want to create a sheet in the frontend, this is rather annoying.
So I changed the constructors to accept argumants of any type and only
convert them if they are strings.


nidico:
LGTM. At some point, we should refactor the sheet import / parsing routine as described in #962.

PR #967, opened 2015-04-01
2015 03 cc mein proposal ui

local-girl:
Some very basic styling to meinberlin proposal form based on preliminary designs. Thought I would push it as it fixes an issue where focused fields look invalid as they are red! This may need more thought.


src/meinberlin/meinberlin/static/stylesheets/scss/a3.scss
```diff
@@ -20,8 +20,12 @@
 
 @import "components/type";
 @import "components/forms";
+@import "components/forms_theme";
 @import "components/button";
+@import "components/button_theme";
 @import "components/login";
+@import "components/map_theme";
+@import "components/map";
```
xi: I think these two need to be exchanged.

local-girl:
I'll put it in my next pull request

local-girl:
map_theme has since been removed as styles are in core

PR #968, opened 2015-04-02
Fix configuration to run all available tests

pallix:


joka:
works, but there are failing python tests in adhocracy_fronten now, so we cannot merge this.
@pallix can you fix this?

buildout-mercator.cfg
```diff
@@ -24,10 +24,10 @@ frontend_package_name = mercator
 backend_package_name = adhocracy_mercator
 
 [test_run_unit]
-package_paths = src/adhocracy_core src/adhocracy_mercator src/adhocracy_sample
+package_paths = src/adhocracy_* src/adhocracy_frontend/adhocracy_frontend/tests/unit
```
joka: we should add "adhocracy_*"  also to buildout-myberlin and builedout-core.cfg

joka: we don not need the line:  ..adhocracy_frontend/tests/unit , theses are only javascript unit tests

joka:
i fixed the failing tests and marked the unfixable tests as xfail.

We just need to merge master, then we can merge

pallix:
Thanks for the help

PR #969, opened 2015-04-02
added new version of maplist widget

MagdaN:


MagdaN:
Designers need to decide which version they want

MagdaN:
Designers decided - for embed view we have the a horizontal list underneath the map

MagdaN:
This can be reviewed - test by going to: /embed/mein-berlin-kiezkassen-proposal-list

PR #970, opened 2015-04-07
Moved button styles to core, updated mercator to reflect

local-girl:
Mercator and MeinBerlin share very similar button styles so I have updated core buttons to reflect this so we need much less sass in the button theme file

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
@@ -14,7 +14,6 @@
 
 @mixin button-color($fg, $bg) {
     color: $fg;
-    border-color: $fg;
     background-color: $bg;
 
     &.m-call-to-action {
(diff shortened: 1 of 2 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
```diff
@@ -68,32 +67,47 @@ input[type="reset"],
...
+    background: transparent;
+    border: 0;
+    color: $color-text-highlight-introvert;
+    font-size: $font-size-normal;
+    font-weight: $font-weight-normal;
+    line-height: $button-height;
+    height: $button-height * 1em;
+    padding: 0;
     text-transform: uppercase;
+    min-width: 1px;
```
xi: I think this can be removed as it is no longer needed to overwrite `min-width: 9em` from line 79.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
```diff
@@ -68,32 +67,47 @@ input[type="reset"],
...
-    &:active:not(:disabled):not(.ng-disabled),
-    &:focus:not(:disabled):not(.ng-disabled) {
-        outline: 0;
-    }
-
-    &:hover:not(:disabled):not(.ng-disabled),
-    &:focus:not(:disabled):not(.ng-disabled) {
+    &:focus:not(:disabled):not(.ng-disabled),
+    &:hover:not(:disabled):not(.ng-disabled) {
         @include button-color($color-text-highlight-introvert, $color-text-inverted);
```
xi: This is overwritten in the next two lines, right?

xi:
Looks good to me. Feel free to merge.

PR #976, opened 2015-04-07
2015 03 cc mein proposal ui2

local-girl:
More general UI styling to mein berlin


xi:
There are few issues:

-   Error messages take up space
-   Some details (mostly spacing) in the proposal create form make it a bit confusing.

Merging anyway.

PR #984, opened 2015-04-08
Redirect after proposal submit

xi:
I also added basic error handling but could not trigger that because `logBackendBatchError` does not handle non-400 errors proberly.


MagdaN:
This works nicely.

PR #987, opened 2015-04-10
watch if map was resized and update map

MagdaN:
This is a hack - if you have a better idea how to solve this, let me know. The problem is, when the columns are moving the map needs to be re-sized.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
@@ -156,7 +156,7 @@ export var mapInput = (
                 leaflet.Util.setOptions(map, {
                     minZoom: map.getZoom()
                 });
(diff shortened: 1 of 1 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
```diff
@@ -150,6 +149,16 @@ export var mapInput = (
                 scope.mapClicked = false;
                 scope.text = "TR__MAP_EXPLAIN_CLICK";
             };
+
+            scope.$watch((<any>map), function(){
```
xi: AFAIK the first argument need to be either a string or a function. Why do you watch map by the way? I guess what we really want to watch is the element width, right?

xi:
This seems to work well. I am not too sure about removing the `maxBounds` but that feature is just too buggy and complicated. We should also port this fix to the other map widgets.

PR #988, opened 2015-04-10
2015 04 mn update proposal script

MagdaN:


nidico:
LGTM. I just slightly fixed the column headings.

PR #991, opened 2015-04-11
2015 04 jk run all tests

joka:
Problem:

   * no script to run all tests
   * documentation how to run tests complicated/broken
   * broken tests because nobody runs them (happend twice in python test)

Proposal:
    
   * user polytester to run all python and available javascript/acceptance tests
   * simplify buildout config to create testscripts and add dependencies to python path
   * updated documentation

Example: 

    run:
    
        bin/polytester -c test.yml
    
    restrict tests:
    
        bin/polytester -c test.yml pyunit
    
    enable verbose output:

           bin/polytester -v -c test.yml pyunit

### DIFF
--- a/buildout-core.cfg
+++ b/buildout-core.cfg
@@ -18,6 +18,7 @@ parts +=
 
 [adhocracy]
 frontend.static_dir = src/adhocracy_frontend/adhocracy_frontend/build
+eggs += mercator[debug]
 
 [test_run_unit]
 package_paths = src/adhocracy_*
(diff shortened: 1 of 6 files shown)


joka:
@nidico  @pallix  can you give some feedback?

pallix:
I had to specify tests with `--config` with polytester 1.2.0.
It seems the default file is named `tests.yml`, we can rename `test.yml` in `tests.yml`

joka:
> It seems the default file is named tests.yml, we can rename test.yml in tests.yml

polytest did also look in  .bin/ directory instead of ./ for tests.yml

joka:
added new pull request #1031

PR #992, opened 2015-04-13
first detail view with map

MagdaN:


MagdaN:
The Designers want the map to be always open - and if the description text gets too long they want a button that scrolls down to the map. But as we don't know the max text length of the description (at the moment its 500 which is quite short) I did not implement the scrolling yet.

local-girl:
map
This is the design, I assume it's this height and when you press the arrows it expends ? To a lightbox ?

local-girl:
Forgot to comment: After speaking with designers we don't need anything like this yet and the designs will change so I have just removed reference to this function & related classes for now. MERGED

PR #998, opened 2015-04-14
update kiezkassen proposal listitem

xi:
mostly styling


local-girl:
This is fine for initial styling, i will fine tune in another branch

PR #999, opened 2015-04-14
Basic comment UI

xi:
The forms (reply, edit) need some more attention.


local-girl:
Will fine tune in another branch

PR #1000, opened 2015-04-14
Fix comment reporting in meinberlin

xi:


local-girl:
LGTM

local-girl:
This is fine for initial styling will need fine tuning

PR #1002, opened 2015-04-15
Automatically redirect old versions to newest version

nidico:
As the versioning system isn't in full use yet, i.e. a linear history is
enforced, and this history isn't exposed to the UI yet, this commit
makes views on old versions of versionables automatically redirect to
the newest version.

This will be changed as soon as versioning will be exposed through the
UI.


xi:
LGTM. Feel free to merge.

PR #1005, opened 2015-04-15
Fixed the signum function as if it receives an undefined number it crash...

local-girl:
This is on hold as (a) we don't need the same function in 3 places and (b) Perhaps this was a friendly error as values added to signum shouldn't be undefined, could make it harder to debug.


src/meinberlin/meinberlin/static/js/Adhocracy.ts
```diff
@@ -141,7 +141,6 @@ export var init = (config : AdhConfig.IService, meta_api) => {
             }]
         });
         $translateProvider.preferredLanguage(config.locale);
-        $translateProvider.fallbackLanguage("en");
```
local-girl: This is weird, is not missing in my code and should not be.

xi:
I will close this for now. We can reopen if this becomes relevant again.

PR #1007, opened 2015-04-15
Remove hardcoding of canonical url in the  export_mercator_proposal

pallix:


MagdaN:
works fine on my local machine, lets see if it works on server

PR #1009, opened 2015-04-15
2015 04 cc call to action2 rename

local-girl:
renamed class renamed m-call-to-action2 to m-call-to-action-secondary and added docs


xi:
LGTM.

PR #1011, opened 2015-04-15
update CSS related gems

xi:
Nothing much. There is not new sass release yet, so this is only about scss-lint and hologram.


nidico:
LGTM.

PR #1013, opened 2015-04-15
update javascript libraries

xi:
I updated some Javascript libraries to their newest versions. Notable exceptions:

-   moment.js: The dates were rendered in chinese. They had a big refactoring in version 2.10.0 so maybe they need some more time to stabilize. We can look into this sometime else.
-   protractor: I was not invloved in protractor integration yet, so I did not want to break something.
-   mailparser/sync-exec: I didn't even know we had these.

### DIFF
--- a/src/adhocracy_frontend/checkcode_and_compile.cfg
+++ b/src/adhocracy_frontend/checkcode_and_compile.cfg
@@ -198,11 +198,11 @@ update-command = ${:command}
 recipe = bowerrecipe
 packages =
     jquery#1.11.2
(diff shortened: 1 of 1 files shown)


pallix:
mailparser is used by the acceptance tests to check emails are correctly send.
sync-exec is used in etc/protractorConf.js to start supervisord and remove the test database once finished.

nidico:
`protractor` needs some dedicated effort, as version 2 doesn't simply work without adaptations.

nidico:
LGTM.

PR #1014, opened 2015-04-16
Disable rating and commenting in mercator backend

pallix:


joka:
lgtm

src/adhocracy_mercator/adhocracy_mercator/evolution/__init__.py
```diff
@@ -55,14 +55,38 @@ def evolve2_disable_add_proposal_permission(root):  # pragma: no cover
...
     set_acl(root, updated_acl, registry=registry)
-    # TODO substanced bug, _p_changed is not set
-    root._p_changed = True
 
     logger.info('Finished substanced evolve step 2:'
                 'remove add_proposal permission')
 
 
+def evolve3_disable_voting_and_commenting(root):
+    """Disable rate and comment permissions."""
```
joka: The text suffix evolve.. is not helping when reading the code. Its clear that this is a evolution step
anyway. The function name should express the one thing thate the function is doing.
The number suffix does not have a real meaning and makes merging more difficult

I propose we generelly remove the suffix for migration functions.

src/adhocracy_mercator/adhocracy_mercator/evolution/__init__.py
```diff
@@ -55,14 +55,38 @@ def evolve2_disable_add_proposal_permission(root):  # pragma: no cover
...
+    from adhocracy_core.utils import set_acl
+    from substanced.util import get_acl
+    from pyramid.threadlocal import get_current_registry
+    from pyramid.security import Deny
+
+    logger.info('Running substanced evolve step 3:'
+                'remove add_rate, add_rateversion, add_comment and'
+                'add_commentversion permissions')
```
joka: We repeat the numer and docstring here. That means if we change this we have to change it 
in 3 places. I propose to remove this comment completly or remove the number and reuse the docsting.

PR #1015, opened 2015-04-16
2015 04 tb map listing

MagdaN:


MagdaN:
@local-girl : applied filters should also be added to map-list. I tried that by adding a dummy method that removes an element from the list - it works, but the width on the list element somehow shrinks. Do you have any  idea, why?

local-girl:
I'm happy to look at this now...

MagdaN:
Can be reviewed

xi:
continued in #1026

PR #1017, opened 2015-04-19
Allow to disable rating in frontend

nidico:
Frontend part of #1014.

Same as in fb88c7c2d2e17aa187c54d0fc65c37015b3db67b, this is only a
temporary solution, which will be removed as soon as this can be covered
with workflows.


nidico:
Note that this is now deployed on advocate-europe.

xi:
LGTM.

PR #1022, opened 2015-04-21
permission-protect create_proposal route

xi:
This is very similar to how it is done on the mercator workbench.


MagdaN:
looks good to me

PR #1023, opened 2015-04-21
script to count 'lost' rates for older versions

joka:
If a user rates an old version, his rate is 'lost'
because we only count the rates for the latest version.


pallix:
LGTM

PR #1024, opened 2015-04-21
Tie up some loose ends

xi:
-   show rate count in proposal list item
-   do not throw warnings when signumFilter receives `undefined` (closes #1005)
-   remove noop location link


MagdaN:
looks good to me and works

PR #1025, opened 2015-04-21
2015 04 cc core moving columns sass change

local-girl:
I have updated core themes as previously links in the moving column menu were referred to as .moving-column-menu a, I have now give this 'a' its own class as it was causing compatibility problems in meinberlin and it is also better practice.


xi:
LGTM.

PR #1026, opened 2015-04-21
Combined map/listing

xi:
This is a cleaned-up version of #1015. I split some commits and regrouped them by topic. Unfortunately, the authorship information on some commits got lost.

On the one hand, a lot has happened in this branch. So I am generally in favor of merging it soon. But there are some tasks that still need to be done, in this or other branches.

-   The map should be initialized as few times as possible.

    -   That is why I reverted the switch from `ngShow` to `ngIf` in d79f75f4ee19cb1c24f436c78ab08963d4c69db5. But that has similar issues to those we dealt with in #1016.
    -   We should switch from using `adhRecompileOnChange` to a `$watch` based mechanism.

-   Next/Prev should have disabled states.

-   The case of 0 items should be handled.

-   Can the height of the map stay fixed or should it react to available space?


xi:
I talked to @MagdaN about the recompile issue:

On the one hand, recompile may take up a lot of resources. It also looses the widget state (e.g. the selected proposal) on each compilation.

On the one hand, it is very easy to implement compared to other possible solutions. In addition, it may actually take up less resources in some situations where the map is not visible but still constantly updating.

A third idea (besides recompile and watch) is to feed the list of items to `ngRepeat` and to create and delete markers based on some events which it exmits (at least we hope that there are such events).

The strategy for now is to get this merged as soon as possible and work on the other aspects separately.

xi:
I just realized that when clicking "test filter", no item from the non-map listing is removed. That might be a bigger issue. Further investigation needed.

MagdaN:
I also think we should merge this - but there are several issues to keep in mind that need to be fixed/implemented:
- when panning the map so that no markers are shown list and arrows should disappear or should be disabled
- when column is closed and reopened map needs to adjust to that
- for items without location the backend returns x=0 and y=0. Could this lead to problems in the future? At the moment a marker is created with those coordinates and the code works, because  the marker is not in the current map view

PR #1027, opened 2015-04-22
highlight selected list item

xi:
just as it is done in mercator. We should write some abstraction for
this.

### DIFF
--- a/src/meinberlin/meinberlin/static/js/Packages/MeinBerlin/Kiezkassen/Proposal/ListItem.html
+++ b/src/meinberlin/meinberlin/static/js/Packages/MeinBerlin/Kiezkassen/Proposal/ListItem.html
@@ -1,4 +1,4 @@
-<a href="{{ path | adhResourceUrl }}" class="mein-berlin-kiezkassen-proposal-list-item">
+<a href="{{ path | adhResourceUrl }}" class="mein-berlin-kiezkassen-proposal-list-item {{selectedState}}">
     <h1>{{ data.title }}</h1>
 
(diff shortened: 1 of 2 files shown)


src/meinberlin/meinberlin/static/js/Packages/MeinBerlin/Kiezkassen/Proposal/ListItem.html
```diff
@@ -1,4 +1,4 @@
-<a href="{{ path | adhResourceUrl }}" class="mein-berlin-kiezkassen-proposal-list-item">
+<a href="{{ path | adhResourceUrl }}" class="mein-berlin-kiezkassen-proposal-list-item {{selectedState}}">
```
local-girl: Any reason this is on the link and not the LI ? I would put it on the LI

xi: The LI is part of the listing which is a very generic widget. It would be more difficult to put this class on the LI because 

-  you would have to somehow get the specific code into the generic listing widget
-  you would have to use child selectors like `.is-selected .mein-berlin-kiezkassen-proposal-list-item` to apply specific styling.  See https://smacss.com/book/applicability why this is problematic

Alternatively, we include the LI in the specific list item template instead of the listing itself. But that does not work well with angular (if you use `ngRepaet` as an attribute it needs an element) and would require to repeat the LI every time you create a new list item.

I agree that this is not totally intuitive. But it was the best solution I could think of.

local-girl:
LGTM

PR #1029, opened 2015-04-22
use adhRecompileOnChange on proposal detail

xi:
This fixes an issue where the rate object did not update when the
proposal changed.

One solution was to only wrap the rate in adhRecompileOnChange. I choose
to wrap the whole proposal detail in order to avoid similar issues in
the future.

Something similar is also used in mercator.


nidico:
LGTM.

PR #1020, opened 2015-04-20
2015 04 cc moving columns styling

local-girl:

PR #1030, opened 2015-04-22
2015 04 cc moving columns styling+1

local-girl:
Continued from #1020


xi:
Looks great! See  my comments above. Also, the user columns need to be updated, too.

local-girl:
Don't know what `moving-column-expand` is please explain tomorrow

local-girl:
I added the tabs to user space but without text for now as this is core, not sure how to target meinberlin or if we need to

xi:
> I added the tabs to user space but without text for now as this is core, not sure how to target meinberlin or if we need to

I had thought that we had two versions of these columns.

I think it is not urgent. But I see two ways in which this could be done:

-   overwrite column templates in meinberlin
-   add text in template, hide via CSS in core

xi:
> Don't know what `moving-column-expand` is please explain tomorrow

It is the class you added in 95a5e6a

local-girl:
OMG I suck, doing this now

local-girl:
It is only referenced in the meinberlin _moving_columns_theme.scss - does it need to be in mercator?

local-girl:
Done

PR #1032, opened 2015-04-22
Added .moving-column-menu-nav-button where required, made some changed t...

local-girl:
Unfortunately my branch #1025 had some unexpected impact to the styling of mercator which I have now fixed up.


xi:
I think the name `moving-column-menu-nav-button` does not really fit for the links where you added it now. There should be a second class (maybe `moving-column-menu-button`?) for these links because they are not related to column navigation but to something that happens *within* a column.

xi:
Some more things I saw when looking through mercator platform:

-  The column close button has some spacing to the right.
-  The column close icon has turned from white to dark grey
-  The column close icon hover has turned from violet to dark grey

xi:
Continued in #1036.

PR #1033, opened 2015-04-22
configure default space paths

xi:
Very similar to mercator. Mazbe this is not urgent, but simple to do and
makes testing easier.


local-girl:
LGTM

PR #1035, opened 2015-04-22
2015 04 cc mercator print stylesheet

local-girl:
Some stuff to make mercator proposal printing pretty good, some core stuff too

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
@@ -1,5 +1,9 @@
 .main-header {
     position: relative;
+
+    @media print {
+        display: none;
+    }
 }
 
 .l-header {
@@ -29,7 +33,6 @@
 
 .l-header-center {
     left: 0;
-    position: absolute;
     text-align: center;
     top: 0;
     width: 100%;
@@ -41,7 +44,6 @@
 }
 
 .l-content {
-    position: absolute;
     top: $l-header-height;
     bottom: 0;
     left: 0;
@@ -54,20 +56,17 @@
     }
 }
 
+.l-header-center,
+.l-content {
+    position: absolute;
+
+    @media print {
+        position: static;
+    }
+}
+
 .l-center {
     max-width: 30em;
     margin: 5em auto 10em;
     padding: 0.5em;
-}
-
-
-@media print {
-    .l-header-center,
-    .l-content {
-        position: static;
-    }
-
-    .main-header {
-        display: none;
-    }
 }
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
@@ -47,6 +47,11 @@
     left: 0;
     right: 0;
     overflow: hidden;
+
+    @media print {
+        position: static;
+        overflow: visible;
+    }
 }
 
 .l-center {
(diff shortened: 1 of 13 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_moving_columns.scss
```diff
@@ -431,4 +439,11 @@ A list of alert messages to notify users in the moving column
     .moving-column-alerts {
         display: none;
     }
+
+    .moving-columns[class^="is-show-show"],
+    .moving-columns[class*=" is-show-show"] {
+        .moving-column.is-show:first-child {
+            display: none;
+        }
+    }
```
xi: This will break in the following case: 

-   have a small window (only one column)
-   navigate to the comment column
-   uncollapse the first column
-   print

This will only print the second column while you only see the first one. The reason for this is that the moving column state is actually `is-show-show-hide` but `focus` is set to the first column.

Note that moving column state and focus are not well defined and are likely to change.

I am also not sure whether this is a generic solution that should be in core.

My suggestion would be to add a `has-focus` state to moving columns via JavaScript and then only show the focused column on print.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_print.scss
```diff
@@ -0,0 +1,26 @@
...
+}
+
+@media print {
+    .print-only {
+        display: block;
+    }
+
+    .screen-only {
+        display: none;
+    }
```
xi: I don't like utility classes like these because they mix up content and presentation (classes should say what something is, not how it should be displayed).

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_print.scss
```diff
@@ -0,0 +1,26 @@
...
+
+    * {
+        // scss-lint:disable ImportantRule, ColorVariable
+        background-color: #fff !important;
+        color: #000 !important;
+        // scss-lint:enable ImportantRule, ColorVariable
+    }
```
xi: Something like this has already been added to the end of `_type.scss` in #997.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_print.scss
local-girl: I only use the utility classes in the case we need a special print header (which we do) in which case I think it's valid.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_moving_columns.scss
local-girl: Hmm OK I'll look in to it.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_print.scss
local-girl: Weird I saw a lot of grey text I thought, will remove one

xi:
In #997 I added print styles by adding a `@media print` block to the end of some scss files. You grouped these blocks with the related widgets instead. I think your approach is more in line with our code style. Could you please convert the older print blocks to your style?

xi:
Looks really impressive over all. Can't wait to get this merged. If we manage to maintain this well we will have great print versions for all time!

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
```diff
@@ -41,28 +44,29 @@
...
     right: 0;
     overflow: hidden;
-}
 
-.l-center {
-    max-width: 30em;
-    margin: 5em auto 10em;
-    padding: 0.5em;
+    @media print {
+        position: static;
```
xi: This is now duplicated in line 61

xi:
LGTM. I am still not happy with the utility classes but that's not that important.

@local-girl can you please review my changes and merge?

local-girl:
LGTM

PR #1036, opened 2015-04-23
Added .moving-column-menu-nav-button where required, made some changed t.

xi:
*Continued from #1032*


xi:
looked through the code with @local-girl

PR #1037, opened 2015-04-23
update mercator translation

xi:
This updates the translation files for core and mercator and translates all new strings. I also changed some new strings that were similar to existing ones. We may have to check for each case how this works out.

I did not yet push to transifex.


local-girl:
LGTM

PR #1043, opened 2015-04-27
improve HTML title

xi:
we did neglect the `<title>` in adhocracy for some time now. This pull requests sets a sane default and uses `site_name` from the frontend config once that is available.

In the future we might want to hook the title into routing, but that is not required just yet.

Note that `site_name` is probably not set properly on all our production setups. @slomo can you please take care of that?


nidico:
LGTM.

nidico:
> @slomo can you please take care of that?

Done in #1048.

PR #1047, opened 2015-04-27
2015 04 cc mercator print stylesheet further

local-girl:
Unfortunately I couldn't rebase to revert the font size change as I accidentally put some print header image sass in the original commit.


xi:
LGTM.

PR #1048, opened 2015-04-27
Use same key for site_name in frontend and backend config

nidico:


nidico:
Merging myself cause trivial.

PR #1049, opened 2015-04-27
2015 04 mn add button on zoomout

MagdaN:
Still needs styling.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
@@ -277,10 +277,12 @@ export interface IMapListScope<T> extends angular.IScope {
     visible: number;
 }
 
-export var mapListingInternal = (adhConfig : AdhConfig.IService,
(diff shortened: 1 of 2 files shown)


xi:
Works fine.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
```diff
@@ -413,6 +428,17 @@ export var mapListingInternal = (adhConfig : AdhConfig.IService, adhHttp : AdhHt
...
+
+            scope.resetMap = () => {
+                map.fitBounds(scope.polygon.getBounds());
+
+                // this is hacky but I could not find how to add
+                // a callback function to fitbounds as this always
+                // triggers the moveend event.
+                $timeout(() => {
+                    scope.showZoomButton = false;
+                }, 300);
```
xi: Maybe we could use [`LatLngBound.contains(<LatLngBounds>)`](http://leafletjs.com/reference.html#latlngbounds) in the moveend handler in order to only set this to `true` if necessary.

PR #1050, opened 2015-04-27
2015 04 cc meinberlin map switcher

local-girl:


MagdaN:
works nicely

src/meinberlin/meinberlin/static/js/Packages/MeinBerlin/Kiezkassen/Kiezkassen.ts
```diff
@@ -24,7 +25,9 @@ export var detailDirective = (adhConfig : AdhConfig.IService, adhHttp : AdhHttp.
                 heading: "Tab2",
                 content: "foo2"
             }];
-
+            scope.$watch(() => column.$scope.shared.isShowMap, function(value) {
```
xi: I think it would be better to use `column.$scope.$watch("shared.isShowMap", ...)`. Also this should be lambda syntax instead of function syntax.

PR #1051, opened 2015-04-27
Embed contexts

xi:
This is a first step of severl canges to adhEmbed that aim at allowing navigation in embedded widgets. This pull request introduces "contexts" to the embed API. A context is basically the whole platform (already availbe on the SDK level as "plain" widget). In the future, different context are supposed to have different functionality.


nidico:
LGTM.

PR #1052, opened 2015-04-27
Hide user profile links if not available

xi:
if no user profiles are available on a platform (e.g. embedded) we should not link to that profile. Thsi pull requests implements that by using information that is already available in adhResourceArea.


nidico:
LGTM.

PR #1053, opened 2015-04-27
route based on embed context

xi:
The embed contexts introduced in #1051 are here put to use in order to have separate routing and a separate template on `adhResourceArea`.

Unfortunately, my local unittest setup seems to be broken and I was not able to execute the tests.


nidico:
LGTM, didn't try to use it though. I fixed some unit tests.

PR #1054, opened 2015-04-28
Configure Kiezkassen Context

xi:
this actually uses the new features introduced in #1051, #1052 and #1053.

If you go to `/embed/kiezkassen` you will now have a navigatable platform that is different from the default one.

The configuration is mostly copied from Workbench.ts for now. But I expect that they will diverge significantly in the future.

Note that the kiezkassen context does not have a user space. Therefore the adhUserMeta directives do not render links (see #1052).

My unittest setup is still broken. I promise I will investigate that very soon. But for now there might by some failing tests that I did not see.

An unfortunate sideeffect of not encoding the embed context in the url is that it will be lost on reload which is quite annoying for manual testing. Note however that you can set an initial Url via `/embed/kiezkassen?initialUrl=...`.

A next step could be to implement the header that should be used in this context.


nidico:
Works nicely. I'm not very happy about the redundant configuration between workbench and embedded, but I'm unsure whether it's worth abstracting this. Merging.

xi:
> I'm not very happy about the redundant configuration between workbench and embedded, but I'm unsure whether it's worth abstracting this.

I really think they will diverge. For example, on the platform everything will be pushed one column to the right due to the browsing column

PR #1055, opened 2015-04-28
2015 04 cc register ux

local-girl:
Added register UX so now you can not register if you are logged in (threw a backend error before)

You should get messages if you try to register and you are (a) already logged in (b) just registered and logged in.

Needs DE translations, on Transifex


xi:
Works fine, though not in chrome due to #1056. However, there is still some stuff to do:

-   The `success && loggedIn` case should not tell me to close the tab (see next bullet point). Instead it should offer a link that takes me back to the place I started at.
-   On activation, I should see a page telling me to close this tab (basically what is now the `success && loggedIn` case).
-   The same needs to be implemented for password reset.

src/adhocracy_frontend/adhocracy_frontend/static/i18n/core_en.json
```diff
@@ -111,6 +113,10 @@
     "TR__REGISTER_SUCCESS": "Thank you for your registration. We sent you an email with an activation link. In case you do not recieve that email, please also check your spam folder.",
     "TR__REGISTRATION_LOGIN_INSTEAD": "I already have an account. Log in [[link:here]].",
     "TR__REGISTRATION_SUPPORT": "Having trouble with the registration?",
+    "TR__REGISTRATION_ALREADY_LOGGED_IN": {
+        "text1": "You are already logged in, you must",
+        "text2": "to re-register."
+    },
```
xi: The system introduced in #789 should be used instead.

xi:
We also need a fallback for people who register from an iframe and have third party cookies disabled. They won't get logged in automatically and therefore need a way to do it manually.

local-girl:
I need to hand this over now. @xi can you explain this properly to @nidico please? Basically if a user has just activated and has the registration page still open then need a message saying thanks for registering, they can now close the tab. I got as far as making it so they could not attempt to register if they were already logged in (as any user) with a message to log out of they wish to re-register. DE translations are still missing.

xi:
Should be complete now.

-   I did not yet push to transifex
-   I did not think about UI yet. Probably needs some polishing.
-   I did not yet think about the case that local storage is not available. IMHO that is low-prio.

local-girl:
For me in Chrome I registered, then opened a new tab to activate. There was no confirmation message in my activation tab and previous tab just gave "Already logged in" message before redirecting back home

local-girl:
This is now working as expected except when I try to logout after successfully changing password & getting the success message, I am directed back to the "Your password reset email has been sent" page. Logout/in should be disabled on these informational pages? Merging anyway.

PR #1058, opened 2015-04-28
Migrate/Refactor permissions

pallix:


src/adhocracy_core/adhocracy_core/authorization/__init__.py
```diff
@@ -87,3 +87,26 @@ def get_local_roles_all(resource) -> dict:
...
+        for permissions in acm['permissions']:
+            permission_name = permissions[0]
+            if permission_name not in registry.content.permissions:
+                raise ValueError('Invalid permission: {0} not in {1}'
+                                 .format(permission_name, registry.content.permissions))
+            action = permissions[idx + 1]
+            if action is not None:
+                ace = (action, principal, permission_name)
+                acl.append(ace)
+        idx = idx + 1
```
joka: Deny actions should be append before Allow to allow overriting Allows

src/adhocracy_core/adhocracy_core/content/__init__.py
```diff
@@ -83,6 +83,39 @@ def resources_meta_addable(self):
             resources_addables[iresource] = all_addables
         return resources_addables
 
+    @property
+    def permissions(self) -> [str]:
+        """Set of all permissions defined in the system."""
+        perms = {'hide_resource', 'edit_group', 'do_transition'}  # built-in permissions
```
joka: if you extract this to a property its easier to extand and you don't neet the comment to explain the meaning.

src/adhocracy_core/adhocracy_core/resources/root.py
```diff
@@ -10,75 +9,52 @@
...
 from adhocracy_core.resources.principal import IGroup
+from adhocracy_core.authorization import acm_to_acl
 import adhocracy_core.sheets.principal
 import adhocracy_core.sheets.name
 
-# Permissons to roles mapping (ACL)
+
+# Access Control Matrix. Permissions are mapped to a role.
 # Every role should only have the permission for the specific actions it is
 # meant to enable.
```
joka: "acm" should be explained in the glossary (docs/glossary)

src/adhocracy_core/adhocracy_core/resources/root.py
```diff
@@ -10,75 +9,52 @@
...
-            (Allow, 'role:admin', 'edit_sheet'),  # edit sheets default
-            (Allow, 'role:admin', 'manage_principals'),
-            (Allow, 'role:admin', 'view_userextended'),
-            (Allow, 'role:admin', 'edit_userextended'),
-            # God role
-            (Allow, 'role:god', ALL_PERMISSIONS),
-            ]
-# TODO? every sheet/resource should have the posibility to add the needed ACEs
+root_acm = \
+    {'principals':                                   ['system.Everyone', 'role:annotator', 'role:contributor', 'role:creator', 'role:manager', 'role:admin', 'role:god'],  # noqa
```
joka: the term "acm" should be explained in the glossary (docs/glossary)

src/adhocracy_core/adhocracy_core/resources/root.py
```diff
@@ -10,75 +9,52 @@
...
+                     ['hide_resource',                 None,              None,             None,               None,           Allow,          None,         Allow],  # noqa
+                     ['manage_principals',             None,              None,             None,               None,           None,           Allow,        Allow],  # noqa
+                     ['message_to_user',               None,              None,             Allow,              None,           None,           None,         Allow],  # noqa
+                     ['view',                          Allow,             Allow,            Allow,              None,           Allow,          Allow,        Allow],  # noqa
+                     ['view_sensitive',                None,              None,             None,               None,           None,           Allow,        Allow],  # noqa
+                     ['view_userextended',             None,              None,             None,               Allow,          None,           Allow,        Allow],  # noqa
+                     ]}
```
joka: this looks so much nicer now

src/adhocracy_core/adhocracy_core/authorization/__init__.py
```diff
@@ -87,3 +87,26 @@ def get_local_roles_all(resource) -> dict:
...
+
+def acm_to_acl(acm: dict, registry: Registry) -> [str]:
+    """Convert an Access Control Matrix into a pyramid ACL.
+
+    To avoid generating too many ACE, action which are None will not
+    generate an ACE.
+
+    """
+    acl = []
+    idx = 0
```
joka: We should use the ACM Schema to validate the given acm.

joka:
To ease migrating the acl we should add a console script or autoupate changed acl on application startup.

src/adhocracy_core/adhocracy_core/authorization/__init__.py
pallix: Is it really necessary as the override will be done with two ACMs and not inside the same ACM? For one ACM we don't have a Deny overriding a Allow for the same principal, permission couple.

pallix:
What do you mean with "autoupdate changed acl on startup"? Do we need something more than setting the ACL on startup as I did in the mercator.py file?

pallix:
The validation for ACM is done when they are created. It's not integrated in the `acm_to_acl` function, as the function receives an appstruct from the workflow module and the appstruct has already been validated by Colander.

pallix:
Code ready for a second review :-)

joka:
python tests are failing

pallix:
Ok tests are running again. We can merge once we:

- fix permission for 'do_transition' in root.py (which principals are allowed to do a workflow transition?)
- fix permission in mercator.py to avoid overriding the Deny which were setup with the migration script

pallix:
I updated the mercator ACM and marked failing functional test with `mark.xfail` decorators. Can you check and merge?

Issue #1028, opened 2015-04-22
meinberlin specific code in adhocracy_core


xi:
in #1026 we unfortunately committed code to adhocracy_core that should really be in meinberlin. A workaround was pushed in 89dc894. But we need a real fix.

It would be a shame to move the whole mapListing to meinberlin. A better option would be to split the meinberlin specific stuff into a subwidget and use transclusion (just as it is done with listings).

xi:
The bigger part of this has been fixed in #1059. However:

> The list item was split from the listing code itself in order to make the listing reusable. Unfortunately we did not manage to transclude it twice (adhMapListing and adhMapListingInternal) so it is still in the wrong place.

PR #1059, opened 2015-04-28
refactor map listing

xi:
This refactors the map listing in order to make it easier to change the list of show items on runtime.

The list item was split from the listing code itself in order to make the listing reusable. Unfortunately we did not manage to transclude it twice (`adhMapListing` and `adhMapListingInternal`) so it is still in the wrong place.

The new idea is that the list of paths is passed to `ngRepeat` and the list items register themselves on the listing. This way we can use all the power of `ngRepeat` for watching the input and concentrate on displaying stuff on the map.

I did not yet implement an initial selection. We will also have to think a bit more what happens if the list changes. The selected item might be removed or its index might change. We then have to select a new item and scroll to the right position. That should be done in a separate pull request.


xi:
Will take care of the merge conflict tomorrow.

xi:
related to #1028.

PR #1062, opened 2015-04-29
participation icons

xi:
This adds css for participation icons ("dot icons").

The comment gap icons still needs to be fixed. But I though we could merge this already anyway.


nidico:
Looks nice, though I'd prefer to have @local-girl give it a look as well.

local-girl:
LGTM although the dots arent completelt centralised, we have made the designer aware of this and can change it later.

PR #1065, opened 2015-04-30
2015 04 jk add organisation

joka:


src/adhocracy_core/adhocracy_core/resources/asset.py
```diff
@@ -161,6 +162,9 @@ class IPoolWithAssets(IPool):
     """A pool with an auto-created asset service pool."""
 
 
+deprecated('IPoolWithAssets', 'BBB use process instead')
```
pallix: what does 'BBB' means? Bla bla bla?

PR #1068, opened 2015-05-04
Move kiezkassen frontend to use actual backend process

xi:
After #1065 there is now an actual kiezkassen process resource in the backend. I migrated all frontend code to use that. Among other things, this implements `getProcessType()` from #884.

You may have to run `bin/sd_evolve etc/development.ini` to actually create the backend resources.


nidico:
LGTM.

PR #1069, opened 2015-05-05
2015 05 cc tabs extra features

local-girl:
Some extra options for tabs. Ability to add extra classes + "Closed by default"

PR #1071, opened 2015-05-05
2015 05 cc phase header ui+3

local-girl:
Navigate to localhost:6551/embed/kiezkassen?initialUrl=/r/organisation/kiezkasse/ to see this

Please merge #1069 first


xi:
I will continue working on this. Specifically, I will work on the following:

-   Context header and phase information should be two separate directives
-   the tabs that are used for phase navigation should be derived from regular tabs instead of overwriting the template directly
-   AFAIK the phase information will not be translated, at least not using the same mechanism
-   will fix the "add proposal" button
-   the "login" button needs to change to "logout" after login

xi:
When #1079, #1080 and #1082 have been merged, most of this should be done. What is left to do is:

-   phase information. This does not only contain plain text but also dates and available features (see #1062).
-   the phase information widget is not a plain tabs widget. The additional features need to be implemented
    -   highlight current phase
    -   separate buttons to trigger filters (@nidico do we need these already?)

PR #1077, opened 2015-05-06
layout rows

xi:
This allows to have multiple rows where the last one stetches to the end of the screen.

We can use this to have variable-height content on top of the moving columns.


local-girl:
LGTM possibly needs some browser testing IE / Safar etc but we can do it another time

PR #1079, opened 2015-05-06
kiezkassen header

xi:
This is a first part that was split out of #1071.

It implements a header for the kiezkassen context.


local-girl:
LGTM

PR #1080, opened 2015-05-06
Add phase tab modifer

xi:
this is a second pull request that was extracted from #1071. It adds a `m-phase` modifier for tabs.


local-girl:
LGTM

PR #1083, opened 2015-05-06
2015 05 cc standalone header

local-girl:
Still not 100% sure which pages need the header and which not.. please take a look


xi:
Looks great!

I have two remarks, though:

-   there should be some documentation on when to use `l-center` and when to use `standalone-wrapper`. You replaced one by the other in some places so it is obvious that they are similar.

-  In the register flow, when I got back to the initial tab after closing the one that I opened from the link in the email, I got a `standalone-wrapper`. I had expected a `l-center` there.

xi:
continued in #1086

PR #1085, opened 2015-05-06
Hide user profile links in indicator if not available

xi:
Fixup to #1052


nidico:
Works and looks good. Some unit tests fail, but this is not due to this branch.

PR #1086, opened 2015-05-06
2015 05 cc standalone header+1

xi:
continued from #1083


xi:
Looked at this with @local-girl

PR #1089, opened 2015-05-07
2015 05 mn edit proposal

MagdaN:
still missing:
-  [x] cancel button redirects to proposal
- [x] show edit button only if user has edit rights
- [x] update context.ts
- [x] change title of create.html based on whether it is used to create or edit


MagdaN:
still missing:
[] cancel button redirects to proposal 
[] update context.ts

nidico:
LGTM, works nicely.

PR #1093, opened 2015-05-07
Add a script to create process for region

MagdaN:
The script can be used to add processes based on a bezirksregion. Then run:

 bin/create_process_for_region path_ to_development.ini name_ of_ the_ bezirgsregion name _of_organisation

eg. bin/create_process_for_region etc/development.ini sudliche-luisenstadt organisation

NOTE: in order to get that working you have to add the bezirke first and after that the bezirksregion:

required: GDAL (Version 1.10.1) is installed.

First run: bin/import_bezirke ./etc/development.ini
Second run: bin/import_bezirksregions ./etc/development.ini


nidico:
Works nicely!

Issue #1094, opened 2015-05-08
Use MultiPolygons in frontend


nidico:
Our backend stores locations as multipolygons. However our frontend code (e.g. `Mapping.ts`) always deals with simple polygons without holes (i.e. two array levels less). This should be aligned to MultiPolygons.

MagdaN:
We probably just need to change to leaflet.multipolygons: http://leafletjs.com/reference.html#multipolygon

xi:
[Wikipedia](https://en.wikipedia.org/wiki/GeoJSON#Geometries) has a good overview of geometries in GeoJSON. The short summary is: Polygons can have holes, Multipolygons can contain multiple disjoint polygons.

The leaflet documentation on [polygons](http://leafletjs.com/reference.html#polygon) mentiones that "You can also create a polygon with holes by passing an array of arrays of latlngs, with the first latlngs array representing the exterior ring while the remaining represent the holes inside". There is no such note for Multipolygons, but I guess it applies there as well.

Unfortunately, DefinitelyTyped does not support these optional holes yet.

PR #1095, opened 2015-05-08
Use actual polygons from process instead of hardcoded

nidico:
Not nice:

- Crashes if no polygon is set (though this is also the case in other
places)
- Mapping in frontend only works with simple polygons instead of
MultiPolygons (can be fixed separately - see #1094)


MagdaN:
@nidico : this works nicely - I am just wondering if we should merge it, as it will crash everybody's frontend code and it would take a while for them to get it working angain (e.g. install gdal, run all the scripts etc.).

MagdaN:
@joka : could you add a polygon to the sample kiezkassen process that get autmatically generated as soon as the database gets created?

joka:
> @joka : could you add a polygon to the sample kiezkassen process that get autmatically generated as soon as the database gets created?

done

PR #1098, opened 2015-05-08
Do not depend meinberlin on adhocracy_sample

nidico:
This makes wheel creation work again.


nidico:
Merging myself. @joka, is this alright?

xi:
I think we depended on this because of some tests. Will check...

xi:
See #922. @nidico can you please reevaluate the situation again?

PR #1099, opened 2015-05-08
Adjust AdhUtil.parentPath docstring to actual functionality

nidico:
This function only works with paths, not with URLs.

I considered modifying the function itself accordingly, but this would
not be the expected behavior when dealing with backend resuorces, as
they may have a SCRIPT_NAME (e.g. `/api`) as part of the URL.

Related to #757.


xi:
LGTM.

PR #1101, opened 2015-05-08
Load maps with https

nidico:


nidico:
Merging myself, cause trivial.

PR #1100, opened 2015-05-08
Raise error if /batch tries to invoke subrequests outside script_name

nidico:
If this would have been there, #1102 would have been much easier to solve.

PR #1102, opened 2015-05-08
Use proper resource URLs in batch request in get_process

nidico:
Otherwise we run into trouble when script_name is set (see #1100).


nidico:
Merging myself.

PR #1104, opened 2015-05-11
2015 05 cc map list tweaks

local-girl:
Tidying horizontal map listing to match design, changed HTML of vertical listing to be almost the same.


xi:
-  The listItem templates should be merged
-  classes should be `list-item`, not `map-list-item`

xi:
@local-girl could you please review my changes?

local-girl:
LGTM

PR #1105, opened 2015-05-11
reference leaflet.css by absolute path (fixes #1097)

### LINKED ISSUE #1097: leaflet.css not loaded in production
As `leaflet.css` is imported relatively in `a3.scss`, this doesn't work when built with production settings, as this ends up in the `min` folder.

This leads to a completely broken map.

xi:
Note that leaflet.css is now also loaded for core and mercator.


nidico:
> Note that leaflet.css is now also loaded for core and mercator.

Not nice. We should change that later.

PR #1108, opened 2015-05-11
keep unittests from failing

xi:
These are just workarounds. It will be a bigger task to make our unittesting useful again.


nidico:
LGTM.

PR #1114, opened 2015-05-11
added watch statements to rebuild list Fixes #1107

### LINKED ISSUE #1107: Resizing of window creates bad layout in horizontal map widget
image

local-girl:
Added a watch to column width and column amount to trigger an list items reload. Looks a bit wonkey as it can only trigger after moving columns have loaded. Fixes #1107

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
@@ -327,19 +327,13 @@ export class MapListingController {
             this.scrollToItem(this.$scope.selectedPath, false);
         });
 
-        var mcs = $element.parents(".moving-columns");
-        this.$scope.$watch(() => {
-            return mcs.find(".moving-column.is-show, .moving-column.is-collapse").length;
-        }, () => {
+        this.$scope.$watch(() => $element.find(".map-list").width(), () => {
+            this.scrollToItem(this.$scope.selectedPath, false);
+
             // FIXME: moving column load time hard coded
             this.$timeout(() => {
                 this.scrollToItem(this.$scope.selectedPath, true);
-            }, 550);
-        });
-        this.$scope.$watch(() => {
-            return this.map.getBounds().getSouthWest().lng;
-        }, () => {
-            this.scrollToItem(this.$scope.selectedPath, false);
+            }, 500);
         });
     }
 
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
@@ -272,6 +272,7 @@ export class MapListingController {
     private selectedItemLeafletIcon;
     private itemLeafletIcon;
     private markers : {[path : string]: L.Marker};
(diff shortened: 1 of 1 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
```diff
@@ -324,6 +324,25 @@ export class MapListingController {
...
+        }, (newValue, oldValue) => {
+            if (newValue !== oldValue) {
+                // fixme: moving column load time hard coded
+                this.$timeout(() => {
+                    this.scrollToItem(this.$scope.selectedPath, true);
+                }, 550);
+            }
+        });
+        this.$scope.$watch(() => {
+            return this.map.getBounds()._southWest.lng;
```
xi: Should be `getSouthWest()`. We should not use private properties.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
```diff
@@ -324,6 +324,25 @@ export class MapListingController {
...
+        }, (newValue, oldValue) => {
+            if (newValue !== oldValue) {
```
xi: This is already done by `$watch`.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
```diff
@@ -324,6 +324,25 @@ export class MapListingController {
         this.$scope.$watch("items", () => {
             this.scrollToItem(this.$scope.selectedPath, false);
         });
+
+        var mcs = $element.parents(".moving-columns");
+        this.$scope.$watch(() => {
+            return mcs.find(".moving-column.is-show, .moving-column.is-collapse").length;
+        }, (newValue, oldValue) => {
+            if (newValue !== oldValue) {
+                // fixme: moving column load time hard coded
```
xi: capital FIXME

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Mapping.ts
```diff
@@ -324,6 +324,25 @@ export class MapListingController {
         this.$scope.$watch("items", () => {
             this.scrollToItem(this.$scope.selectedPath, false);
         });
+
+        var mcs = $element.parents(".moving-columns");
+        this.$scope.$watch(() => {
+            return mcs.find(".moving-column.is-show, .moving-column.is-collapse").length;
```
xi: why not simply watch `$element.width()`?

xi: another option would be to use `adhTopLevelState.on("movingColumns")`.

xi:
I am not so sure about the watch expressions. Will have a closer look tomorrow.

xi:
I changed the watch statement. @local-girl could you please review?

Some context on why I think this is better:

-   does not reply on columns. So it also works if we use this widget in another context, e.g. on its own.
-   more generic: Similar could be reused in other places.
-   simpler / less code.

local-girl:
Yeah that's much better, merged!

PR #1115, opened 2015-05-11
Fix layout rows in IE11

xi:
Layout rows have been introduced in #1077 and basically are a hack. This makes them work in IE11 by adding yet another level of nesting.


MagdaN:
Looks good

PR #1117, opened 2015-05-11
add 'super_types' for sheets/resource types to meta_api API CHANGE

joka:


nidico:
What's the purpose of super types? Are they documented somewhere?

xi:
I discussed this briefly with @joka. I would like to have this in order to be able to identify process resources. See #1068 / ff68613 for reference. With this change we could do something like

```
if (RIProcess.content_type in resource.super_types) {
    ...
}
```

nidico:
Thanks for the explanation!

Issue #1113, opened 2015-05-11
Embeded page does not work in IE11


MagdaN:
I tried to look at the emebeded page in IE11 and it looked like that:
![screenshot - 11 05 2015 - 16 34 12](https://cloud.githubusercontent.com/assets/2905743/7567166/98935edc-f7fb-11e4-8aac-95965acceb62.png)

xi:
These are two issues (as far as I can see ATM):

-  layout rows are not rendered properly (fixed in #1115)
-  tab pane height does not work

MagdaN:
Now the above shown bug is fixed but when switching between list and map view there is always the list item for 'TR__MAP_NO_ITEMS_IN_VIEW' row is shown.

xi:
I cannot really work on the `TR__MAP_NO_ITEMS_IN_VIEW` issues until #1119 is fixed.

xi:
> Now the above shown bug is fixed but when switching between list and map view there is always the list item for 'TR__MAP_NO_ITEMS_IN_VIEW' row is shown.

This has a separate issue at #1116.

PR #1121, opened 2015-05-12
Fix tab size detection in IE

xi:
*Fixes part of #1113*.


MagdaN:
This only works if you open and close the same tab. But if you open one tab and then click on the next tab, that maybe has more text in it then the first one, then the height is not adapted to the new tab.

xi:
Works for me.

xi:
Pushed a fixup in 9d3df01

PR #1122, opened 2015-05-12
2015 05 cc comments ui

local-girl:
Improved styling for commenting to match current designs


src/meinberlin/meinberlin/static/stylesheets/scss/components/_comment_theme.scss
```diff
@@ -1,13 +1,80 @@
...
+}
+
+.comment-main {
+    margin-bottom: 24px / $font-size-normal * 1em;
+    padding-bottom: 0;
+}
+
+.rate .rate-difference {
+    @include icon-button;
+    margin-right: -2px;
```
xi: This is strange because this element does not have any click functionality. SO why should it be a button? I also think that this rules should be in rate.scss.

local-girl: according to the docs its just on this page

src/meinberlin/meinberlin/static/stylesheets/scss/components/_comment_theme.scss
```diff
@@ -1,13 +1,80 @@
...
+    margin-right: 1px;
+    padding: 0;
+
+    .comment-header-link {
+        @include icon-button;
+        margin: 0 -1px 0 0;
+    }
+
+    .comment-header-link-text {
+        @include squish-text;
```
xi: I don't like this because it does leave most users with only an icon, i.e. without a tooltip. Can we maybe change it? Have a `title` attribute in the HTML and use `:after { content: attr(title) }` in the core CSS?

xi:
LGTM

PR #1138, opened 2015-05-13
2015 05 cc phase header ui tweak

local-girl:
Mde phase header more like current designs. Commit marked 'Gardening' is mislabeled! Would re-base but millions of commits from master now. Merge mast commit also has a few fixups in it as there was a conflict, best to check Files Changed and not per commit.


src/meinberlin/meinberlin/static/stylesheets/scss/components/_phase_tabs.scss
```diff
@@ -1,51 +1,111 @@
...
+
+        .tab-button, &.is-active .tab-button {
+            background: transparent;
+            border-bottom: 2px solid $color-background-base;
+            color: $color-text-highlight-introvert;
+            font-weight: $font-weight-extrovert;
+            padding: 0;
+
+            &:after {
+                background: url("/static/images/chevron.png") no-repeat right center;
```
local-girl: Yes I tried to do this pure CSS, gave up.

xi:
I do not like the approach this pull request takes to style the phase navigation. Before this, there was a clear separation between the tabset widgets and the phase modifier. This clear separation is blurred  by this pull request. We will need to restore a clear separation.

local-girl:
Be my guest!

PR #1139, opened 2015-05-13
flexbox layout row

### LINKED ISSUE #1112: 404 does not work in embed in Firefox
when visiting a page that returns 404 - the 404 gets not rendered propoerly in Firefox, check out here: http://downloads.liqd.net/website-mock/berlin-mock.html#!/r/organisation/biesdorf/BlaBla/VERSION_0000001/

When clicking on back, also the rest of the page is not rendered properly anymore.

### LINKED ISSUE #1116: IE11: when switching between list and map view there is always the list item for 'TR__MAP_NO_ITEMS_IN_VIEW' row is shown.

xi:
this adds a flexbox implementation for layout rows (see #1077). The old table layout is still used if flexbox is not available. This should fix #1112.


MagdaN:
Works

PR #1126, opened 2015-05-12
create phase widget with metadata

xi:

PR #1144, opened 2015-05-14
phase header directive

xi:
This splits the phase header into its own directive and allows to highlight the currently active phase.

While working on this I realized that there have been some issues with #1138. For example, it reverted parts of #1126. So I had to fix this first.


nidico:
LGTM and works, but @local-girl should also have a look.

We should add a phase name to the `phases` object which correlates what comes from the backend and match `currentPhase` with that one. Can be done separately though.

PR #1146, opened 2015-05-16
2015 05 jk update roles permission

joka:
- cleanup permissions

- add system principal Anonymous

- add roles participant, moderator, initiator

-  rm roles reader, contribuor, annotator


nidico:
Is there a reason to write `Anonymous` and `Everyone` in capital initial letters, but not the other roles (including `god`!)?

PR #1148, opened 2015-05-17
Edit Kiezkassen process

nidico:
Add a simple frontend for Kiezkassen process editing.

This includes

* changing the process title,
* performing workflow transitions.

Future work:

* add more properties (e.g. description texts)
* make code reusable for other processes. ideally generate form from `meta_api`


MagdaN:
works

PR #1152, opened 2015-05-18
fix input background in standalone header

xi:
see for example in password reset flow (after you clicked the link in the email).

The inputs were gray on grey before.


local-girl:
LGTM

PR #1153, opened 2015-05-18
Add BPlan Statement Form

MagdaN:


nidico:
I wonder if we shouldn't call the resource `Stellungnahme` instead of `Proposal`.

MagdaN:
Should we switch between englisch and german - so we should switch to statement or something @local-girl : do you have an idea?

xi:
> I wonder if we shouldn't call the resource Stellungnahme instead of Proposal.

I guess this can be changed later. Merging now to avoid conflicts.

PR #1154, opened 2015-05-18
Update translations

xi:
this updates the translations. The version proposed here differes significantly from the one currently on transifex because there was some confusion about du/Du/Sie in german.

I would like to get this reviewed quickly so I can push this version to transifex.


nidico:
LGTM.

PR #1156, opened 2015-05-18
Reflect Kiezkassen workflow state in frontend

nidico:
see commits


xi:
LGTM. A bit hard to test currently, but at least the phase seems to be ok.

An issue is that the kiezkassenDetailColumn has other stuff that should be changed in the announce phase. For example, there is the switch between list an map in the column header. Maybe we should just use a different column?

xi:
LGTM.

PR #1158, opened 2015-05-18
item/version routing

xi:
this adds functionality to `adhResourceArea` to easily register routes for versionable resources.

The assumption is that we always need an item and a version for routing in these cases. If the URL points to an item, the newest version is used.

This does not yet implement any links to the newly created item routes. I will do that in a separate pull request.


xi:
Edit: I pushed a commit that uses the new routes.

nidico:
LGTM, works.

PR #1159, opened 2015-05-18
2015 05 cc meinberlin 404

local-girl:


src/meinberlin/meinberlin/static/stylesheets/scss/components/_error_pages_theme.scss
```diff
@@ -0,0 +1,11 @@
+.error-page-404 {
+    .error-page-image {
+        max-width: 300px;
```
xi: I would switch this to `width: 300px` so that the `max-width: 100%` is not overwritten.

xi:
I only have a tiny remark (see comment). Apart from that, feel free to merge.

local-girl:
I agree, merging

PR #1162, opened 2015-05-19
http: noCredentials option

xi:
This adds an option `noCredentials` to (nearly) all http methods. This will overwrite the headers to not send authentication credentials on this request.

To avoid code duplication I refactored the http methods to take a generic `config` map similar to the one in `$http`. I hope that this will also be easy to extend in the future.


nidico:
Code looks good and works.

I like the `config` parameter, as it makes the passed parameter explicit in the function call.

PR #1164, opened 2015-05-19
Reflect permissions in frontend

xi:
with this, some functionality is visually disabled if a user is logged in but still does not have the required permissions. This is the case e.g. if the action is not permitted in the current workflow state.


nidico:
This is now broken after #1175.

nidico:
Works nicely.

The process phase is currently not watched, but that can be done at another time.

PR #1166, opened 2015-05-19
change map input button semantics

xi:
-   save now happens implicitly
-   "clear" is now roughly what cancel did before
-   "reset" is a 1-level undo


MagdaN:
Looks good and works

nidico:
Nice!

PR #1167, opened 2015-05-19
2015 05 mn render markdown

MagdaN:
This code adds a directive that renders markdown to html and I also added a directive for you to test it. Use buildout for meinberlin and then go to http://localhost:6551/embed/test-parse-markdown and start typing to the textfield. 

I used this library https://github.com/markdown-it/markdown-it#usage-examples it does not render script tags to html.


xi:
LGTM.

PR #1172, opened 2015-05-20
Fix #1168, remove meinberlin evolution script to create sampel content,

### LINKED ISSUE #1168: meinberlin evolution script tries to create object twice
Running `bin/sd_evolve --latest etc/development.ini` on an empty database fails, as example location `alt-treptow` and example organisation `organisation` already exist.

joka:
the sampe content objects are created twice, with the migration step and
with the after_create function for the root resource.


nidico:
Fine.

PR #1165, opened 2015-05-19
2015 05 cc meinb form button tidy

local-girl:
Lots of commits but fairly minor stylistic updates to all forms and buttons in Meinberlin

PR #1173, opened 2015-05-20
merge #1165/#1166

xi:
there were some semanatic (non-git) conflicts between #1165 and #1166.
This resolves them.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_map.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_map.scss
@@ -11,7 +11,8 @@ category: Widgets
         @include clearfix;
     }
 
-    .map-button-cancel {
+    .map-button-clear,
+    .map-button-reset {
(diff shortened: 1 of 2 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_map.scss
```diff
@@ -11,17 +11,7 @@ category: Widgets
...
-            color: $color-brand-two-introvert;
-        }
-
-        &:active {
-            color: $color-brand-one-introvert;
-        }
-    }
-
-    .map-button-save, .map-button-cancel {
+    .map-button-cancel {
```
local-girl: This class does not appear in any code?

xi: You are right. I thought it should be used for the two new buttons. Could you please fix this?

local-girl:
LGTM, merging

PR #1175, opened 2015-05-20
make http requests wait for credentials

xi:
this makes sure that http requests are stalled until the credentials have been loaded from localStorage.

You may see some OPTIONS requests without the token headers. We think that these are preflight requests issued by the browser.


nidico:
Code looks good and works. As this is a major refactoring and our acceptance test suite doesn't pass currently, some things might be broken, but we'll notice soon...

PR #1183, opened 2015-05-21
2015 05 mn update translation file and create script

MagdaN:
This creates a new core_de.json or meinberlin_de.json with 'Sie' instead of 'Du'

Run by:  
bin/change_german_salutation <path to file>/core_de.json

creates a new file in the same folder


xi:
LGTM.

PR #1191, opened 2015-05-26
2015 05 mn edit process fields

MagdaN:
could have been more generic, but this was faster.


nidico:
Works. I got red fields from the angular validation system on the first try, but the changes got saved anyway. As this is only for "internal" use at the moment, we don't need to investigate that any further.

PR #1194, opened 2015-05-26
Refactor workbench

xi:
this is some refactoring of the frontend workbench code. The main purpose is to come up with a convention for having views with same name but different contentTypes.


nidico:
LGTM.

PR #1198, opened 2015-05-26
translation system vor backen, custom translation for meinberlin

joka:


xi:
transifex integration might be good as well

nidico:
I'd do transifex only once we really do translations in the backend, i.e. backend knows the user's locale and sends emails in that locale.

PR #1202, opened 2015-05-27
2015 05 cc login form tidy+1

local-girl:


src/meinberlin/meinberlin/static/stylesheets/scss/components/_login_theme.scss
```diff
@@ -12,9 +12,23 @@
...
+}
+
 .login-cancel {
     font-variant: normal;
     font-size: $font-size-small;
     margin-top: 0;
     text-transform: capitalize;
 }
+
+.user-form-wrapper {
```
xi: needs to be a modifier

xi:
I think the approach taken here does not scale. We somehow need a reusable header, not just in terms of CSS. I doubt that we can theme the header by pure CSS because a header may contain a logo (which is probably not always monochromatic). What makes things worse is that the header sometimes containes additional stuff (e.g. the "create proposal" button both in mercator and in kiezkassen).

local-girl:
I feel this works for what we have now and we can change it in the future when additional options are needed ?

xi:
I had issues with headers in another branch. Will talk to the designers and try to understand what they want.

xi:
replaced by #1226

PR #1205, opened 2015-05-27
BPläne feedback view

xi:
this continues on #1153 by adding a feedback view and some other details. Notably API communication, UI and translations are missing. I plan to keep working on this and create several pull requests in order to not end up with one giant one.


joka:
Should I merge #1212  so you can work with the real backend?

xi:
> Should I merge #1212 so you can work with the real backend?

This pull request is independent of that. But yes, a working backend would be great so we can implement the communication part.

MagdaN:
looks good - I will do the transaltions in a seperate pull request

PR #1211, opened 2015-05-27
2015 05 jk fix user is missing default group

joka:


nidico:
This indeed fixes the issue. One functional test fails, but I'm merging anyway now.

xi:
Which issue does this fix?

nidico:
When two persons registered soon after each other, the second person isn't in the `authenticated` group and thus lacks permissions.

Issue #1061, opened 2015-04-29
cameFrom broken


xi:
There is some code in the frontend (`adhTopLevelState`) that is supposed to redirect users back to the place they came from after login/register. Sometimes this works, but far from reliable. We should check why it is so unreliable and fix it.

xi:
This issue is very vague. We should get a better understanding of the concrete issues.

PR #1213, opened 2015-05-28
deny setCameFrom on paths that are likely to be redirected to

xi:
this fixes the following issue:

-   go to the login page
-   click on register in the user indicator
-   click on cancel

expected result: go back to the initial page

actual result: go back to login


xi:
related to #1061

nidico:
Doesn't compile.

nidico:
LGTM.

PR #1217, opened 2015-05-28
2015 05 cc mein ui context header tweaks

local-girl:
Added a gap under context header and made phase navigation text smaller


xi:
LGTM.

PR #1226, opened 2015-05-28
Added meinberlin specific user header

local-girl:


src/meinberlin/meinberlin/static/js/Packages/User/UserManagementHeader.html
```diff
@@ -0,0 +1,8 @@
+<header class="l-header main-header m-user-management">
+    <div class="l-header-left">
+        <i class="mein-berlin-keizkassen-context-header-logo icon-meinberlin-logo" title="mein Berin"></i>
```
xi: typo: "mein Berlin"

xi: I don't really like the `mein-berlin-keizkassen-context-header-logo` in this context. But we can refactor that separately.

PR #1230, opened 2015-05-28
Check values when using import script

pallix:


pallix:
I got an error in `test_import_resource_create_group` since IGroup deserialize with an empty `users: []` list we get `colander.Invalid: {'users': 'This field is ``readonly``.'}` when creating the group.

I tried setting `missing = colander.drop` but it breaks other tests and I'm not sure of the implication, so I did not commit this change.

pallix:
Also got an `colander.Invalid: {'location': 'This resource path does not exist.'}` on this data 

`{"path": "/treptow-koepenick",
     "creator": "god",
     "content_type": "adhocracy_core.resources.organisation.IOrganisation",
     "data": {"adhocracy_core.sheets.name.IName":
              {"name": "kiezkassen"}
             }     
    },
    {"path": "/treptow-koepenick/kiezkassen",
     "creator": "Name",
     "content_type": "adhocracy_meinberlin.resources.kiezkassen.IProcess",
     "data": {"adhocracy_core.sheets.name.IName":
              {"name": "alt-treptow-2015"},
              "adhocracy_core.sheets.geo.ILocationReference":
              {"location": "/locations/alt-treptow"},
              "adhocracy_core.sheets.title.ITitle":
              {"title": "Kiezkasse Alt-Treptow 2015"}}
    }
`

Do path to other resources aren't resolved by colander?

joka:
> I got an error in test_import_resource_create_group since IGroup deserialize with an empty users: [] list we get colander.Invalid: {'users': 'This field isreadonly.'} when creating the group.
> 
> I tried setting missing = colander.drop but it breaks other tests and I'm not sure of the implication, so I did not commit this change.

the  code is deserializing the cstruct  twice, thats way the 'user' field appears and raises an error

joka:
> Do path to other resources aren't resolved by colander?

can you check if its now working?

joka:
The tests pass now,  I also refactored and reordered the functions, so the diff is a bit confusing.

pallix:
I don't see where I did the deserialize twice but OK.

PR #1232, opened 2015-05-29
2015 05 jk badges

joka:


nidico:
Refactoring looks good. Badge modelling is fine for now, but I'm not happy with the badge assignment modelling. Instead of adding the assigned objects to the badge, I'd rather use a separate resource `BadgeAssignment`, similar to what is done with ratings, because:

- The assignee doesn't need edit permission on the `Badge` resource itself. This would be a problem as different roles might create and assign badges, and they may also be in different contexts.
- Accumulating all badged objects in the badge resource itself might turn the badge resource very large, in particular for globally used badges.

nidico:
One backend unit test (bplan something) fails.

joka:
master merged, test fixted,
Still todo : badge groups and search by badge

PR #1229, opened 2015-05-28
new translations

MagdaN:

PR #1233, opened 2015-06-01
Fixup "new translations"

xi:
This is a fixup to #1229. There was some gardening to do and one translatable string was missing.


xi:
already pushed to transifex

PR #1223, opened 2015-05-28
Created user management header directive, added to core pages

local-girl:

PR #1219, opened 2015-05-28
Document headers

xi:
*related to #1202*

I will refactor the existing headers to fit with this concept later. This pull request is only about documenting. Please do not merge before the designers had a chance to review it.

PR #1235, opened 2015-06-01
mein berlin headers

xi:
This continues work from #1219 and #1223. There is now some more structure to headers:

-   There is the default header defined in `TopLevelState/templates/Header.html`

-   The default header (as part of the adhTopLevelState template) can be overwritten in embed contexts. This is done for example for kiezkassen:

    ```
    adhResourceAreaProvider
        .template("kiezkassen", ["adhConfig", "$templateRequest", (
            adhConfig : AdhConfig.IService,
            $templateRequest : angular.ITemplateRequestService
        ) => {
            return $templateRequest(adhConfig.pkg_path + pkgLocation + "/template.html");
        }]);
    ```

-   There is a separate header for user management areas in `User/UserManagementHeader.html`

-   There is a separate header for the embed area in `Embed/Header.html`


nidico:
LGTM.

PR #1236, opened 2015-06-01
BPlan UI

xi:
This implements the UI for the embedded BPlan proposal. Some notes:

-   I am not too convinced of my SCSS code. I propose to merge this now, but @local-girl should still take a look when she is back.
-   I guess we should use the `nocenter` embed parameter with this
-   The button does not yet look like a button. This is because the buttons do not look like in the design in general. This will be taken care of separately.
-   AFAIK there is currently no way to get to the success screen because the request always fails. So you will have to hack around that when doing review.


MagdaN:
looks good, except for the case when an error is shown - then title of the next field is pushed to the side.

PR #1240, opened 2015-06-01
2015 05 jk fix errors when creating pblan proposals

joka:


nidico:
Works! We should remove that workaround when refactoring the zero-version away.

PR #1244, opened 2015-06-02
2015 06 mn add spd files

MagdaN:


nidico:
I'm not happy with the lowercase file names (`spdDocument`), but there's no clear solution. It's a mess, which we should clean up another time.

nidico:
LGTM.

PR #1246, opened 2015-06-02
2015 06 cc 404 image move+1

local-girl:
Move "mercator" 404 image from mercator to core as we have a broken image on new installations now. Either a core 404 image has to be made, or there is no image by default, meaning we would have to make theme files for 404 from now on. Not sure the best course of action.


xi:
I think it would be better to have a more neutral 404 image in core, especially in terms of color.

local-girl:
I agree but this is better than a broken image while we wait, have informed designers

PR #1247, opened 2015-06-02
SPD UI theme

xi:
this implements the UI theme for spd. That is, it fixes some things in the mercator theme, copies most of that to spd and changes some variables.


local-girl:
LGTM - Mercator is broken currently so I couldn't have a good look at it but it's just styling so it's not a big deal.

PR #1248, opened 2015-06-02
install fonts from bower

xi:
this removes Lato files from our repo and instead installs it from bower. It also adds Open Sans which is used in SPD.


local-girl:
LGTM - works

PR #1251, opened 2015-06-02
Fix missing eggs folder

nidico:
The eggs folder is sometimes created too late, which causes issues. So
we created it manually.

This seems to be a timing issue, as `bin/buildout -vvv` mitigates the
problem.


local-girl:
LGTM

PR #1253, opened 2015-06-02
spd list item

xi:
continues on #1252


MagdaN:
I get an error in the frontend: Error: [$injector:unpr] Unknown provider: adhImageUriFilterProvider

xi:
Which buildout did you use? What do I have to do to reproduce that error?

MagdaN:
Works now!

PR #1254, opened 2015-06-02
save full url, not only path in cameFrom

### LINKED ISSUE #1250: embed parameters are lost on navigation
## Steps to reproduce

-   Go to https://meinberlin-stage.liqd.net/bplan.html
-   Click on "login"
-   Clock on "cancel"

## Expected result

-   Go back to original page

## actual result

`nocenter` and `path` parameters are gone.

xi:
fixes #1250


nidico:
LGTM.

PR #1256, opened 2015-06-03
Make sample platform a IProcess

nidico:
This should fix the create-or-show-comment-listing widget, which is used
in core acceptance tests and in PolicyCompass.


MagdaN:
The frontend core tests are now running again

PR #1259, opened 2015-06-03
rm integration tests

xi:
these have not been used for a long time and are broken all the time.

I do think that integration tests are useful. But in their current form
they do more harm than good.

@nidico already agreed that we should remove them.


nidico:
LGTM.

PR #1260, opened 2015-06-03
Do not ignore all __init__ files for code coverage

pallix:
I was too liberal in my usage of stars for the coverage configuration! We are only at 99% after this patch. We can merge this first and fix the coverage later.


joka:
oh no, the poor coverage

PR #1263, opened 2015-06-03
2015 06 pa add mercator workflow

pallix:
Acceptance tests (vote, comment) and one functional tests are still broken on this branch.


pallix:
`test_cannot_edit_other_users_proposal` still failing after second merge

pallix:
@joka can you have a look?

pallix:
@joka I found the issue! PR ready for review.

pallix:
after_creation are not runned properly... investigating..

pallix:
http://localhost:6541/mercator is not accessible as anonymous:

403 Forbidden
Access was denied to this resource.
Unauthorized: PoolRESTView failed permission check

Not sure if this is what we want

nidico:
As just discussed: we need to perform a transition to the `announce` or `participate` phases in order to make it visible in the frontend.

pallix:
I'm done with the fixes.

pallix:
Ok unit tests, func tests and acceptance tests which success on master, success again

src/adhocracy_mercator/adhocracy_mercator/resources/root.py
```diff
@@ -19,6 +21,21 @@ def add_mercator_process(context: IPool, registry: Registry, options: dict):
...
+    root.__acl__ = [(Allow, 'role:god', ALL_PERMISSIONS)]
+    request = Request.blank('/dummy')
+    request.root = root
+    request.registry = registry
+    request.__cached_principals__ = ['role:god']
+    mercator_process = root['mercator']
+    workflow = registry.content.workflows['mercator']
+    workflow.initialize(mercator_process)
+    workflow.transition_to_state(mercator_process, request, 'announce')
+    workflow.transition_to_state(mercator_process, request, 'participate')
```
joka: The proper way would be to change the inital state for the mercator workflow. I propose we: 
* move this code to a helper function (we will need this again)
* add field 'state_initial' to the workflow schema  to set the inital state instead of "missusing"  the state_order field.

src/adhocracy_mercator/adhocracy_mercator/workflows/test_init.py
```diff
@@ -0,0 +1,168 @@
...
+            self, app_broken_token):
+        resp = _batch_post_full_sample_proposal(app_broken_token)
+        assert resp.status_code == 400
+
+
+@mark.functional
+class TestMercatorWorkflow:
+
+    ## the state is currently set to 'participate' when creating the mercator process
+
```
joka: instead of commenting add  mark.xfail with and explanation.
We should also extend the workflow api to reset the the state to the the inital state

pallix: Ok for tests but I would prefer the workflow api to be extended in a separate branch.

pallix:
@joka can you have a look at the refactoring?

joka:
lgtm

PR #1272, opened 2015-06-08
2015 06 cc spd chapter styles

local-girl:
Markdown H2 headers in SPD detail form will be styled exactly as chapter headers, this involved de-nesting a core style. Also tidied up jump nav


xi:
I am not sure whether we need it, but it works.

local-girl:
Toby says LGTM

PR #1275, opened 2015-06-08
2015 06 cc mercator badges ui+1

local-girl:
Im a little unclear exactly how this will eventually work, whether with a class modifier (how I've done it) or with a physical element, either way this will be very easy to change when the time comes.

Also unclear about whether this should be in core, very easy to move

Test by adding classes 
```has-badge m-winner```
or
```has-badge m-community-award```
to mercator-proposal-list-item


xi:
LGTM. I would like some documentation, especially about the fact that you need non-static positioning on the container. Feel free to merge once that has been added.

local-girl:
Merged

PR #1281, opened 2015-06-09
fix angular boostrap

xi:
in an effort to increase acceptance test stability I did a bit of reading on the `angular never provided resumeBootstrap` error.

I found https://github.com/angular/protractor/issues/1124 which was interesting but did not really provide a solution.

However, I noticed that we passed `document` as a first parameter to `angular.bootstrap()`.  This is invalid because a DOMElement is required there.

I changed it to `document.body` because this is also the default value for [rootElement](https://github.com/angular/protractor/blob/master/docs/referenceConf.js) in the protractor config. This will probably not make much of a difference but does not hurt either.


MagdaN:
I don't have full understanding of this but on the angular bootstrap documentation it says: "Note that Protractor based end-to-end tests cannot use this function to bootstrap manually. They must use ngApp."

https://docs.angularjs.org/api/ng/function/angular.bootstrap

PR #1282, opened 2015-06-09
add document, paragraph resource type, remove sample_* resource types

joka:


xi:
travis reports 1 failing functional test (I think it is src/adhocracy_core/docs/badges.rst)

PR #1285, opened 2015-06-09
Added jury decision styles

local-girl:
See styleguide for docs


xi:
LGTM. Added some tweaks.

PR #1290, opened 2015-06-10
wire up jury statement

MagdaN:
at the moment the same design is shown for both winning projects and the project that won the community award - this can stay like that according to the designer


local-girl:
LGTM

PR #1292, opened 2015-06-10
reflect resource inheritance in frontend

xi:
This adds the `super_types` from the meta api to the generated frontend types. It also uses them in `adhResourceArea.getProcess()`.


nidico:
Code looks good, but I can't test it due to network problems where I am currently located. As travis passed once, it looks like things generally still work. According to @xi, it works.

PR #1293, opened 2015-06-10
Restore IPostPoolSheet and mark it as deprecated

pallix:
This is need for migration scripts which tries to work with pickled
value having this interface


joka:
thank you

PR #1295, opened 2015-06-10
Add mercator badge filters

xi:


xi:
@MagdaN reported that she had a blank screen when testing this. That is probably cause by the issue that is fixed in #1297.

xi:
I just noticed that bades have the `ITitle` sheet. Maybe it's better to use those rather than defining new translatable strings?

MagdaN:
@xi I had to merge master to get it working again - now it works fine

MagdaN:
@xi: how would you do translations, then?

PR #1296, opened 2015-06-10
Mercator migration add badge assignments service

pallix:
@joka the trick with the iteration over the sheets of the metadata is nice but then we will have to write the same code for every new sheet in a new evolution step each time we add a new sheet?


joka:
> @joka the trick with the iteration over the sheets of the metadata is nice but then we will have to write the same code for every new sheet in a new evolution step each time we add a new sheet?

if we add new sheet but the resource type is the same, we can just use the helper script in adhocrac_core.evolution
`migrate_new_sheet(root, IMercatorProposalVersion,  IBadgeable) `

joka:
lgtm, @pallix can you have a quick look at my changes and then merge?

PR #1297, opened 2015-06-10
Fix mercator workbench processType

xi:
With the merge of #1292 `/mercator` is recognized as a process by the frontend. So now it is required to define the routes in the context of this process.

I checked and I think this is the only process where this change is required.


xi:
Similar code for spd can be found in 14efbd5740c24e6cb4aa8ec551a1a5e005c2104e

local-girl:
Works for me

local-girl:
Does SPD need this fix too because it seems to have the same problem @xi ? @nidico ?

xi:
> Does SPD need this fix too because it seems to have the same problem

see https://github.com/liqd/adhocracy3.mercator/pull/1297#issuecomment-110827017

local-girl:
Oh oops

PR #1294, opened 2015-06-10
spd workbench

xi:
This implements the mercator workbench and the communication with the backend. Sorry for the large pull request.

Note that this is rebased against #1292 so that should be merged first.

Also note that due to a backend bug, the paragraphs are scrambled by the backend. Among other things, this breaks editing.

I did not yet spend much thought on workflow. So there might be some more work to do.

PR #1300, opened 2015-06-10
2015 06 cc proposal list item core style

local-girl:
Moves styling of proposal list item to core. Effects SPD was started in #1294 and cherry picked in to here. Also affects Mercator - not sure about MeinBerlin though as the proposal list is so different, then it's weird not to call it "proposal-list-item", thoughts ?


xi:
@local-girl could you please merge master?

xi:
Looks good in mercator and meinberlin. We will have to check how it works with spd when merging with #1294.

PR #1305, opened 2015-06-11
Code for meinberlin badges UI

local-girl:
See documentation at 2015-06-cc-meinberlin-badges (after build)

### DIFF
--- a/src/meinberlin/meinberlin/static/stylesheets/scss/components/_badges_theme.scss
+++ b/src/meinberlin/meinberlin/static/stylesheets/scss/components/_badges_theme.scss
@@ -1,0 +1,31 @@
+/*doc
+---
+title: badges
+name: badges
+category: Type
+---
+
+Badges for labeling if a proposal has been realized or not.
+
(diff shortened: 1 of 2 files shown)


src/meinberlin/meinberlin/static/stylesheets/scss/components/_badges_theme.scss
```diff
@@ -0,0 +1,31 @@
...
+```
+
+*/
+.has-badge {
+    @include inline-block;
+    color: $color-text-inverted;
+    font-size: $font-size-small;
+    font-variant: small-caps;
+    padding: 0 5px;
+    text-transform: uppercase;
```
xi: The combination of `small-caps` and `uppercase` does not make much sense. I propose to skip `small-caps`.

xi: Talked to designers and changed in 2c34156

xi:
LGTM.

PR #1306, opened 2015-06-12
hide data in result phase

MagdaN:


src/mercator/mercator/static/js/Packages/MercatorProposal/MercatorProposal.ts
```diff
@@ -170,6 +171,9 @@ export interface IScopeData {
 
     // 7. badges
     assignments : BadgeAssignment[];
+
+    // 8. current phase
+    currentPhase: string;
```
xi: The comments (7, 8) do not make sense here. The above ones relate to the sections where these variables are used. I would move `assignments` and `currentPhase` to the top.

xi:
LGTM

PR #1307, opened 2015-06-14
Fix: order of references is not preserved

joka:


src/adhocracy_core/adhocracy_core/catalog/__init__.py
```diff
@@ -87,8 +83,13 @@ def _search_elements(self, query) -> IResultSet:
             allowed_index = self._get_index('allowed')
             principals, permission = query.allows
             index_query &= allowed_index.allows(principals, permission)
-
         elements = index_query.execute(resolver=None)
+        if query.references:
+            index = self._get_index('reference')
+            for reference in query.references:
+                referencences = index.search_with_order(reference)
```
pallix: typo or wrong name

PR #1310, opened 2015-06-15
Fixed up styles for create form

local-girl:
Create form now looks better. Involved fixing some core styles I will check in mercator / meinberlin bt I can't imagine a conflict


local-girl:
Does not seem to have effected the other sites

local-girl:
OK done

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_forms.scss
```diff
@@ -336,5 +336,6 @@ A generic form footer containing one CTA and a regular button
 
     .form-footer-button {
         float: left;
+        margin-right: 5px;
```
xi: What is this for?

PR #1314, opened 2015-06-15
Improve code coverage in authentification module

pallix:
@joka it's not clear to me why you added these lines since they are not used by the tests. Maybe a later refactoring made them unnecessary.

Since this code is not executed, the code coverage is not 100% for this file.

Can you have a look and merge or just close the request if ther is a reason to have that?


joka:
> @joka it's not clear to me why you added these lines since they are not used by the tests. Maybe a later refactoring made them unnecessary.
> 
> Since this code is not executed, the code coverage is not 100% for this file.
> 
> Can you have a look and merge or just close the request if ther is a reason to have that?

This was not only test related. If you run  a pyramid script there is no authentication and request.root is None.

joka:
I add a test an did a rebase

PR #1316, opened 2015-06-15
Fix code coverage for catalog/subscriber

pallix:
@joka Here also it's not clear to me why you added these lines since they are not used by the tests. Maybe a later refactoring made them unnecessary.

Since this code is not executed, the code coverage is not 100% for this file.

Can you have a look and merge or just close the request if ther is a reason to have that?


joka:
> Can you have a look and merge or just close the request if ther is a reason to have that?

If all test pass pass its fine.

PR #1323, opened 2015-06-16
Move spdDocument to core

xi:
originally I only wanted to do the translations. But I did not want to re-do them once this has been moved to core. So I just did the moving first.


local-girl:
LGTM

PR #1325, opened 2015-06-17
add missing abuse dependency

xi:
due to a missing angular dependency, the report form for comments was not rendered.


local-girl:
LGTM

PR #1327, opened 2015-06-17
Fix documentUrl on @comments route

xi:
-   create a document
-   edit the document and only add one paragraph
-   go to the comments of one of the initial paragraphs

the new paragraph is displayed

the new paragraph is not displayed

during routing, the wrong document version was selected

I also fixed a similar issue for the comment route in the same commit.


local-girl:
LGTM

PR #1328, opened 2015-06-17
sort spd document listing by name (not title!)

xi:


local-girl:
Seems to be in reverse, by design ?
image

xi:
I will check again

xi:
It turns out that the backend works as expected. Unfortunately, the "name" this index sorts by is the name of the *version* (e.g. `VERSION_0000003`), not of the *item*. So this can not be implemented by simply using an existing index.

PR #1331, opened 2015-06-17
Add adhocracy_sample to spd buildout config

slomo:


xi:
It looks to me like this is related to 41bad3db. Could you please extend that documentation?

xi:
Is this still relevant?

slomo:
I greped thorught the source and it seems not package other than sample is using stuff from sample. Also the setup.py of spd doesn't include sample.

Seems to be irrelevant.

PR #1332, opened 2015-06-17
Evolution script to set digital_leben workflow to 'participate' phase

pallix:


src/adhocracy_spd/adhocracy_spd/evolution/__init__.py
```diff
@@ -0,0 +1,22 @@
+"""Scripts to migrate legacy objects in existing databases."""
```
nidico: Docstring doesn't match content.

nidico:
I'd prefer to have this as a command line script, as - I assume - we don't want to evolve the production system to `participate` immediately.

pallix:
Having a command line script is probably a good idea, then developers can also play locally with different phases while developping.

nidico:
Is this obsolete now after merging #1340?

pallix:
Yep.

PR #1335, opened 2015-06-17
fix redirect after form success

xi:
The order in which resources are returned from `deepPost()` has been fixed in 223525ed. The redirect after creating/editing a MercatorProposal relied on the previous order. This should fix the regression.


nidico:
Works fine!

PR #1336, opened 2015-06-17
Clean up rate/like disable

xi:
this removes some hacks and adds a `is-disabled` class to like buttons.

Actually disabeling functionality and styling the disabled button is still missing.


MagdaN:
I tried this with my mercator-database being in resultphase and the button was shown and I after logging in I could rate - so I guess, that is more of a backend problem.

MagdaN:
button is still there and clickable but does not do anything

PR #1342, opened 2015-06-22
Allow to sort by title

nidico:
This adds a catalog index for `ITitle.title`.

@joka: As discssed, this doesn't work yet and yields something like:

    hypatia.exc.Unsortable: [-5337356932914536409]

Could you please check?

Also,

- tests are missing;
- maybe the sort request in the frontend needs to be in reverse.


joka:
> @joka: As discssed, this doesn't work yet and yields something like:
> 
> hypatia.exc.Unsortable: [-5337356932914536409]
> 
> Could you please chec

Testet with mercator, works fine.
Maybe you search did include resources not indexed by the title index.

nidico:
> Maybe you search did include resources not indexed by the title index.

True! The `IDocument` resource doesn't use `ITitle`, but a custom `SingleLine` field in the `IDocument` sheet. (Why?)

Due to my frontend commit f96297d, the digital_leben listing is now broken.

xi:
Just for reference: this continues on #1328.

PR #1346, opened 2015-06-22
Added disabled styles to like button

local-girl:


src/mercator/mercator/static/js/Packages/MercatorProposal/Detail.html
```diff
@@ -7,7 +7,7 @@
     }">
     <h2 class="print-only section-jump-cover-header" data-aria-hidden="true">{{ data.title }}</h2>
     <div class="section-jump-cover">
-        <ul class="meta-list mercator-proposal-detail-meta-list">
+        <ul class="meta-list mercator-proposal-detail-meta-list m-phase-{{data.currentPhase}}">
```
local-girl: Perhaps this should be numeric in case we change the phase names. Fred already told that what we call "participate" they call "join"

local-girl:
@xi does the disabled assignment on the like button fit the phase. EG does
```!optionsPostPool.POST && optionsPostPool.loggedIn``` match up ```with currentPhase != 'participate'``` should we make our own switch for mercator so the like button is disabled depending on phase not postPool?

xi:
I guess that #1347 would fix that issue but it is postponed for now. So I agree using the phase information is the best way to go. You could for example revert 09bde8a6bc09ca6d344316c2738dbd69cad865d7 and use the `disabled`.

local-girl:
OK I think this is sorted now. I reverted the removal of the disable flag for Rate, added it to the Like template - now Mercator phases can use this to disable the like button based on phases too

MagdaN:
looks goog

PR #1360, opened 2015-06-23
2015 06 mn spd add userbadges

MagdaN:
I am going to wire up the ui from @local-girl as well, and then you can review

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
@@ -10,6 +10,12 @@ import AdhTopLevelState = require("../TopLevelState/TopLevelState");
 import AdhCredentials = require("./Credentials");
 import AdhUser = require("./User");
 
+import SIBadgeable = require("../../Resources_/adhocracy_core/sheets/badge/IBadgeable");
+import SIBadgeAssignment = require("../../Resources_/adhocracy_core/sheets/badge/IBadgeAssignment");
+import SIDescription = require("../../Resources_/adhocracy_core/sheets/description/IDescription");
+import RIBadgeAssignment = require("../../Resources_/adhocracy_core/resources/badge/IBadgeAssignment");
+import SIName = require("../../Resources_/adhocracy_core/sheets/name/IName");
+import SITitle = require("../../Resources_/adhocracy_core/sheets/title/ITitle");
 import RIUser = require("../../Resources_/adhocracy_core/resources/principal/IUser");
 import RIUsersService = require("../../Resources_/adhocracy_core/resources/principal/IUsersService");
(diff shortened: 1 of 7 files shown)


MagdaN:
This can be reviewed now - if you need a database, let me know. The badge jsons need to changed due to changes in another branch, but I will do this in the other branch (https://github.com/liqd/adhocracy3.mercator/pull/1363).

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
```diff
@@ -352,12 +358,36 @@ export var metaDirective = (adhConfig : AdhConfig.IService, adhResourceArea : Ad
...
+        return adhHttp.get(assignmentPath).then((assignment : RIBadgeAssignment) => {
+            var description = assignment.data[SIDescription.nick].description;
+            return adhHttp.get(assignment.data[SIBadgeAssignment.nick].badge).then((badge) => {
+                var title = badge.data[SITitle.nick].title;
+                var name = badge.data[SIName.nick].name;
+                return new BadgeAssignment(title, description, name);
+            });
+        });
+    }));
+};
```
xi: This code is duplicated in `MeinBerlin/Kiezkassen/Proposal/Proposal.ts` and in `MercatorProposal/MercatorProposal.ts`. Please move it to a single place.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
```diff
@@ -430,6 +464,9 @@ export var userProfileDirective = (
                 adhHttp.resolve(scope.path)
                     .then((res) => {
                         scope.userBasic = res.data[SIUserBasic.nick];
+                        getBadges(adhHttp, $q)(res).then((assignments) => {
+                            scope.assignments = assignments;
+                        });
```
xi: I think this should be named `badges`, not `assignments`. I think we should rename it here and also in the other places where `getBadges()` is used.

xi:
The bades should also be shown in the `adhUserMeta` directive.

xi:
Apart from the comments above this looks good.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
MagdaN: I would like to do that in another branch, as it effects all badges not just user badges

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
MagdaN: I would like to do that in another branch, as it effects all badges not just user badges.

xi:
LGTM.

PR #1368, opened 2015-06-24
backend test coverage back to 100%

joka:


src/adhocracy_core/adhocracy_core/resources/principal.py
```diff
@@ -290,7 +290,7 @@ def get_user_by_activation_path(self, activation_path: str) -> IUser:
         """Find user per activation path or return None."""
         users = self._get_users()
         for user in users:
-            if user.activation_path == activation_path:
+            if user.activation_path == activation_path:  # pragma: no branch
```
pallix: i don't see why the branchment would not be executed

Issue #1359, opened 2015-06-23
set_workflow_state script has a useless parameter


pallix:
The workflow parameter is useless in the set_workflow_state since it can be retrieved from the resource.

PR #1369, opened 2015-06-25
Refactor setup workflow function and script

pallix:
This solve #1359. 

It should be merged **after the SPD release**, to avoid changing the prod config at the last moment.

Once merged @slomo should update the salt configuration for the calls to `setup_worklow_state` script. Expected parameters are:

<ini-file> <resource-path> <state>


pallix:
@joka ready for review

pallix:
@slomo please update the installation setup

PR #1370, opened 2015-06-25
Refactor pool autonaming

pallix:
Each prefix gets an associated number which get automatically
incremented. This avoid having, for example, a document having two
paragraphs getting the name VERSION_0000001 and a second document
VERSION_000005 because of the (previously implemented) shared counter.


src/adhocracy_core/adhocracy_core/resources/pool.py
```diff
@@ -80,6 +79,15 @@ def add_next(self, subobject, prefix=''):
     def _zfill(self, name):
         return str(int(name)).zfill(self._autoname_length)
 
+    def _get_next_number(self, prefix):
+        if prefix in self._autoname_lasts:
```
pallix: Here I wanted to use `defaultdict` but I had a problem with pickle

pallix:
Maybe a migration script is required also?

joka:
> Maybe a migration script is required also?

for all pools we shoud:

      - add self._autoname_lasts = {<prefix>: last_count} as needed
      - remove attribute   _autoname_last

pallix:
@joka ready for review

PR #1372, opened 2015-06-27
2015 06 jk customizable invitation mail

joka:


nidico:
Works nicely. The broken acceptance test is also broken in master.

However we should change a couple of things in the future:

- We should require the invitation templates to be pyramid assets, because:
    - Users of the script (be it admin or a future frontend implementation) should be able to fully customize & personalize the text (decreases participation barrier!) without updating the repo
    - Invitation texts aren't public in general, as they're only sent to a limited set of persons.
- Once this is done, invitation templates shouldn't be mako templates in order to avoid potential arbitrary code execution.

PR #1374, opened 2015-06-27
independed image upload directive

xi:
This duplicates some code from `MercatorProposal.ts` for uploading images. My first idea was to do a proper abstraction and refactor the mercator specific code. Unfortunately I did not have the time to finish this yet.

This part already implements a widget that can be used to upload an image and reference it from any compatible resource (one that has the `IVersionable` and `IImageReference` sheets). It can be used via the embed area.

Note that for some reason there are permission errors when trying to access an uploaded image. @joka can you please have a look?


nidico:
Works nicely.

For the reference, you need to use a URL such as the following in order to add images:

> http://localhost:6551/embed/upload-image?pool-path=http:%2F%2Flocalhost:6541%2Fdigital_leben%2F&path=http:%2F%2Flocalhost:6541%2Fdigital_leben%2Ftitle_12%2FVERSION_0000002%2F

PR #1379, opened 2015-06-29
Fix: Email subject shall not contain newlines

nidico:
However files in POSIX typically end with newlines (because each *line*
has to end with a newline character).


nidico:
Merging myself, cause urgent.

MagdaN:
looks good

Issue #1382, opened 2015-06-29
Import user script allow to import already existing user


pallix:
Import script allows to import user with its email even if a user with the same email already exists

PR #1385, opened 2015-06-30
Allow import user script to modify user names

pallix:
Also it updates user if either the login or the email already exists (previously the check was done only on the login name)


pallix:
@slomo you can update user names with this script once merged

pallix:
Fixes #1382

joka:
one backend test is failing

pallix:
Tests running ok after merge of master

pallix:
@joka can you merge?

PR #1387, opened 2015-06-30
2015 05 jk fix password resets are visible

joka:


pallix:
@slomo @nidico should be tested on staging

PR #1388, opened 2015-06-30
Fix core frontend: Don't export non-existing module

nidico:
Note that the core frontend can currently only be used in embedded
mode.


nidico:
Merge

nidico:
Merging myself.

xi:
I don't get this. The module you removed *does* exist. So why should it be removed?

PR #1392, opened 2015-07-01
hide proposal list for spd users

MagdaN:
This hides the proposal list for users in the spd project as users never have such a list


nidico:
LGTM.

PR #1393, opened 2015-07-01
add  and use logbook resource type for mercator

joka:
Mercator proposals have now a 'IHasLogbookPool' sheet pointing to the pool to post 
documents.

Missing:   migration


src/adhocracy_core/adhocracy_core/sheets/logbook
```diff
@@ -0,0 +1,205 @@
+"""Badge sheet."""
```
pallix: This whole file needs to be deleted

PR #1394, opened 2015-07-02
Email confirmation bplan submission

pallix:
The works is not fully finished but can be reviewed

Missing:
- [x] full modelling of ProcessSettings. We still don't know which information are missing. Most probably just a few more dates need to be required for the email template
- [x] Template for the email subject + (maybe) associated data in the ProcessSettings since the subject will maybe be different for each bplan (to clarify)
- [x] final version of the email template


pallix:
We can import a bplan with the `import_resource` script but to link to the `processSettings` to a `office_worker` / `user` in the system, the only way now is to give the path of the user.

Is this problematic for the deployment? 

If yes we will need to add one more script...

joka:
> We can import a bplan with the import_resource script but to link to the processSettings to a office_worker / user in the system, the only way now is to give the path of the user.
> 
> Is this problematic for the deployment? 
> If yes we will need to add one more script...
> Probably it's problematic because then we cannot hardcode the name of the user in the json file.

Its not nice but i  guess it will work. 
We have to:

   run the import user script with productive database
   get the wanted user path
   update import resource json file
   run import resource

src/adhocracy_meinberlin/adhocracy_meinberlin/resources/root.py
```diff
@@ -47,10 +50,21 @@ def add_example_process(context: IPool, registry: Registry, options: dict):
                             parent=context['organisation'],
                             appstructs=appstructs)
     # sample bplan
+    office_worker = None
+    users = find_service(context, 'principals', 'users')
+    if users is not None:  # pragma: no cover
+        # ease testing
```
joka: why you set "pragma: no cover"?  Is this check for "None" really needed?

pallix: one functional test was failing

src/adhocracy_meinberlin/adhocracy_meinberlin/resources/subscriber.py
```diff
@@ -1,8 +1,68 @@
...
+from adhocracy_meinberlin import resources
+
+
+def _send_bplan_submission_confirmation_email_subscriber(event):
+    if not hasattr(event.registry, 'messenger'):  # ease testing
+        return
+    messenger = event.registry.messenger
+    proposal_version = event.object
+    proposal_item = find_interface(proposal_version, IProposal)
+    if not _bplan_proposal_has_been_created(proposal_item):
```
joka: I did not understand this at once,  maybe better name "is_proposal_creation_finished(proposal_version)"

src/adhocracy_meinberlin/adhocracy_meinberlin/resources/subscriber.py
```diff
@@ -1,8 +1,68 @@
...
+from adhocracy_meinberlin import sheets
+from adhocracy_meinberlin import resources
+
+
+def _send_bplan_submission_confirmation_email_subscriber(event):
+    if not hasattr(event.registry, 'messenger'):  # ease testing
+        return
+    messenger = event.registry.messenger
+    proposal_version = event.object
+    proposal_item = find_interface(proposal_version, IProposal)
```
joka: proposal_item is only needed for  _bplan_proposal_ fucntion ..., so we can move it there

src/adhocracy_meinberlin/adhocracy_meinberlin/resources/subscriber.py
```diff
@@ -1,8 +1,68 @@
...
+    if not _bplan_proposal_has_been_created(proposal_item):
+        return
+    proposal_values = _get_proposal_values(proposal_version)
+    process_settings = _get_process_settings(proposal_item)
+    if process_settings['plan_number'] == 0 or process_settings['office_worker'] is None:
```
joka: the proposal_ suffix for every  varialbe feels a bit useless. 
Maybe we could make this context information in the function name, like "send_confirmation_mail_after_proposal_creation "

pallix: how would you then name `proposal_version`?

PR #1398, opened 2015-07-06
Document how to set the workflow phase in the installations instructions

pallix:


xi:
I think some more information could be helpful here:

-   Why do I have to set the workflow phase?
-   How do I know which arguments I have to pass to the script?

Additionally, I think the "official" term for this is *workflow state*, not *workflow phase*. Is it?

pallix:
The first argument is the path to the process, so it's dependent of the modelisation.

PR #1401, opened 2015-07-06
rm IName sheet from item, comment, proposals resource types API CHANGE

joka:


src/adhocracy_core/adhocracy_core/resources/item.py
```diff
@@ -33,8 +31,7 @@ def create_initial_content_for_item(context, registry, options):
 
 item_meta = pool_meta._replace(
     iresource=IItem,
-    basic_sheets=[adhocracy_core.sheets.name.IName,
```
pallix: so is a migration required here?

pallix:
6 functionals tests are failing on this branch

src/adhocracy_core/adhocracy_core/resources/item.py
joka: > so is a migration required here?

yup

xi:
Why is the `IName` sheet removed when changing to autonaming? I would have expected that it would simply become read-only.

joka:
> Why is the IName sheet removed when changing to autonaming? 

The frontend team wanted to have it  that way.

>I would have expected that it would simply become read-only.

joka:
Test fixed

pallix:
Migration is written, should be tested on staging first.

nidico:
Looks fine on staging.

PR #1403, opened 2015-07-06
change wording on top of user space

MagdaN:
This fixes an issue in spd and all the other projects. On top of the user page it now says 'Nachricht schreiben' bzw 'Write Message' instead of 'Benachrichtigungen' bzw 'Messages'


nidico:
Right. I also changed it in Transifex.

PR #1404, opened 2015-07-07
Adding pre-wrap styles to user ediable text

local-girl:
Fixes lack of text-wrapping in user entered sections. Tested in mercator, SPD & MeinBerlin


nidico:
This would also apply to the `p` items generated by the `adh-parse-markdown` directive. This would encourage users to write incorrect markdown (e.g. assume that a single line-break indeed breaks the line).

Can you suppress that?

local-girl:
This was deliberate, i tested it with markdown and it seemed to work

local-girl:
OK fixed

local-girl:
continued in #1410

PR #1409, opened 2015-07-08
2015 07 cc spd bugfixes

### LINKED ISSUE #1400: Envelope "send message" button does not work on Safari
The envelope button in the user profile does not trigger the "send message" overlay in the respective column on Safari (e.g. version 7.1.6).

local-girl:
Fixes #1400 and "Selected paragraph in detail view is not highlighted when corresponding comment column is opened" from https://pad.liqd.net/p/bpm-2015-07


MagdaN:
looks good

PR #1378, opened 2015-06-28
[Breaking] Customizable post reset password workflow

nidico:
See commits.

Note: This breaks installations which are currently only embedded, e.g.
meinberlin.

Such installations should add to the frontend config:

    adhocracy.custom.embed_only = true

PR #1412, opened 2015-07-09
Added home link to activation success message

local-girl:
OK so the functionality for this is working, the trouble is the translation key is ```TR__ACTIVATION_CLOSE_THIS_TAB_AND_PROCEED``` and I guess they don't need to close this tab any more? 
download 3 so not only should the key name change but the translation should be altered. who is responsible for this ? @nidico ?

I just used a / link to link to the homepage as ```adhTopLevelState.redirectToSpaceHome("content")``` fails (exceeded stack call) as it's outside of resource area, this is in a FIXME in the function.


local-girl:
In fact I'm not sure the "close this tab" wording ever made sense if you look in the context of SPD

download 4

I am a little unsure of the flow here. In the embedded context like SPD it should read something like "Thanks for your registration, you are now logged in - please follow this link to <u>site name</u> to proceed"

local-girl:
OK I have added this new text (EN only) can someone double check this as the embedding context is a little confusing to me

MagdaN:
It worked fine for me but, please, can somebody else also review this.

nidico:
I would have expected this to be more like #1378, i.e. just change if `adhocracy.custom.embed_only` is false.

xi:
Superseded by #1427.

PR #1415, opened 2015-07-10
Fix: Allow to update documents without image (fixes #1397)

### LINKED ISSUE #1397: cant update document
Whenever I try to update a document resource I get the following error:

This Resource does not provide interface adhocracy_core.sheets.image.IImageMetadata

I am not sure, should that be resolved in backend or in frontend?

nidico:
This works around a backend bug, addendum to
077c4398d0a5fd844668451f57fc9ec0f268245b

If the image reference was null and we *do* set the sheet, the backend
complains:

> This Resource does not provide interface
adhocracy_core.sheets.image.IImageMetadata


MagdaN:
Works, but still needs to be fixed in backend

PR #1419, opened 2015-07-13
2015 07 mn use different lato source

MagdaN:


xi:
I still feel like this is not optimal (especially the kerning). But It fixes the obvious issues.

PR #1421, opened 2015-07-13
Improve installation instruction wrt. to workflow

pallix:


xi:
I guess this is a response to https://github.com/liqd/adhocracy3.mercator/pull/1398#issuecomment-120899797

Looks good to me.

PR #1423, opened 2015-07-13
Fixed white background hover

local-girl:
Fixed white background hover on standard buttons https://tree.taiga.io/project/a3/task/259


xi:
This issue is actually not specific to spd. It exists just the same in e.g. mercator. Therefore I moved the fix to core. Can somebody please review my change?

local-girl:
LGTM

PR #1426, opened 2015-07-13
Fixed white background hover

local-girl:
Continued from #1423 which I had to revert. There is call for this to be a core change but as we use this mixin
```
@mixin button-color($fg, $bg) {
    color: $fg;
    background-color: $bg;

    &.m-call-to-action {
        color: $bg;
        background-color: $fg;
    }
}
```
You see the FG and BG are reversed for m-call-to-action meaning the font-color would become transparent. I think this mixin is pretty redundant but as it's in core I'd like to sit down and look at it properly before making a core change. Keeping this to just an SPD change for now, will revisit soon.

### DIFF
--- a/src/meinberlin/meinberlin/static/stylesheets/scss/components/_button_theme.scss
+++ b/src/meinberlin/meinberlin/static/stylesheets/scss/components/_button_theme.scss
@@ -43,6 +43,7 @@ input[type="button"] {
 
         &.m-call-to-action,
         &.m-call-to-action-secondary {
+            color: $color-text-inverted;
             height: 24px / $font-size-normal * 1em;
             line-height: 24px / $font-size-normal * 1em;
             font-size: $font-size-normal;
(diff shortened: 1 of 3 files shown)


xi:
In that case we should simply copy the fix over to mercator.

local-girl:
OK I adjusted the core again - I got rid of that mixin which doesn't make sense any more. Hopefully the core in line with current designs except SPD which has a different color scheme for buttons

src/meinberlin/meinberlin/static/stylesheets/scss/components/_button_theme.scss
```diff
@@ -97,6 +93,25 @@ input[type="button"] {
...
 }
 
+/*doc
+---
+title: Call to action button index of moving column or embed
+name: button.m-call-to-action.inside
+parent: button
+---
+
+For Mein Berlin CTA buttons inside of the moving columns or embed contexr appear square
```
xi: typo: contexr

PR #1428, opened 2015-07-14
Fix crosswindowmessaging auth exposure

nidico:
Fixup bd8faa8f960ceb77b379903823db5cb13e2e2690


xi:
I have two issues with this:

-   The change needs to be applied to all versions of `Adhocracy.ts`.
-   Why does this check exist in the first place? It only decides on whether `adhUser` and `adhCredentials` are injected. There is a second check before they are used which actually checks if the origin is trusted. I would prefer to remove this check completely.

nidico:
> The change needs to be applied to all versions of Adhocracy.ts.

True.

> Why does this check exist in the first place? It only decides on whether adhUser and adhCredentials are injected. There is a second check before they are used which actually checks if the origin is trusted. I would prefer to remove this check completely.

Because we can avoid unnecessary injections.

PR #1429, opened 2015-07-14
Fix password reset acceptance test

xi:
one acceptance test has been failing for a while now. I hope I got it fixed. Please wait for the travis result before merging this.

Note: The test will still fail in SPD because a different mail template is used there. That should be fixed separately.


nidico:
Nice!

PR #1430, opened 2015-07-14
Fix: modifications with attribute storage sheets are not persistent

joka:
All modifications with the following sheets have not been persistent.
Modifications by import scripts are not related. 

The following sheets are affected: 

- IGroup (used only by import scripts)
- IMetadata (visibility change by evolutions scripts?,  
                    wrong modification date !!
                   )
- IUserBasic (used only by import scripts)
- IUserExtendended ( used only by import scripts)
- IPermissions (not used)


nidico:
If the tests pass in travis, I'll later test manually on the staging server whether the caching issue is fixed with this.

PR #1432, opened 2015-07-14
Fix reset password acceptance test for spd installation

nidico:
As proposed in #1429, this should fix the reset password acceptance test also for the spd installation.

All acceptance tests are supposed to pass now...


xi:
yeay! travis is happy again

PR #1436, opened 2015-07-15
invalidate cache on login

xi:
the frontend cache was completely purged on logout. However, it was not purged on login. This meant that after logging in, cached responses to OPTIONS requests would be reused. We noticed that on the example of blog posts.


nidico:
I also noticed this recently. The fix works fine.

PR #1438, opened 2015-07-15
2015 07 mn add alexanderplatzpolygon

MagdaN:


MagdaN:
@pallix I added the alexanderplatz geojson and also added a function to load this. Is this helpful? If yes, I  can refactor the whole script (also the other functions)

pallix:
Is it not possible to just reuse the `import_resources` script?

MagdaN:
@pallix ah I did not think of this - will try

pallix:
simpler solution in #1447

PR #1440, opened 2015-07-15
2015 07 cc button refactor

local-girl:
OK ! Quite a lot of changes, mainly:

1. although @xi advised against buttons are now style only by class and not selector. Buttons must be given either a ```button``` ```button-cta``` or ```button-cta-secondary``` class to be styled. My main reasoning for this is that ```input[type="select"]``` has a higher specificity than ```.button``` which means ```input[type="select"]``` (or similar) has to be specified every time an overwrite is needed. This is causing very messy CSS.
2. Buttons have been split in to 3 separate(ish) widgets ```button``` ```cta-button``` and ```cta-button-secondary``` this means overwriting the styles of one will not effect others so greatly. 
3. Buttons now have their own color variables. In most cases they don't have to be changed, but where there is a need for this in most cases we don't need a whole new "button_theme.css" just an overwrite of colors in "variables_theme.scss"


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
```diff
@@ -28,97 +37,107 @@ It may have the following states:
...
+<input type="submit" class="button" value="Submit" />
+<input type="submit" class="button" value="Disabled" disabled="disabled" />
 ```
 */
 
 /*doc
 ---
-title: call-to-action modifier
-name: button.m-call-to-action
+title: button-cta
```
xi: the title should be descriptive rather than technical. What about "Call-to-action button"

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
```diff
@@ -28,97 +37,107 @@ It may have the following states:
...
+name: button.button-cta
```
xi: this is no longer a modifier to button, so this should be only `button-cta`

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
```diff
@@ -28,97 +37,107 @@ It may have the following states:
...
 ```
 */
 
 /*doc
 ---
-title: call-to-action-secondary modifier
-name: button.m-call-to-action-secondary
+title: button-cta-secondary
+name: button.button-cta-secondary
 parent: button
```
xi: see comments for `.button-cta`

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_button.scss
```diff
@@ -28,97 +37,107 @@ It may have the following states:
...
 ---
 
-If more than one CTA is required per page, `m-call-to-action-secondary` should
+If more than one CTA is req uired per page, `button-cta-secondary` should
```
xi: typo

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/general/_variables.scss
```diff
@@ -57,6 +57,16 @@ $moving-column-menu-height: 24px / $font-size-normal * 1em;
...
 $annotation-width: 34%;
 
+//Buttons
+$button-height: 24px / $font-size-normal;
+$color-button-text: $color-text-highlight-introvert;
+$color-button-text-hover: $color-text-highlight-normal;
+$color-button-cta-text: $color-text-inverted;
+$color-button-cta-base: $color-text-highlight-normal;
+$color-button-cta-hover-background: $color-brand-two-extrovert;
+$color-button-cta-active-background: $color-brand-two-introvert;
```
xi: These variables are defined before the theme had a chance to overwrite the default variables. So if for example the theme overwrites `$color-test-highlight-normal`, `$color-button-text-hover` will still have the value from core.

To fix this, the variables need to be moved to a file that is loaded between `_variables_theme.scss` and `_button.scss`. Not sure how to do this without making it extremely complex.

xi:
The text is too low sometimes:

clipped_outlines_after1

I guess this is due to the combination of `line-height`, `height`, `border` and `box-sizing: border-box`.

xi:
This all looks really good to me. I found some minor issues that should not be hard to fix (see comments).

Apart from that, I still have to test it with different themes. But that can be done once the other issues are fixed.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/general/_variables.scss
local-girl: Ah, this explains a lot... Would a variables_extra.scss sheet work ?

local-girl: I created a variables_extra.scss but technically we'd need a variables_extra_docs.scss - is this worth it ?

local-girl:
@MagdaN Could you check my fixes today - particularly to the order or stylesheets + browser check. It would be good if we get this merged asap as it's a bit change. Thanks!

MagdaN:
OK, the buttons on meinberlin and mercator look good to me - just a minor issue on firefox: 

screenshot - 16 07 2015 - 14 08 32

xi:
LGTM. @local-girl could you please review my change?

local-girl:
They kind of look too high now! But zoomed it it all looks perfect so it must be a rounding issue. merging

PR #1442, opened 2015-07-15
Add Alexanderplatz workflow

pallix:
Devs must delete their db after this PR is merged. Sample workflow is not use anymore. A new standard workflow is implemented and used for alexanderplatz.


pallix:
alles im grünen Bereich :-)

joka:
lgtm 

I would use a more descriptive naming than "Standard" workflow, but we can do this later when refactoring time.

PR #1444, opened 2015-07-17
2015 07 mn allow owners to delete

### LINKED ISSUE #1439: creators of logbook entry should be able to also delete entry
Only admins can delete logbook entries, but a requirement is, that also the creator of the entry should be able to do that. Is this something that could quickly be implemented?

MagdaN:
This enables creators (and also admins) of a blog entry to delete their blog entry instead of hiding. Fixes #1439 

PROBLEM: the blogentry gets deleted but I need to refresh the page to see the changes. How can that be fixed? Is this a backend or a frontend issue?


xi:
Works fine!

> PROBLEM: the blogentry gets deleted but I need to refresh the page to see the changes. How can that be fixed? Is this a backend or a frontend issue?

Could not reproduce that problem.

Issue #748, opened 2015-02-19
logBackendError broken


nidico:
`Http.Error.logBackendError` regularly breaks itself instead of showing an error. This makes debugging harder instead of improving it.

Example: Open an non-existing page such as http://localhost:6551/r/principals/users/00000i

xi:
> Example: Open an non-existing page such as http://localhost:6551/r/principals/users/00000i

This particular case is probably caused by #256. Still, I would also appreciate a fix in the frontend.

xi:
Fixed by #1445 (68f1332)

Issue #1407, opened 2015-07-08
Passwords can appear in the log files


pallix:
When registering with an already existing users, the REST error is logged and the password is logged in the log file. This sensitive information should not be stored in a clear-text format.

Example of the log:
```
WARNI [adhocracy_core.rest.views][MainThread] Found 1 validation errors in request: <b'{"content_type":"adhocracy_core.resources.principal.IUser","data":{"adhocracy_core.sheets.principal.IUserBasic":{"name":"lol"},"adhocracy_core.sheets.principal.IUserExtended":{"email":"lol@example.COM"},"adhocracy_core.sheets.principal.IPasswordAuthentication":{"password":"passwordlol2"}}}'>
```

nidico:
A similar issue has been fixed in #698.

pallix:
Fix by #1435

Issue #256, opened 2014-11-19
Backend responds with HTML (not json).


mf59816:
The following:

    curl -H "Content-Type: application/json" http://localhost:6541/no-such-path

should produce json data on stdout, but it produces HTML.  Please instruct how to use this correctly or fix.

This is related to #246.

xi:
Fixed in #1445.

PR #1445, opened 2015-07-18
refactor error handling rest api

### LINKED ISSUE #1019: No errors in batch response
Some batch responses do not contain errors. For example:

    $ curl 'http://localhost:6541/batch' -H 'Content-Type: application/json;charset=UTF-8' --data-binary '[{"method" : "POST","path" : "http://localhost:6541"}]'
    {"responses": [{"body": {}, "code": 403}], "updated_resources": {"created": [], "modified": [], "removed": [], "changed_descendants": []}}

I would have expected that there were one or more errors in `responses.body.errors`.

This might be connected to #256.

### LINKED ISSUE #1277: too much noise in console output when running bin/polytester
The main reason for this is that we always include pyramid_exclog.
to make sure  every backend exception is logged.

We should only include this in productive installations (just move the include to the ini file)

joka:
API Changens:

Batch request errors:
   - same error descriptions like without batch (before only 400 BadRequest)

Request errors:
   - all HTTPClientErrors have json body with error description (before only 400, 403, 410 and 500)

Internal Changes:

   - mv request error handling / logging to rest.exceptions (SRP)
   - add json body with description to all HTTPClientErrors Fix #1019
   - make sure all request headers, body, errors entries are logged nicely on error 
   - rm dependency to cornice (reduce complexity)
   - reduce noise in bin/polytester output Fix #1277
   - Fixup: passwords in batch requests may appear in log file  #1407 
   - Fix: arbitrary long request body may appear in log file, max length is now  5000 chars
   - add global test fixture request_ 
   - add global test fixture log to capture log messages


joka:
Still to do:  Test coverage is broken

joka:
juhu my first green mark

xi:
I think the frontend needs adjustments. I will talk to @joka about this today.

@joka: Is this somehow related to #256?

src/adhocracy_core/adhocracy_core/rest/test_exceptions.py
```diff
@@ -3,9 +3,140 @@
...
+        request_.body = '[]'
+        self.make_one([], request_)
+        assert '[]' in str(log)
+
+    def test_log_ignore_if_request_body_json_other(self, request_, log):
+        request_.body = b'None'
+        self.make_one([], request_)
+        assert 'None' not in str(log)
+
+    def test_log_abbrivated_request_body_if_gt_5000(self, request_, log):
```
pallix: abbrivated => abbreviated

joka:
> @joka can you maybe comment here what you explained to me?

ok

xi:
I adapted the frontend and potentailly fixed #748 along the way. However, I still found some issues in the backend:

-   The status text for batch responses is always "Bad Request" even if the status code is 404 or 410.
-   The `description` in 404 responses is the requested path. Something like "The requested resource {path} could not be found." would be much nicer IMHO.
-   The response body to an OPTIONS request to a deleted resource is `{"errors":[{"location":"body","name":null,"description":"Invalid JSON request body"}],"status":"error"}`
-   The response body to an OPTIONS request to a non existing resource is empty

I did not do any structured testing. So this is only what I found be accident.

joka:
> The status text for batch responses is always "Bad Request" even if the status code is 404 or 410.
fixed
>     The description in 404 responses is the requested path. Something like "The requested resource {path} could not be found." would be much nicer IMHO.
Now its the status code / title, more description is no available.
>     The response body to an OPTIONS request to a deleted resource is {"errors":[{"location":"body","name":null,"description":"Invalid JSON request body"}],"status":"error"}
cannot reproduce, can you show me?
>     The response body to an OPTIONS request to a non existing resource is empty
cannot reproduce, can you show me?

xi:
>> The response body to an OPTIONS request to a deleted resource is {"errors":[{"location":"body","name":null,"description":"Invalid JSON request body"}],"status":"error"}
>
> cannot reproduce, can you show me?

I checked again and noticed that I only get this error when I add this header to the request: `Content-Type: application/json`.

>> The response body to an OPTIONS request to a non existing resource is empty
>
> cannot reproduce, can you show me?

Cannot reproduce anymore, too.

joka:
> The response body to an OPTIONS request to a deleted resource is {"errors":[{"location":"body","name":null,"description":"Invalid JSON request body"}],"status":"error"}
> cannot reproduce, can you show me?
> I checked again and noticed that I only get this error when I add this header to the request: Content-Type: application/json.

This is the intended behaviour.  If you set the content type to json, the backend expects a valid json body

xi:
>>> The response body to an OPTIONS request to a non existing resource is empty
>>
>> cannot reproduce, can you show me?
>
> Cannot reproduce anymore, too.

I had this issue again. However, it only showed up in chromium. Not in firefox or curl. So I guess it is a chromium bug.

PR #1447, opened 2015-07-20
Sample alexanderplatz data and process as JSON

pallix:
The  `sample_alexanderplatz.json` file can be imported with `import_resource`

This avoid creating new code like in #1438 and also avoid having too much hardcoded/sample data in the `root.py` data.

Only problem: If an alexanderplatz process is needed for the tests (it's not the case now), what would be the best way to create it?


pallix:
Devs can execute the following command to work on alexanderplatz 
```
$ import_resources   etc/development.ini ~/projects/adhocracy3.mercator/src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/sample_alexanderplatz.json 

$ set_workflow_state etc/development.ini /organisation/alexanderplatz participate
```

joka:
For me it makes no different if we have test data in a python module or a json file. 
What is the benifit?

pallix:
Benefits:

- we can reuse the json sample to create a json for production
- avoid cluttering the source code with coordinates for the polygon
- easier to change by non-backend developpers

Drawbacks:

- maybe we still need some data in the code for the tests (but it could be smaller data, like small polygons)
- ??

joka:
oki

Issue #1449, opened 2015-07-20
Upgrade to Typescript 1.5


xi:
Typescript 1.5 has been released [today](http://blogs.msdn.com/b/typescript/archive/2015/07/20/announcing-typescript-1-5.aspx). tslint has already released a beta version with support for most new features ([changelog](https://github.com/palantir/tslint/blob/master/CHANGELOG.md)).

This new version adds support for many exciting ES6 features. We should make a plan on which of these features we want to use. The decision needs to be documented and implemented.

I am in favor of making the following changes:

-  use [for-of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) instead of `_.forOwn` where it is easy to do so (e.g. iterating over a list)
-  use [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) where it makes sense
-  always use [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let)/[const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const) instead of `var`
-  switch to the new module syntax
-  build UMD instead of AMD modules by default.

In general I think we should not get too fancy. We should not expect new team members to know all of this stuff, even if it is part of ES6. Still I think that these can actually help us write understandable and robust code, e.g. the `let` keyword may help to avoid common errors.

xi:
Since #1450 has been merged, we are actually using TypeScript 1.5 (and tslint 2.4.0). We still need to decide on which new features we want to use.

xi:
#1636 makes the switch to UMD module system.

xi:
We have already moved on to TypeScript 1.8 and the general opinion is that we might be better of with vanilla JavaScript. So using more of the features is not desirable at the moment. closing.

PR #1450, opened 2015-07-20
Update javascript dependencies

xi:
It seems like we did not do this for a long time so there are many relevant changes.

The most important updates are

-   [TypeScript](blogs.msdn.com/b/typescript/archive/2015/07/20/announcing-typescript-1-5.aspx) (1.4.1 -> 1.5.3)
-   [tslint](https://github.com/palantir/tslint/blob/master/CHANGELOG.md) (2.1.1 -> 2.4.0)
-   [angular](https://github.com/angular/angular.js/blob/master/CHANGELOG.md) (1.3.15 -> 1.4.3)
-   [angular-cache](https://github.com/jmdobry/angular-cache/blob/master/CHANGELOG.md) (3.2.5 -> 4.3.2)
-   [angular-translate](https://github.com/angular-translate/angular-translate/blob/master/CHANGELOG.md) (2.6.1 -> 2.7.2)
-   [lodash](https://github.com/lodash/lodash/wiki/Changelog) (3.6.0 -> 3.10.0)

In spite of the big version jumps the updated was relatively painless. However, we should check for any issues. Additionally, we should check the list of new features and see which of them we want to use. For TypeScript, see #1449.

Open issues:

-   [ ] I feel like the cache is not working anymore
-   [x] angular-translate logs a security warning. This is probably relevant and should be fixed as soon as possible.
-   [ ] also update protractor
-   [ ] also update DefinitelyTyped


nidico:
After the `moment` update, the dates now include chinese characters.

nidico:
> I feel like the cache is not working anymore

I just checked - the cache is working. However due to the batch request in the routing code which fetches all ancestor resources, we invalidate too much. This is unrelated to that PR though.

nidico:
Looks all good.

PR #1451, opened 2015-07-20
Run linters on travis

xi:
this extends `bin/check_code` to be able to check all files, not only the staged ones. This is then used to run the linters on travis.

Unfortunately, flake8 reports many issues. I guess this is due to wrong config. Can someone tell what needs to be changed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ install:
   - "python bootstrap.py --setuptools-version=12.0.4"
   - "PATH=${TRAVIS_BUILD_DIR}/bin:${PATH} ./bin/buildout -c buildout-${BUILDOUT_TARGET}.cfg 2>/dev/null"
 script:
+  - "./bin/check_code -a"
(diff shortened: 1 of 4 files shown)


xi:
There also seems to be an issue with tslint which should be fixed by merging #1450

xi:
I saw that most issues are for lines with a `# noqa` comment. According to the [pep8 docs](https://pep8.readthedocs.org/en/latest/intro.html#error-codes), `# noqa` only disables few rules. We can either disable the rules completely or disable linting for the whole file (using `# flake8: noqa`).

Site note: `# flake8: noqa` is used in some places at the end of the line. In this case, it also disables flake8 for the complete file. So I would suggest to move these comments to their own lines on the top of the file.

xi:
I removed the linters from travis again. So now we can merge this PR, fix all open issues and then add it again.

pallix:
So is the plan to check the whole code on each PR with travis? I guess it will make checks slower but relative to the time taken by the acceptance tests it's probably acceptable.

pallix:
LGTM

PR #1452, opened 2015-07-21
Make flake8 happy

pallix:


xi:
I think the occurrences of `# flake8: noqa` at line endings should be removed/changed, too.

PR #1453, opened 2015-07-21
avoid making a batch request for GET only

xi:
batch requests invalidate many cache items by default. Optimising that
is hard. So we should avoid using them for GET requests only.


nidico:
LGTM and helps!

PR #1455, opened 2015-07-21
Fix dependance on process

xi:
If you open a user profile with a non-empty list of mercator proposals you will get a error like this:

    TypeError: Cannot read property 'workflow_state' of undefined
    at MercatorProposal.ts:976

The reason is that there is no process in the URL and therefore `processUrl` is not set in topLevelState.

I worked around this by making the use of `processUrl` optional. An actual fix would probably be to find a process for each proposal instead of taking one from the current route. But that would also be relatively slow. Not sure what to do about this.


nidico:
The fix works. I would have expected to fix this by setting `processUrl` to `undefined` instead of `/` if there is no process in the ancestors hierarchy. Is there a reason you did not?

xi:
> The fix works. I would have expected to fix this by setting processUrl to undefined instead of / if there is no process in the ancestors hierarchy. Is there a reason you did not?

I think there is a link that points to `processUrl` somewhere. Maybe I should have changed that link instead...

nidico:
Understand. Feel free to changethe link or merge as-is.

PR #1457, opened 2015-07-21
set login form to pristine on backend error (fixes #1454)

### LINKED ISSUE #1454: Invalid error message when logging with invalid password
When entering an invalid password on the login page, the error is, in the german translation:

"Passwort erforderlich".

It should be something like "Invalid password."

xi:
we clear the password field which resultes in a "password required" error on it. Setting the whole form to pristine fixes that.


pallix:
thanks, much better.

PR #1458, opened 2015-07-21
simple check for existance of a sheet on resources

xi:
This adds some information to generated TypeScript classes that will hopefully help us to write more generic widgets. The idea is that we can add some functionality to a widget depending on the existance of a specific sheet.


nidico:
What's the advantage over extracting the information from the meta_api? I agree it's coherent with having the generated resources in general.

xi:
> I agree it's coherent with having the generated resources in general.

Mostly that. But it also means you do not have to inject an extra service. The information comes directly with the resource on `adhHttp.get()`.

nidico:
LGTM.

PR #1461, opened 2015-07-23
map-aware document

xi:
This changes the document to optionally contain coordinates. In that case, a different resource type is used.

I am not at all happy with this solution. But I feel like a better solution is only possible if we do a lot of refactoring.

Note that this breaks spd because it requires meinberlin-specific resource types in `Document.ts`. My proposal is to move those resource types to core.

Also note that the only user facing change I did was adding a `adhDocumentMapListItem` directive. AFAIK maps should also be in the create and edit form, but not on the detail directive.


xi:
I updated the document types so now this should work fine in once #1467 is merged.

xi:
I now also added `adh-document-map-list-item` to the map listing in the alexanderplatz workbench. It does not seem to work though (probably #1028).

MagdaN:
@xi can you merge master

MagdaN:
looks good

PR #1462, opened 2015-07-23
Move mein berlin specific code from core to meinberlin (fixes #1028)

### LINKED ISSUE #1028: meinberlin specific code in adhocracy_core

xi:
The reason this was here in the first place is that we had trouble with nested injection (`adhMapListing` and `adhMapListingInternal`).  But when I tried this now it just worked.

I also tried to reproduce the original issue by creating a minimal application: http://plnkr.co/edit/YhwT9X1ra4R6E7PeDEnF . I was not able to reproduce the issue, even with older angular versions.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Listing.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Mapping/Listing.html
@@ -4,6 +4,5 @@
     data-height="400"
     data-empty-text="{{emptyText}}"
     data-polygon="polygon">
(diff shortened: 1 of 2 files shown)


MagdaN:
cnat remember what the problem was either, but tested it and it works fine

PR #1466, opened 2015-07-27
Migrate all scripts to argparse

### LINKED ISSUE #1271: migrate from optparse to argparse
[optparse](https://docs.python.org/2/library/optparse.html) has been deprecated since python 2.7. It is recommended to use [argparse](https://docs.python.org/2/library/argparse.html) instead. We still use optparse in the following scripts:

-  src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/create_process_for_region.py
-  src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/import_geodata.py
-  src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/translations.py
-  src/adhocracy_mercator/adhocracy_mercator/scripts/export_proposals.py

They should be migrated.

See src/adhocracy_mercator/adhocracy_mercator/scripts/export_users.py for an example of how to use argparse.

xi:
fixes #1271

### DIFF
--- a/src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/create_process_for_region.py
+++ b/src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/create_process_for_region.py
@@ -1,7 +1,7 @@
 """Create Process for region."""
 
 import sys
-import optparse
+import argparse
 import textwrap
 import inspect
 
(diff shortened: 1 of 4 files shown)


pallix:
thanks!

PR #1469, opened 2015-07-27
reduce code duplication in comment column

xi:
This specifically targets the comment column in meinberlin-alexanderplatz. Before this, the comment column as defined in meinberlin-kiezkassen was used there. The only thing that did not work was the close button.

I saw two options to fix this:

1.  Define a separate `commentCloseUrl` in the routing
2.  Use the `commentableUrl` and create a route for it if required

I went with the first option because it seemed to be easier to implement.

I also compared the differen comment columns and found that they are very similar. I smoothed out some differences in the implementation and redueced some code duplication. There are still template overrides in spd and mercator. I hope that we can remove them some time in the future, too.


MagdaN:
looks good

PR #1472, opened 2015-07-27
Changed button styline meinberlin

local-girl:
The pseudo class :first-letter doesn't really work for us (a) it doesn't work on input type="button" and (b) it doesn't work on a button that has a :before pseudo class (as that will act as the first-letter)

I have now made a general style of text-transform: capitalize; on all MB buttons, with the disadvantage that every letter is capitalized. I will ask the designers tomorrow.

### DIFF
--- a/src/meinberlin/meinberlin/static/stylesheets/scss/components/_button_theme.scss
+++ b/src/meinberlin/meinberlin/static/stylesheets/scss/components/_button_theme.scss
@@ -15,11 +15,6 @@
     font-size: $font-size-small;
     height: 24px / $font-size-small * 1em;
     line-height: 24px / $font-size-small * 1em;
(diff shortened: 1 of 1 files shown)


xi:
The `:first-letter` pseudo-class is also used in `src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_comment.scss`. Could you check whether it works there?

Apart from that: Feel free to merge as soon as you have the ok from designers.

xi:
I talked to the designers and got their OK.

> The `:first-letter` pseudo-class is also used in `src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_comment.scss`.

It does not have a hover, is not a link and has no `:before`. So everything is fine there.

Issue #1418, opened 2015-07-13
Documentation missing for item_creation_date filter


pallix:
item_creation_date filter catalog is implemented but not documented in the REST API.

pallix:
Fix by #1473

PR #1473, opened 2015-07-28
Improve rest api doc

pallix:


pallix:
Fixes #1418

xi:
Looks good to me, except for one detail: In rest_api.rst you reference `IMetadata`. But the actual definition is at `MetadataSchema`. Is that intended?

pallix:
The interface is used as a marker and the `MetadataSchema` for validation. We can say the concrete type is opaque and of type `IMetadata` so it's correct.

xi:
It is consitent with the existing documentation. That's fair enough for me.

PR #1475, opened 2015-07-28
Add map input to document form

xi:
This continues on #1461 by adding an optional map input to the document create/edit form.


MagdaN:
looks good

PR #1476, opened 2015-07-28
Add more options to set workflow script

pallix:
- Add an `--absolute` option to `set_workflow_script`

Allow to use the absolute path to reach the wanted workflow state, whatever the current state is. There will be no transition to the states expressed in the path that are before the current state.

Example

     /bin/set_workflow_state --absolute etc/development.ini /mercator draft announce

- Add an `--info` option to `set_workflow_script`

Since everybody get confused with workflows, this option display the current workflow name, state, next reachable states and defined states

    ./bin/set_workflow_state --info etc/development.ini /mercator 

    name: mercator
    current state: announce
    next states: ['participate']
    all states (unordered): {'closed', 'evaluate', 'announce', 'participate', 'draft', 'result'}

@slomo should update production scripts

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -84,12 +84,17 @@ Check that everything is running smoothly::
 
     ./bin/supervisorctl status
 
+Get information about the current workflow::
+
+  ./bin/set_workflow_state --info etc/development.ini <path-to-process>
+  # Example
+  ./bin/set_workflow_state --info etc/development.ini /mercator
+
 Change the workflow state (most actions are not allowed for a normal user in the initial 'draft' state)::
 
-  ./bin/set_workflow_state etc/development.ini <path-to-process> <workflow-state>
-  # examples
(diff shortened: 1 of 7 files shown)


xi:
Looks good to me. There are some interesting edge cases, e.g. having one state multiple times in a single workflow (e.g. `prepare review participate review result`) or using `--absolute` and `--reset` the same time. Maybe you can clarify the behavior in these cases. But that's nice-to-have from my point of view.

I also noticed that `bin/set_workflow_state etc/development.ini /organisation/alexanderplatz draft announce participate` works without errors even if the worklfow is already in *participate*. This is still a bug.

joka:
nice, 
Not part of this pull request but:

   if we change the workflow state like with this script, the client/server varnish cache is not notified about resource changes.  
   Instead of using the workflow object directly we should use the WorkflowAssignment sheet. to fix this.

PR #1487, opened 2015-07-29
2015 06 cc alex map switcher

local-girl:
Added a map switcher to alex. Manu says its OK to have the switcher in the moving column bar


local-girl:
Added a fixme for now, will review later

xi:
I merged a part of this manually (e96a596) and rebased the other part to #1491.

PR #1492, opened 2015-07-29
modify map listing for meinberlin

xi:
*split from #1487*

This moves `map-list-zoom` to the top of the map list map. To avoid overcrowding the top part, the zoom controls are then moved to the bottom right.

I personally feel like this should either be implemented for all maps or for none. Additionally, the zoom controls are moved for all packages while the `map-list-zoom` is only moved in meinberlin.

@local-girl will talk to the designers about this tomorrow.


xi:
Closing for now. We can reopen of this comes up again.

PR #1493, opened 2015-07-29
Refine standard workflow permissions

pallix:


pallix:
Two functionals tests need to be fixed

pallix:
@xi functional tests are fixed

PR #1496, opened 2015-07-30
Fixed up list item style

local-girl:
Fixed up list item style for alex. Baddge is dummy data currently, also "location" seems to be missing, this is not in the document form itself - who knows about this? 
download 6


MagdaN:
In Chrome and Firefox the writing inside the badge seems to be a little bit stuck at the top:

badges

local-girl:
I say merge I'll look at it another time

MagdaN:
ok

Issue #1498, opened 2015-07-30
Initiator/Moderator can not see the proposals on 'draft' phase


pallix:
The 'participant' role is given to everybody who is 'authenticated'. The person having the initiator role so gets the 'participant' role in addition. In draft mode view permission is 'Deny' for participant, and 'Allow' for initiator. 

This cause the GET (view) request on the proposal to be denied.

pallix:
Maybe initiators should not get the participant roles. Participant role should probably not be given to all authenticated users but  maybe only assigned to user which are created without a role (in the colander struct)?

pallix:
@joka If we want to keep having multiple roles used in an ACM for principals, we need to define rules for the ordering of Allow/Deny action or precedence rules for principals (moderator takes precedence on participant)

joka:
This should be closed no, @slomo ?

slomo:
Yes ... resolved as far as I know

PR #1500, opened 2015-07-30
Allow initiator to read / edit proposals in draft

pallix:
Temporary fix for #1498


pallix:
A regression test is missing

pallix:
@xi can you review/merge?

pallix:
I'm closing this PR since the way initiator are defined, is not coherent with our definitions of the concepts or roles where we have:

Initiator:

    Can create processes
    Can configure workflow (for example set dates for time based transitions)
    Can do general workflow transition

PR #1501, opened 2015-07-30
Fixed info badge alignment

### LINKED ISSUE #1497: Alex: In Chrome and Firefox the writing inside the badge seems to be a little bit stuck at the top
see #1496

local-girl:
fixes #1497


xi:
There actually wasn't supposed to be a red border on top of `.badge.m-info`. That was part of the bug. What happened was that the `inline-block` mixin automatically set `vertical-align: middle`. So just resetting that to `baseline` should have been enough. I adjusted the fix in 7476bc1 accordingly. If the bug shows up again, we can just revert my commit.

PR #1502, opened 2015-07-31
add simple form to assign badges

MagdaN:
This implements a simple way to assign badges. Got to url like that: 

http://localhost:6551/embed/assign-badge?badges-path=http:%2F%2Flocalhost:6541%2Forganisation%2Falexanderplatz%2Fbadges%2F&poolpath=http:%2F%2Flocalhost:6541%2Forganisation%2Falexanderplatz%2Fdocument_0000003%2Fbadge_assignments%2F&badgeable-path=http:%2F%2Flocalhost:6541%2Forganisation%2Falexanderplatz%2Fdocument_0000003%2FVERSION_0000002%2F

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Badge/Assignment.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Badge/Assignment.html
@@ -1,27 +1,26 @@
 <h2>Badge Assignment</h2>
 <form
-	name="form"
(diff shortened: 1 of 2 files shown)


xi:
I did some gardening. Among other things, I changed `poolpath` to `poolPath` so the example url needs to be adapted accordingly.

Apart from that this works fine. I would like to have some success/error feedback, buts that can be added later.

PR #1511, opened 2015-08-04
Add option to turn user token validation off

ChristianSi:
This allows running A3 behind a Thentos proxy that handles user authentication and only grants properly authenticated users access to the A3 backend.

### DIFF
--- a/etc/development.ini.in
+++ b/etc/development.ini.in
@@ -37,6 +37,8 @@ adhocracy.abuse_handler_mail = abuse_handler@unconfigured.domain
 adhocracy.message_user_subject = [{site_name}] Message from {sender_name}: {title}
 # If true, new user accounts are activated immediately without email verification
 adhocracy.skip_registration_mail = false
+# Allows turning off user token validation if behind a validating proxy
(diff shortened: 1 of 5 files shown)


pallix:
Can you add a unit test for the `authenticated_userid` and fix the code coverage?

@slomo will need to add this parameter to `backend.ini` when we deploy a3 with Thentos.

ChristianSi:
Tests added.

pallix:
Thanks. Merging.

joka:
The proper pyramid way to handle different authentication policies is:

- use [RemoteUserAuthenticationPolicy](http://docs.pylonsproject.org/projects/pyramid//en/latest/api/authentication.html) instead of in adhocracy_core.authentication.TokenHeaderAuthenticationPolicy

- for convenience add config options for existing authentication policies: 
    "adhocracy.use_remote_user_authentication"
    "adhocracy.use_token_header_authentication"

I would add a nice to have refactoring task for this. ok?

ChristianSi:
@joka:

> I would add a nice to have refactoring task for this. ok?

Do that, sure.

PR #1515, opened 2015-08-04
Different icons for proposal and document

xi:
This allows to add map icons via a provider and then select which to use per directive.

Note that the document icon is selected in the document code itself so it applies to all instances of documents. The used icon is currently only defined in the alexanderplatz workbench.


MagdaN:
looks good

PR #1516, opened 2015-08-04
Refine standard permissions

pallix:


pallix:
automerging because @xi is busy :-)

PR #1521, opened 2015-08-05
Add permissions for documents

pallix:
Create 'create_document' and 'edit_document' permissions for document and paragraph.

This allow Participants to create proposal but not Documents in the participate state (for example).


pallix:
@xi  please have a look when you have time

src/adhocracy_core/adhocracy_core/workflows/standard.py
```diff
@@ -10,9 +10,11 @@
     'states': {
         'draft': {'title': 'Draft',
                   'description': 'This phase is for internal review.',
-                  'acm': {'principals':         ['anonymous',   'participant', 'moderator', 'creator', 'initiator'],
+                  'acm': {'principals':         ['anonymous',   'moderator',   'participant', 'creator', 'initiator'],
                           'permissions':
-                          [['view',              'Deny',        'Deny',        'Allow',     'Allow',   'Allow'],
+                          [['view',              'Deny',        'Allow',       'Deny',        'Allow',   'Allow'],
+                           ['create_document',    None,         'Allow',        None,          None,      None],
+                           ['edit_document',      None,         'Allow',        None,         'Allow',   'Allow'],
```
xi: IMHO moderators should also be allowed to create/edit documents in later phases. Can you please clarify this?

pallix:
@slomo the state of the workflow should be reset with set_workflow_state script so that new permissions get set

xi:
This seems to work fine. However, the frontend needs adoption:

We currently only check whether we can POST to the process. But now we can post proposals, but not documents. So we need a better check (concerns the create proposal link and the respective route)

pallix:
I'll add permissions to add/edit documents in announce/participate/result phase for mercator in this branch

src/adhocracy_core/adhocracy_core/workflows/standard.py
xi: Talked to @rukala: The moderators need create_document and edit_document in all phases except for freeze (or however that is called now)

pallix: It's called 'evaluate'. They can create/edit in evaluate. But not in 'result'. Should we allow them to do in result?

pallix: i guess we only want them to create/edit in draft, announce, participate then.

xi:
I am still not sure about the exact permissions. But I guess they can still be changed in another pull request. @pallix can you please check again with @rukala?

pallix:
I get confused myself, my comment does not reflect the code. 

I think the best is to have a look at:

https://github.com/liqd/adhocracy3.mercator/blob/master/src/adhocracy_meinberlin/adhocracy_meinberlin/workflows/test_alexanderplatz.py

Permissions correction in another PR is fine.

PR #1522, opened 2015-08-05
Fix cameFrom in alexanderplatz workbench

xi:
I fixed all action/cancel links in the workbench to correctly set and get `cameFrom`. While working on that I also refactored the cameFrom related functions. Note however that #1061 is still in effect.

An interesting observation: While working on this I also tried to keep the cancel URLs and use them as default url for `goToCameFrom()`. However, canceling all other event handlers with `event.preventDefault()` or even `event.stopImmediatePropagation()` did not work because of our href hack introduced in #406. We should definitely look into it again soon.


MagdaN:
looks good

PR #1527, opened 2015-08-06
Add 'authenticated' group to all imported users

pallix:
When a group was defined in the JSON file for the import_users script,
the default group, 'authenticated', were not associated to the created
user if groups were already defined. This caused permissions problems
for logged in users.


joka:
This was the intendend behavior.

The default group is for users that loggin and want to participate, therefor they get the role "participator". Its not intended for special users like initiators, moderators. We don't  them to
have the "participator" role. For example to prevent permission conflicts or to disallow voting.

Note: "authenticated" is a bad name for the default group, because its not meant for all "authenticated" users.

pallix:
Ok but then 1) the script is now wrong 2) there is a bug because upon loggin moderators get assigned to the authenticated group

joka:
We can fix this together with  #1573

PR #1528, opened 2015-08-07
2016 mn alex add context

MagdaN:
this adds the context for embedding to alexanderplatz and also adds the process widget.


MagdaN:
There is a problem with the phase widgets tabs - I need to click the several times until it opens. @xi any idea, why?

xi:
Continued in #1531. I will create another pull request for the phase information widget soon.

PR #1530, opened 2015-08-10
Fixed up header styles for mobile

### LINKED ISSUE #1529: Register button cut off on mobile

local-girl:
Fixes #1529

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
@@ -173,7 +173,7 @@ depends on a lot of factors:
 
 .l-header,
 .l-header-center {
-    @extend %moving-column-single-width;
+    max-width: $moving-column-single-width-max;
 }
 
 .l-header-left {
@@ -201,6 +201,12 @@ depends on a lot of factors:
 
 .l-header-right {
     float: right;
+
+    @media (max-width: $moving-column-single-width-min) {
+        i[class*=icon-] {
+            display: none;
+        }
+    }
 }
 
 .l-center {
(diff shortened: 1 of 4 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_moving_columns.scss
```diff
@@ -155,8 +155,8 @@ The menu should always have the same state as its columns.
 }
 
 %moving-column-single-width {
-    max-width: $font-size-normal * 55;
-    min-width: $font-size-normal * 35;
+    max-width: $moving-column-single-width-max;
+    min-width: $moving-column-single-width-min;
 }
```
xi: This is not used anymore and should be removed

xi:
I checked again and saw that `minShowWidth` in the JavaScript code is only used to define the breakpoint on which we switch from two-column to one-column mode. It is not enforced as a minimal column width. So that's why it is perfectly safe to remove that rule from CSS. Great!

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/_layout.scss
```diff
@@ -201,6 +201,12 @@ depends on a lot of factors:
 
 .l-header-right {
     float: right;
+
+    @media (max-width: $moving-column-single-width-min) {
+        i[class*=icon-] {
+            display: none;
+        }
+    }
```
xi: Why this change? I see no real benefit in hiding the icon.

local-girl: Just thought it gives a bit more breathing room

local-girl: minShowWidth is used in css kind of. When A3 was designed - when you have 2 columns the header width remains a 1 column width as opposed to stretching out, this would look weird on a wide monitor

xi:
LGTM.

PR #1532, opened 2015-08-10
Document how to set the initial workflow state

ChristianSi:


xi:
Thanks for the contribution! I am not totally sure though if this is 100% correct. I think the example only works if the workflow is already in a state that can transition `participate`. This might not necessarily be the case. @pallix can you please check?

pallix:
@ChristianSi it's already documented in the "Run the application" section. set_workflow_state takes the path where to go (since workflows are a graph). If this is not clear enough modify the doc in the "run the application" section, otherwise close the PR.

ChristianSi:
Ah, hadn''t know that, I think it wasn' there when I inquired about this.  But isn't a WORKFLOW_NAME argument needed? I'm pretty sure that I called

    ./bin/set_workflow_state etc/development.ini /mercator mercator participate

rather than

    ./bin/set_workflow_state etc/development.ini /mercator announce participate

pallix:
The workflow name is not necessary anymore. I tottally understand your confusion since we didn't many changes to this important script in a short period.

Current options on master:

```
usage: set_workflow_state [-h] [-i] [-a] [-r]
                          ini_file resource_path [states [states ...]]

Set a workflow state for a given resource. usage:: bin/set_workflow_state
etc/development.ini <resource-path> <state>

positional arguments:
  ini_file        path to the adhocracy backend ini file
  resource_path   path of the resource
  states          list of state name to do transition to

optional arguments:
  -h, --help      show this help message and exit
  -i, --info      display information about the workflow
  -a, --absolute  use an absolute path for the list of states
  -r, --reset     reset workflow to initial state
```

Now I see that the docstring is not adequate it should not mention `<state>`.

pallix:
It's also possible that you call `./bin/set_workflow_state etc/development.ini /mercator mercator participate` and that it worked before we merged #1476. If you were already in the participate state the script would not do anything, even not checking the validity of the state names

ChristianSi:
OK. It was quite same time ago when I actually called it.

PR #1534, opened 2015-08-10
2015 08 mn alex check post permissions

MagdaN:


MagdaN:
Code is tested with db from that branch #1532

xi:
I made it more generic by moving the check to the `options` object itself. @MagdaN can you please review?

PR #1535, opened 2015-08-10
alexanderplatz phase information

xi:
*continued from #1528*

This adds a phase information widget to the alexanderplatz context. There are still two issues:

-   [ ] Clicking on a phase tab will only open the drop down on second try. This also affects kiezkassen.
-   [ ] `currentPhase` is currently bound to the workflow state. It should not be in this case.

These issues can be fix in this branch or in separate ones. I don't care.


MagdaN:
looks good

PR #1538, opened 2015-08-10
add hard link to config instead of getRequest to backend

MagdaN:


xi:
@slomo, @nidico Do we need to add this to some config file?

xi:
Review no longer relevant

PR #1539, opened 2015-08-10
Reduce bugginess of tabs

local-girl:
This works better as height cannot be calculated on a div with display: none


xi:
This does not work because the panes are displayed all the time. They may not be visible, but not all users interact with adhocracy by visual means. So `display: none` is actually completely right.

local-girl:
what about aria-hidden ?

xi:
We dicussed this offline. We will try again with `display: none`. If that absolutely does not work we will use `visibility: hidden` or something similar.

PR #1548, opened 2015-08-11
Add resources for the burgerhaushalt process

pallix:
Add backend and resources sheets. Use standard workflow.


pallix:
I didn't use any inheritance or reuse the resources from kiezkassen as what is in common between both projects is small and the modelling is rather simple. I'm not convinced it's a good idea to link both resources ; if a requirement changes in kiezkassen it would probably not imply a changes in burgerhaushalt.

pallix:
Developers should execute the following command to work on this process:

`./bin/import_resources etc/development.ini src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/sample_burgerhaushalt.json`

`./bin/set_workflow_state   etc/development.ini /organisation/burgerhaushalt announce participate`

joka:
lgtm

PR #1559, opened 2015-08-12
Initial stylesheet variables

local-girl:
I'm sure the vars themselves will change but it's something for now


xi:
The colors of `.like-button` on hover are white-on-white. I guess it is not a good idea to set `$color-brand-four-normal` and `$color-brand-one-extrovert` to white. I will talk to the designers about this.

PR #1569, opened 2015-08-17
fix kiezkassen not badgeable

joka:
API Change:

all meinberlin proposals item/versions l inherit from resource interface  adhocracy_core.proposal.IProposal/IProposalVersion now

all meinberlin proposal items have a badge_assignment service pool now

MANUAL UPGRADE:

Please run first:
     bin/sd_reindex etc/development.ini
then the upgrade step:
     bin/sd_evolve --latest etc/developmnet.init


pallix:
Ok this potential case cannot happen as the pool for badge assignment is added under the item, and the iteration occurs on all items ( = proposals).

PR #1575, opened 2015-08-18
Add documentation for administration scripts

pallix:


docs/administration.rst
```diff
@@ -0,0 +1,82 @@
+Administration
+==============
+
+Several console scripts are provided to facilitate the administration of
+Adhocracy 3.
+
+Workflow state
+--------------
```
xi: I would prefer to extend the help that is given on `./bin/set_workflow_state -h` and keeping this section short.

pallix:
In my opinion it would be too much documentation for the `--help` option. I rewrote the documentation in the script to make it clearer.

joka:
nice, but we have now documentation for these scripts at two different places. This is more work and 
its quite risky that one place will have outdated information in the future.

I also suggest to move this to the docstring / help options.
In  administration.rst we should just  just references the script functions docstrings and or the script module docstring.

pallix:
@joka can you merge?

src/adhocracy_core/adhocracy_core/scripts/set_workflow_state.py
```diff
@@ -16,14 +16,39 @@
 
 
 def set_workflow_state():  # pragma: no cover
-    """Set a workflow state for a given resource.
+    """Set a workflow state for a given resource."""
+    epilog = """
+Below are some usages examples. We assume there is process
```
xi: +a+ process

src/adhocracy_core/adhocracy_core/scripts/set_workflow_state.py
```diff
@@ -16,14 +16,39 @@
...
+is entered::
 
-        bin/set_workflow_state etc/development.ini <resource-path> <state>
-    """
+    ./bin/set_workflow_state etc/development.ini
+    /organisation/workshop evaluate result closed
+
+An absolute path can be given instead of a relative one with the
+`absolute` option. The following command will put the workflow in the
+'closed' phase, whatever the current state is::
```
xi: replace 'phase' by 'state'

src/adhocracy_core/adhocracy_core/scripts/set_workflow_state.py
```diff
@@ -16,14 +16,39 @@
...
+
+    ./bin/set_workflow_state --absolute etc/development.ini /organisation/workshop announce participate evaluate result closed
+
+The current state and information about the workflow can be obtained
+with the `info` option::
+
+    ./bin/set_workflow_state --info etc/development.ini /organisation/workshop
+
+To `reset` option is used to reset the workflow before setting the state::
```
xi: replace 'To' by 'the'

xi:
LGTM.

PR #1580, opened 2015-08-19
Add Bürgerhaushalt Workbench and Context

xi:
I implemented the full workbench and context for bürgerhaushalt. I did this by first doing some minor refactoring in `Kiezkassen` and then basically copying it over to the `Burgerhaushalt` package.

There might be some small things missing. But apart from that, it should be mostly complete. Note however that I did not yet include a phase widget.

Please take **special care** to test d45578c and 3ff2037 as I did not yet have time to do that yet.


xi:
This broke kiezkassen because the module name `adhMeinBerlinWorkbench` was used twice. Angular seems to silently fail when a module with an existing name is registered. Fixed in 79bf2d2.

PR #1581, opened 2015-08-20
Make bplan numbers strings, not integers.

pallix:
This allow to have bplan numbers such as "1-35ba".


joka:
Is the frontend still working?

joka:
> Is the frontend still working?

frontend should adapt automatically to this tye change

pallix:
Frontend is working

PR #1582, opened 2015-08-20
Allow different proposal versions to be badged

pallix:
@MagdaN it should solve the problem


xi:
Description of the problem by @MagdaN:

> it is impossible to add a badge to a version if the same badge was added to an earlier version.

PR #1586, opened 2015-08-24
2015 08 jk index workflow state

joka:
API Extension:

- new filtering pool search index: 'workflow_state':  workflow state, for versions it indexes the state of the item.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/catalog/adhocracy.py
+++ b/src/adhocracy_core/adhocracy_core/catalog/adhocracy.py
@@ -46,7 +46,7 @@ class AdhocracyCatalogIndexes:
     rates = catalog.Field()
     creator = catalog.Field()
     item_creation_date = catalog.Field()
-    workflow_state = catalog.Keyword()
+    workflow_state = catalog.Field()
     reference = Reference()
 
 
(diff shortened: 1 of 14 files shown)


src/adhocracy_core/adhocracy_core/catalog/adhocracy.py
```diff
@@ -137,6 +142,23 @@ def index_badge(resource, default) -> [str]:
...
+
+def index_workflow_state_of_item(resource, default) -> [str]:
+    """Find item and return its value for the workflow_state index."""
+    item = find_interface(resource, IItem)
+    try:
+        state = get_sheet_field(item, IWorkflowAssignment, 'workflow_state')
+    except (RuntimeConfigurationError, AttributeError):
+        return default
+    else:
+        return [state]
```
pallix: why return an array instead of the value?

joka:
> why return an array instead of the value?

its a list not an array to be precis (I didnt know about python arrays until last week ...)
The "workflow_state" index is a Keyword index, so we need a iteratable here.
But there can be only one workflow state, so we should change this index to a FieldIndex,.

pallix:
LGTM

PR #1587, opened 2015-08-24
Fix autonaming pools

### LINKED ISSUE #1583: Users created with path/name containing the date
Users get created with a name containig the date.

https://embed-meinberlin-stage.liqd.net/api/principals/users/
https://spd.liqd.net/api/principals/users/

When using autonoming the indexes used to name the children of a pool are incrementally for each creation. If the name already exists the current date is appended to the name to avoid a conflict.

This mean we have conflicts when creating users.

The bug is probably due to several threads executing the pool code at the same time.

pallix:
@slomo All installations should be updated with:

`./bin/sd_evolve  --mark-unfinished adhocracy_core.evolution.change_pools_autonaming_scheme etc/development.ini`

`bin/sd_evolve --latest etc/development.ini`

@joka can you double check that the evolution script is right?

Should fix  #1583.


joka:
lgtm

PR #1590, opened 2015-08-24
2015 08 jk make items badgeable

joka:
API Extension:

- add filtering pool index 'item_bagde (index badge names of item for all versions)


joka:
merge chaos

PR #1591, opened 2015-08-24
2015 08 jk make items badgeable+3

joka:
API Extension:

add filtering pool index 'item_bagde (index badge names of item for all versions)


pallix:
/home/alphonse/projects/adhocracy3.mercator/src/adhocracy_meinberlin/adhocracy_meinberlin/resources/test_bplan.py is failing

Also master needs to be remerged

PR #1594, opened 2015-08-25
allow to set cameFrom in URL

xi:
This is useful when you want to embed some login protected functionality: You can in that case embed the login form and set `cameFrom`.

Another option to get to the same behavior would be to directly embed the protected feature. The routing should then redirect to login and set `cameFrom` automatically. (I only realised that the second option exists when writing this. I still think the code could be helpful).


nidico:
We just discussed that we don't merge that for now as we don't need it currently and as it might interfere with the routing. We can recycle it when we need it.

PR #1596, opened 2015-08-25
Enable users by login name resolution for import

pallix:
Allow to resolve user by login name in the import_resource script.
Strings having the form 'user_by_login: <username>' are resolved
to the user's path.

@slomo should make deployment easier


pallix:
@joka can you review?

PR #1597, opened 2015-08-25
check register permissions

xi:
This only shows register UI if registration is allowed. Activation is still available.


nidico:
Is there already an easy way to change the permission to register in the backend in order to test that?

xi:
> Is there already an easy way to change the permission to register in the backend in order to test that?

Don't know.

xi:
I just merged #1620 so this can now be tested with the kit package.

nidico:
This needs a merge / rebase.

Issue #1546, opened 2015-08-11
Address field on kiezkassen sheet not used


pallix:
The address field in the kiezkassen sheet is not used by the frontend.

I guess it was modelled but the requirements changed during development. We could remove it with a migration script, if necessary.

xi:
Fixed by #1598

PR #1598, opened 2015-08-25
rm traces of address from frontend

xi:
*see #1546*

is not used in the UI anyway and is up for deprecation in the backend.


nidico:
The typescript does not build.

xi:
> The typescript does not build.

It seems like address is required in the backend. So this is on hold until the backend has been fixed.

xi:
Could someone please review this? @pallix?

pallix:
@joka don't we need a migration?

joka:
> @joka don't we need a migration?

If we have never set values for this field, we don't need a migration

pallix:
Good. We can merge then.

PR #1599, opened 2015-08-26
migration frontend to new IWorkflowAssignment.state_data

### LINKED ISSUE #1589: frontend needs migration of IWorkflowAssignment.state_data
In #1464 the `IWorkflowAssignment` sheet has been changed. The information regarding specific workflow states has been moved to a new `state_data` field.

The frontend has not yet  been adopted to that change. In some places this has been attempted (e.g. d45578cdab9d517907700c5b2e6d2e6da558ed8c) but not everywhere and not in a sufficient way.

-   `state_data` is a list rather than an object, so `state_data.announce` does not do what we want it to do. 
-   entries in `state_data` are optional. So we can not assume that data for a specific state exists.

xi:
*Fixes #1589*


nidico:
LGTM, works.

PR #1600, opened 2015-08-26
Added show results to rate widget

local-girl:
rate widget now accepts modifiers 

m-show-results-persist // results always show
m-show-results-click //results show on vote
m-show-results-hover //results show on hover


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
```diff
@@ -166,6 +236,8 @@ the `is-disabled` state.
     }
 
     .rate-difference {
+        vertical-align: top;
```
xi: This does not seem to make a difference. I would also prefer baseline alignment with the buttons if that is possible.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Rate/Rate.html
```diff
@@ -8,24 +12,30 @@
...
-        href=""
-        class="rate-contra"
-        data-ng-class="{
-            'is-rate-button-active': myRate === -1,
-            'is-disabled': disabled || (!optionsPostPool.POST && optionsPostPool.loggedIn)
-        }"
-        data-ng-click="toggle(-1)">
-        <i class="icon-vote-down" title="{{'TR__RATE_VOTE_DOWN' | translate}}"></i>
-    </a>
+    <div class="rate-button-container">
```
xi: Until now the rate widget was `inline-block`ish. Can we use spans instead of divs?

xi:
This looks good! However, I have still two issues with it:

-   Before this change, rate widets were properly aligned inline elements (except for the floating in meinberlin:

    rate_aligned

    I already proposed to remove the `vertical-align: top` and make `.rate-button-container` `inline-block`. But even then the alignment is not what it used to be:

    rate_unaligned

    I guess a `position: absolute` solution could be done that preserves the original alignment. But that would have other issues. Maybe we can find a solution with a length value in `vertical-align` (see https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align).

   This is not a total must have, but would be very nice.

-  In the disabled state, the results are never shown. But they still take up place. A quick fix would be to hide them completely in that state. But the proper solution would probably be to show them somehow. However, IIRC there are no designs for that.

    rate_disabled

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Rate/Rate.ts
```diff
@@ -318,6 +321,7 @@ export var directiveFactory = (template : string, adapter : IRateAdapter<RIRateV
                             })
                             .then(() => undefined);
                     }).finally<void>(() => {
+                        scope.isCast = true;
```
xi: This way the results will only show up when I click. When I reload the page I will still have rated, but I cannot see the results anymore. Wouldn't it be better to set `isCast` in `updateMyRate()` so it reflects whether I have *rated* rather than whether I have *clicked*? (In that case we should also rename the modifier to `m-show-results-cast`)

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
local-girl: Removed

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Rate/Rate.html
local-girl: Done

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Rate/Rate.ts
local-girl: I'm not sure about that. I have seen this many times implemented as "click to see the results" I think seeing the results can effect how you vote. m-show-results-cast seems like additional functionality + I'm not sure we even need m-show-results-click !

xi: Ok, so we agree that `m-show-results-click` does not make too much sense in its current form. But we can just leave it in in case it becomes helpful.

xi:
Works fine.

PR #1606, opened 2015-08-27
refactor markdown

xi:
- avoid using implicit global variable `angular`
- use jquery instead of `quereySelector` for improved compatibility
- avoid markdown-it errors on undefined `parsetext`


nidico:
LGTM.

PR #1607, opened 2015-08-27
mv mercator specific code to mercator package

xi:
There was mercator specific code in core which was then overwritten by a more general version in spd. This fixes that by having the general code in core and overwriting it in mercator.


nidico:
LGTM.

PR #1612, opened 2015-08-31
comparators for filtering pool parameters API EXTENSION

joka:
Without comparators:

Example:

     testapp.get('/Documents/document_0000000',  params={'name': 'VERSION_000000'})

Comparators with single search value::

   eq, noteq, gt, ge, lt, let

Example::

     testapp.get('/Documents/document_0000000',  params={'name': '["eq","VERSION_000000"]'})

Comparators with multiple search values::

    any, notany

Example::

    testapp.get('/Documents/document_0000000/rates/',  params={'name': '["any", ["rate_0000000", rate_0000001"]]'})

For more information see docs/rest_api.rst


nidico:
One `pyfunc` test fails after merge, @joka - could you please check?

joka:
All bugs should be solved now.  This request parameter validation should now also work for indexes without example value.

PR #1613, opened 2015-08-31
Add Burgerhaushalt phase header

xi:
This merges the phase headers from meinberlin/kiezkassen and
meinberlin/alexanderplatz and creates a new one for burgerhaushalt.

Note: The description texts are likely to change.


nidico:
Works nicely.

PR #1615, opened 2015-08-31
Implement burgerhaushalt badge styling

xi:
Still missing: Matching the new modifiers to the actual badges.


local-girl:
@xi if you like my changes then merge it

PR #1623, opened 2015-09-01
Single a3.scss file

local-girl:
Instead of having several a3.scss files with mostly the same info we just have one that imports a `_theme.scss` file - it seems to work well. **I urge you to re-check the styles** this is quite a large re-structure as the order of scss files has changed + I did everything manually so there might be forgotten files.

1. As you can see I added `map` and `document` to core, thoughts ? It's not needed for every project but I feel its such a light weight it might as well be there
2. I also looked at <https://rubygems.org/gems/sass-globbing/versions/1.1.1> which could further simplify things bu just saying `@include "components/*"`


xi:
Nice! This is a big improvement, and with something like sass-globbing it could get even better.

I did not find any issues. If we spot any, I do not expect that they are hard to fix.

Importing `map.scss` and `document.scss` is fine IMO, becasue they are already in core anyway, just not imported.

Issue #1593, opened 2015-08-25
Local moderator can't view other public resources


slomo:
As a local moderator for a process I can not view any resources outside the process, even though they are readable for participants and non logged in users.

This has been observed on alexanderplatz live, you can use alexanderplatz stage to reproduce.

Issue #1549, opened 2015-08-11
Specify how to model roles with conflicting permissions


pallix:
When defining an ACM where mutiple roles have conflicting permissions, for instance Moderator with 'view' 'Allow' and Participant with 'view' 'Deny', there is no clear way to define which permission should take the priority.

See #1498 for a concrete example of the problem

joka:
This is something we cannot solve completly, guess it helps if we:
   - try not to set explicit deny permissions
   - set workflow acm in the "right" order: first "basic" roles then more powerfull  roles

joka:
Can you list where we have this kinde of permission problems?

pallix:
@xi suggested to prioritize roles, for example moderator would takes precedence on participant, always.

I had to put moderator before participant in the [alexanderplatz workflow](https://github.com/liqd/adhocracy3.mercator/blob/master/src/adhocracy_meinberlin/adhocracy_meinberlin/workflows/alexanderplatz.py#L15) but I don't find the solution elegant. It's not self describing.

PR #1624, opened 2015-09-01
fix permission problems

joka:
Backend fixes:

 - #1593  local moderators have no view permission 
 - undo "Add 'authenticated' group (role:participator) to all imported users, we don't want 
   to force  every user to have the participator role.
   This default group won't be necessary if we have a concept of "Membership" for organisations/processes.
 -   #1549  make high priority role permissions override low priority role permisssions


joka:
@slomo if we import a user with a special group (like for moderators), he does not implicit get the participator role anymore.
If no group is set, the user gets the groups 'authenticated' aka "Members" with the participator role.

nidico:
LGTM.

src/adhocracy_core/docs/authorization.rst
```diff
@@ -62,12 +62,17 @@ Roles with example permission mapping:
...
+# TODO the role definition is outdated
+
 Mappings of principals to local roles are associate with resources and
 are inherited within the object hierarchy in the database.
 The creator is the principal who created the local context.
 The creator role is automatically set for a specific local context and is not
 inherited.
 
+Permission mappings for roles with high priority override those with lower
+priorty. The order is determined by :var:`adhocracy_core.schema.ROLE_PRINCIPALS`
```
pallix: typo

PR #1625, opened 2015-09-02
Removed text mixin, removed other useless things from variables

local-girl:
We barely use the text mixin it is just a legacy plugin - have removed it in preparation for greater things (EG rem-sizing)


xi:
LGTM. (yeay, refactoring!)

PR #1626, opened 2015-09-04
Change default autobahn port to 6561

nidico:
Port 8080 is frequently used by other tools (e.g. tomcat, which is
installed in the policycompass project) and may conflict with autobahn.

The intention of using 8080 for autobahn is that this is on some
firewalls not blocked, but this doesn't matter for us anyway as we
typically put nginx in frontend of autobahn and use a dedicated path
/websockets for autobahn.

Port 6561 is somewhat inline with 6551 and 6541.

@slomo: please check whether this is relevant for some manual deployments.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -70,7 +70,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   config.vm.network :forwarded_port, guest: 6541, host: 6541
   config.vm.network :forwarded_port, guest: 6551, host: 6551
-  config.vm.network :forwarded_port, guest: 8080, host: 8080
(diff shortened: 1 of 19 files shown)


xi:
Just to be sure: An alternative solution would be to change the config for deployment, right?

slomo:
No the deplyoment already runs autobahn on a different port (expect for policycompass). I guess this is a feature for @nidico localsetup.

nidico:
> I guess this is a feature for @nidico localsetup.

Not for me, but for fellow pcompass devs which used the pcompass Makefile and have now tomcat running by default.

joka:
lgtm

PR #1636, opened 2015-09-09
refactor: change module structure

### LINKED ISSUE #1189: FE unittests can not be run on node.js
We originally planned to run frontend unittests both in the browser and with node.js. This has the following benefits:

-   better tool support
    -   e.g. CI servers do not need to a browser to run the unittests
-   we can make sure that our code does not depend on browser features directly (use dependency injection instead). That improves modularity and eliminates site effects in testing.

Unfortunately, this did not work too well. The documentation for running unittests with node was actually removed in 063161b6.

The main issue with node is currently that some modules import third party libraries via require.js. If you try to run these with node this will result in an exception like this:

    { [Error: Cannot find module 'socialSharePrivacy'] code: 'MODULE_NOT_FOUND' }

Note that the third party modules are not actually used in the tests (at least they should). This also messes up coverage reports and similar.

I propose to change the module structure to only load the required modules for tests. The modules should be split into one containing the actual code and another one that wires that to third party libs and everything else. So basically all the code would be moved to a separate module (I propose `Code.ts`) and only the `register()` function would remain. The tests would than only import `Code.ts`.

xi:
*fixes #1189*

The idea and motivation for this have been outlined in #1189 and in the documentation (see first few commits). The bulk of this pull request is about migrating the existing code. The last few commits are about trying jasmine-node with the new structure.

Sorry for the huge PR. I had planned to split it, but as this is essentially about the *relations* between packages it was hard to find a place where a split made sense.

As this is guaranteed to conflict with other frontend changes, I would be happy if this could be merged soon.

There is a lot more to be done, e.g.

-   actually use jasmine-node
    -   update documentation
    -   integrate with polytester
-   some module names are extremely long because we keep adding pre- and suffixes to them (e.g. `AdhMeinBerlinAlexanderplatzWorkbenchModule`). I chose to use shortened versions in some places. We should find a solution and resolve existing inconsitencies.
-   Write more and better tests
-   In MeinBerlin we chose to create sub-packages, while for Mercator and spd we prefixed the package names. There should be a consistent solution.
-   We need a better overview of our packages and their relations
    -   visualise the graph of packages and relations
    -   I think there are different kinds of packages (e.g. workbenches, contexts, resource-related [e.g. proposal or document], ...). We should make this structure explicit.


nidico:
Looks very good. I don't have time to test it properly, but the tests pass and we don't have to deploy anything in the upcoming days, so I'm merging this now.

PR #1637, opened 2015-09-09
Fix timeing issues in CrossWindowMessaging.sendLoginState

### LINKED ISSUE #1635: CrossWindowMessaging.sendLoginState doesn't send userData
`CrossWindowMessaging.sendLoginState` is expected post a message with an object with a `userData` key. This isn't sent due to the user object not being fetched yet (in `AdhUser.loadUser`), but `sendLoginState` being triggered by watching `adhCredentials.loggedIn`, which is set earlier.

xi:
*fixes #1635*

Untested, but should work.


nidico:
Works after a small fixup.

PR #1638, opened 2015-09-09
replace deps2dot.sh by a more advanced python script

xi:
builds on #1636


nidico:
Looks nice, didn't check correctness.

PR #1639, opened 2015-09-09
switch to ES6 module syntax

xi:
*this build on #1636*
*implements part of #1449*

ES6 module syntax has been available in TypeScript since [version 1.5](http://blogs.msdn.com/b/typescript/archive/2015/07/20/announcing-typescript-1-5.aspx).

It has no real advantage other than being a standard. But that already means that developers with no TypeScript experience but knowledge of ES6 will feel right at home.


nidico:
Nice!

PR #1647, opened 2015-09-14
remove needless dictonaries (save 15% database size)

joka:
- use attribute storage for resources with rate sheet 
- use attribute storage for autoname last counters 
- rm needless autoname last  counters
- store data for sheets with annotation storage in attributes referencing a dictionary

UPDATE NOTE:

3 long running migration scripts added,
backend server should be shutdown or readonly ( @slomo ).
Check that all migration script pass.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/evolution/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/evolution/__init__.py
@@ -24,6 +24,7 @@ from adhocracy_core.sheets.title import ITitle
 from adhocracy_core.sheets.badge import IHasBadgesPool
 from adhocracy_core.sheets.badge import IBadgeable
 from adhocracy_core.sheets.principal import IUserExtended
+from adhocracy_core.sheets.rate import IRate
 from adhocracy_core.sheets.workflow import IWorkflowAssignment
 from adhocracy_core.resources.pool import IBasicPool
 from adhocracy_core.resources.asset import IPoolWithAssets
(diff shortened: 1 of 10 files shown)


pallix:
LGTM

joka:
One question remains:  Do we want to stick with AnnotationStorage for sheet data and use AttributeStorage only as a exception or switch to AttributeStorage for all Sheets?


Reasons for AnnotationStorage (additional dictionary to store sheet data):

- allow conflicting field names
- low memory profile for the resource object
- more easy to change the stoarge engine for sheet data (we are forced to call the sheet object for read/write data)

Reasons for AttributeStorage(store sheet data directly in resource attributes):

-  simple, like in substanced 
- more easy to read data (we are not forced to call the sheet object)
-

@nidico whats your opinion in this?

pallix:
How important it is to avoid conflicting name? In OOP, conflicts can also occurs when implementing different interfaces having methods with the same name but practicaly this rarely happens (to me at least).

On the other end I don't find having all the value as attributes attached to the object very elegant. Is your point "more easy to read data" important since reading a sheet is abstracted by a function?

joka:
> How important it is to avoid conflicting name? In OOP, conflicts can also occurs when implementing different interfaces having methods with the same name but practicaly this rarely happens (to me at least).

I don't think this is important,  we don't have many fields anyway.

> On the other end I don't find having all the values as attributes attached to the object very elegant.
 
> Is your other point "more easy to read data" important since reading a sheet is abstracted by a function?

For some data (visibitlity, user group_names,/email/name) we don't use the abstraction, instead we use direct attribute access  to improve performance. If  we use AttributeStorage for all sheets we could allow this in general, to simplify things and have similar behavior like in substanced.

For me the main point to stick with AnnotationStorage is the last point:
  
-  "more easy to change the stoarge engine",   (If we take this serious we should be more strict in using the abstraction for reading/writing sheet data)

I'm just unsure if this is a more theoretical or real  benift.

pallix:
"more easy to change the storage engine" => We need to discuss with more people internally as some have express some doubt, founded on not, on the scalability of ZODB so it may be a good idea to not close the option of having a different storage engine now.

nidico:
Two remarks:

- I'd strongly advise against not allowing a resource to have multiple sheets which have fields with the same name. I'm not sure whether this restriction already applies in this PR. Couldn't the attributes also contain the sheet name as a prefix?
- I wouldn't worry about switching to other databases right now. If we really want to do that, we can also switch back to AnnotationStoarge then.

joka:
> I'd strongly advise against not allowing a resource to have multiple sheets which have fields with the same name. I'm not sure whether this restriction already applies in this PR.
no, but I don't see any problem with this. In the real world we  don't have any fields with the
same name, and we could add a validator to check this. 

> Couldn't the attributes also contain the sheet name as a prefix?
yes, this would also be possible

PR #1651, opened 2015-09-14
Integrate jasmine-node in polytester

### LINKED ISSUE #1278: frontend unit tests not executed
I noticed that in all our recent travis builds, jasmine for unit tests is executed, but it reports 0 test cases. I tried to debug the issue but did not come up with much:

-  It seems to be slightly undeterministic whether the tests are found or not.
-  `jsApiReporter.finished` reports that jasmine is done. If I change the check to `jsApiReporter.finished && jsApiReporter.specs().length` it actually reports test cases, but not necessarily all of them.
-  I checked the jasmine changelog. We are some versions behind, but the issue still exists with the latest version
-  The only recent change that might be related was the blanket update in #1275. But that does not seem to have caused the issue.

xi:
*fixes #1278*

In this pull request I also removed dead test code and added a softlink `src/current/` that always points to the currently build frontend package. The latter change was required in order to not have to generate `tests.yml` with buildout.

*Note*: We might want to run these tests in a (or even more than one) browser some day. I still removed the browser integration for now as it was buggy and this whole topic needs to be discussed anyway.


xi:
I think this creates an empty directory `src/*/`. Will investigate.

nidico:
Works nicely.

Note that polytester states success even if jasmine-node errors, as jasmine-node returns error code `0` on error and `1` on fail, see mhevery/jasmine-node#353. Note that `--captureExceptions` fails in our case, should be checked separately.

PR #1655, opened 2015-09-16
Attempting to add inline badge editing

local-girl:
I've created a new module and you can see it "working" if you go to edit a document, however I feel I'm a long way off making it update badges I get stuck in the bindPath - I'm afraid I can't complete this alone, need help


local-girl:
Going to try again with a new branch

PR #1657, opened 2015-09-17
Prevent creating empty first versions within batch request (save 10% database size)

joka:
depends on  #1647

API EXTENSION:

- Update first version instead of creating a new version within a batch request. This  prevents creating an empty first version for every item we create.

- migration script to delete exsiting first versions without any data.

UPGRADE NOTE:

Long running migration script that delets resources, have a backup by hand. @slomo .


pallix:
Comment above, LGTM otherwise.

xi:
Does this require changes in the frontend?

joka:
> Does this require changes in the frontend?

no, but after migration we should do a quick manual test if commenting works.

joka:
@pallix  I fixed the docs/comments and rebased master, can you merge?

PR #1658, opened 2015-09-17
Add resource for policycompass

pallix:


xi:
I am not sure if it is a good idea to call this a "policycompass process" rather than a "policycompass request process".

pallix:
@joka by the way, proposals from core can't be posted to the process in core [1], which is why I created this resource. Seems strange to me.

Maybe we need core process with a standard workflow having core proposals. This could cause a problem for the tests since they don't set the workflow state at all.

[1] which has the 'sample' workflow, not standard.

pallix:
@xi you want `IRequestProcess` and `request_process_meta` in the code instead?

xi:
> you want `IRequestProcess` and `request_process_meta` in the code instead?

I am not sure about `request_process_meta`, but I would rather like to have `adhocracy_pcompass.resources.request.IProcess` than `adhocracy_pcompass.resource.pcompass.IRequestProcess`. (Like in meinberlin)

pallix:
Done.

nidico:
LGTM.

PR #1660, opened 2015-09-17
Initiate new process

local-girl:
This is a straight copy of kiezkassen with find and replace to stadtforum. I commented out with FIXME all the references to missing backend files for now so the .ts builds. This is as far as I can get without a backend


xi:
While working on this branch I realized that we do not really need a full workbench but just a single widget. So I created #1718 instead.

PR #1666, opened 2015-09-21
Add resources for stadtforum

pallix:
- Add `IPolarizable` sheet in core with attribute `position` that can be
  'pro' or 'contra'
- Add IPolarizedComment resource in the meinberlin.resources.stadforum
  module
  IPolarizedComment inherits IComment so comments pools should be
  reusable for polarized comments.


pallix:
@xi have a look at the new resources

xi:
I get the following in the meta API:

```json
"adhocracy_core.sheets.polarization.IPolarizable": {
  "super_types": [
    "adhocracy_core.interfaces.IPredicateSheet"
  ],
  "fields": [
    {
      "editable": false,
      "name": "position",
      "valuetype": "adhocracy_core.sheets.polarization.Position",
      "creatable": false,
      "readable": true,
      "create_mandatory": false
    }
  ]
}
```

IMHO `editable`, `creatable` and `create_mandatory` should all be true, or am I missing something?

xi:
This seems to be reasonable to me. Some aspects where I am not sure:

-  Should we modifiy the existing `IComment` instead of adding `IPolarizedComment `?
-  Is "polarization" a good term for this?
-  Should the sheet be more flexible? e.g.:
   -   allow for future extensions such as "neutral"
   -   model as relation **to** something rather than just some attribute

pallix:
"Should we modifiy the existing IComment instead of adding IPolarizedComment?" => it's a good question.
Also, if we answer not, do we want IPolarizedComment in core?

"Should the sheet be more flexible?" => do we have any requirements for that currently?


"Is "polarization" a good term for this?" => the term [polarization](http://dictionary.reference.com/browse/polarization) is often used in politics and debates

joka:
fine

joka:
> "Should we modifiy the existing IComment instead of adding IPolarizedComment?" => it's a good question.
> Also, if we answer not, do we want IPolarizedComment in core?

yes, we can use this in many use cases.

joka:
> allow for future extensions such as "neutral"

I don't see the use case yet.

> model as relation to something rather than just some attribute

+1 For me this is an important point. What we want to express is:
User X agrees/unagrees to statement Y.  Its good to do some explicit semantic modeling
with references. This  make  it easier to analyse our data or add a  RDF  representation.

joka:
When i talk with pierre about the modeling last week I had the following resources in mind:

     <agreeable> <-- <agree> --> <user> 
     <agreeable> <-- <unagree> -->  <user>

unagree/agree  has the follwoing sheets:

 - description
 - commentable

src/adhocracy_core/adhocracy_core/sheets/polarization.py
```diff
@@ -0,0 +1,47 @@
...
+
+class PolarizableSchema(colander.MappingSchema):
+
+    """
+    Polarizable sheet data structure.
+
+    `position`: the position in the debate, 'pro' or 'contra'.
+    """
+
+    position = Position()
```
joka: IPredicateSheet sheets have a "predicate" name like like, vote, and the reference fields subject and object.
This is missing here.

src/adhocracy_core/adhocracy_core/sheets/polarization.py
```diff
@@ -0,0 +1,47 @@
+"""Polarization sheet."""
+import colander
+
+from adhocracy_core.sheets import sheet_meta
+from adhocracy_core.interfaces import IPredicateSheet
+from adhocracy_core.sheets import add_sheet_to_registry
+from adhocracy_core.schema import AdhocracySchemaNode
+
+
+class IPolarizable(IPredicateSheet):
```
nidico: No need to use a `PredicateSheet` here, as the `refers_to` field from the `IComment` sheet *is* the object, and the creator of the comment is the subject.

src/adhocracy_core/adhocracy_core/sheets/polarization.py
nidico: > IPredicateSheet sheets have a "predicate" name like like, vote, and the reference fields subject and object.

See my other comment, IMO we don't need this to be a *predicate*.

nidico:
Semantically, the `IPolarizedComment` is equivalent to specific `Agree` / `Disagree` resources @joka proposed above. I'm totally fine with how they're modelled, apart from that `IPolarizable` shall not be a `IPredicateSheet`.

joka:
> Semantically, the IPolarizedComment is equivalent to specific Agree / Disagree resources @joka proposed above. I'm totally fine with how they're modelled, apart from that IPolarizable shall not be a IPredicateSheet.


This is only true if you know what is actually meant. For example if you just traverse the references you would never find these pro and contra resources.

nidico:
> For example if you just traverse the references you would never find these pro and contra resources.

Why not query for PolarizedComments and filter / aggregate by position?

> Also the creator is not equal to "Subject" if we want to add delegation or "offline" content.

Maybe true. I haven't thought about these two points yet, but don't have time to do so right now...

joka:
> Why not query for PolarizedComments and filter / aggregate by position?

WIth traverse i meant follwoing the sheet reference and call the the reference target sheet.

Anyway, for me the modelling is not about how to do it with minimal effort, but  how to do it nice on the semantic level and to work with the data we produce in the long run. And reusing the comments does not feel good in this sense. I propose we just vote for the possible scenarios: 

 A resue comments 

    <commentable>  <--refers_to-- <comment>(+<polarizable>)

  B special resource type with predicate sheet

      <polarizable> <--object--<polarize>--subject--> <user>

  C special resource type with more simple sheet

      <polarizable> <--?polarize?--<polarize>


I`m  in favor of B or C.

nidico:
A practical argument in favor of modelling *polarizable comments* as comments is that the frontend wants to get all comments or a count of all comments (no matter whether "polarizable" top level comments or normal comments further down) in one request.

pallix:
For solution B, in the case of having comment expressing an pro/contra opinion, how would be the relation between the new resource type and the comment modelled?

nidico:
> Also the creator is not equal to "Subject" if we want to add delegation or "offline" content.

The case for offline content is valid, however this applies to *all* our content and should be solved on a more general level once we know what we really want.

Delegations: I don't really know what you have in mind here. Delegating whether a comment agrees or disagrees with the proposal doesn't make much sense. What makes sense is

- delegating the *main* agreement with the proposal (which shall be modelled as normal ratings anyway and would just work)
- delegating the rating of comments (which are also modelled as normal ratings and would just work) - however IIRC  the comments in Stadtforum shall not be rated, though I don't know why.

joka:
> For solution B, in the case of having comment expressing an pro/contra opinion, how would be the relation between the new resource type and the comment modelled?

The polarization resource type is commentable

joka:
> A practical argument in favor of modelling polarizable comments as comments is that the frontend wants to get all comments or a count of all comments (no matter whether "polarizable" top level comments or normal comments further down) in one request.

I contrast this is a argument against version A. 
 If you search for comments you get all comment plus all polarizations (aka subtype of comment), but this is most likely not what you expect. 
In my view Polarizations are no comments,  but can itself be commented.

joka:
> A practical argument in favor of modelling polarizable comments as comments is that the frontend wants to get all comments or a count of all comments (no matter whether "polarizable" top level comments or normal comments further down) in one request.

I you really want this in version B/C  you can just search for IComment and IPolarization

pallix:
"The polarization resource type is commentable" => isn't it problematic for the frontend to have two resources which are almost similar but are not?

joka:
> "The polarization resource type is commentable" => isn't it problematic for the frontend to have two resources which are almost similar but are not?

From technical point of view I don't see any disadvantage or advantage.  Besides many resource types are almost similar.

joka:
I would like to finish the resource modelling this week, so the frontend can start working.
 If i implement Version C  in my free time, are there any objections?

pallix:
when do you want to begin? I'm currently working on modelling polarisation similar to rates

joka:
> when do you want to begin? I'm currently working on modelling polarisation similar to rates

I can do this today or tomorro, but if you work on this its also fine.

If you model version B (rate like, predicate sheet), I would record Nicos note about the duplicated user reference (predicate sheet subject && metadata creator references).  This  leads to Version C (only metadata creator reference).

joka:
> I can do this today or tomorow, but if you work on this its also fine.
> 
> If you model version B (rate like, predicate sheet), I would record Nicos note about the duplicated user reference (predicate sheet subject && metadata creator references). This leads to Version C (only metadata creator reference).
> 
> The main advantage of the predicate sheet is to play freely with the subject/creator and make it easy to add a more advance reference search index for all predicate sheets.

I'm indifferent between B and C and need some breakfast, do as you think is best.

nidico:
@joka: I'd love to have an offline conversation about that as being able to draw things on paper would help.

pallix:
Ok so my current implementation is solution B, maybe with some missing parts. Do we need new indexes for the catalogs for instance?

xi:
Sorry to be late to the party, but I am strongly against option B. and in favour of option A for the following reason:

As far as I understand, we have rates to express a persons opinion on a topic. Polarizations on the other hand are meant to give more detail on the relation between a comment and a commentable. Or more explicitly: Polarizations don't say anything about the relation of the author to the commentable.

Imagine for example there is a proposal that I am not yet sure about. I might add several comments, some agreeing and some disagreeing, in order to make up my mind.

Another option that would express this concept would be to have the polarization not as a sheet on the comment itself but as a separate resource like so (option D):

```
<commentable>  <--refers_to--  <comment>
          ^                      ^
           \                    /
          object            subject
              \              /
               <polarization>

**Comment by joka** [comment]
> Sorry to be late to the party, but I am strongly against option B. and in favour of option A for the following reason:

Some praticall reasons agains A:

- You cannot search for comments but not for polarizations.
- You cannot add a polarization without being forced to add a comment text.
- The sematic is not expressed explicit in my point of view (see below and above)
 
> As far as I understand, we have rates to express a persons opinion on a topic. Polarizations on the other hand are meant to give more detail on the relation between a comment and a commentable. Or more explicitly: Polarizations don't say anything about the relation of the author to the commentable.

In my understanding its the oposit. Polarizations express like rates the persons opion to a topic: "This statement is true or not". This concept may be usefull in many use cases and can also be adapted for the stadtforum survey: ak "The answer to this question is yes or no (or both yes and no)". 
Optionally the polarization relates to a descriptive text. In Version B/C this is the description sheet, but this could also be a relation to comment or propsal.

> Imagine for example there is a proposal that I am not yet sure about. I might add several comments, some agreeing and some disagreeing, in order to make up my mind.
> 
> Another option that would express this concept would be to have the polarization not as a sheet on the comment itself but as a separate resource like so (option D):

This is also fine for me.

**Comment by xi** [comment]
> You cannot add a polarization without being forced to add a comment text.

IMO this should not be possible indeed.

**Comment by joka** [comment]
> As far as I understand, we have rates to express a persons opinion on a topic. Polarizations on the other hand are meant to give more detail on the relation between a comment and a commentable. Or more explicitly: Polarizations don't say anything about the relation of the author to the commentable.
> In my understanding its the oposit. Polarizations express like rates the persons opion to a topic: "This statement is true or not". This concept may be usefull in many use cases and can also be adapted for the stadtforum survey: ak "The answer to this question is yes or no (or both yes and no)". 
> Optionally the polarization relates to a descriptive text. In Version B/C this is the description sheet, but this could also be a relation to comment or propsal.

But you are right, we could use the Rate resource to express the users opionion and the polarization
to express on kind of relation between statements. 
Then I'm for Version D.

Will we have similar relations in the future, like "This is an antithesis" or "This is similar"?

**Comment by joka** [comment]
Just to explain main different understanding of the semantic: 
When I did the modelling last week, the requirement was different.
The user could rate one time and optional add a comment to its rate.

**Comment by joka** [comment]
I implemented Version D. functional test are missing.

The backrefernce from the comment to the polarization is not shown.
We should add a field 'polarization' to the ICanPolarize sheet.

**Comment by nidico** [comment]
I'm fine with *D*.

File: src/adhocracy_core/adhocracy_core/sheets/relation.py [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+    `position`: the position in the debate, 'pro' or 'contra'.
+    """
+
+    subject = ReferenceSchema(reftype=PolarizationSubjectReference)
+    object = ReferenceSchema(reftype=PolarizationObjectReference)
+    position = Position()
+
+polarization_meta = sheet_meta._replace(isheet=IPolarization,
+                                        schema_class=PolarizationSchema,
+                                        create_mandatory=True)
```
joka: we should set implicit_addable to True, otherwise we cannot add the polarization to the relations service pool.

src/adhocracy_core/adhocracy_core/resources/relation.py
```diff
@@ -0,0 +1,82 @@
...
+
+polarization_meta = item_meta._replace(
+    content_name='Polarization',
+    iresource=IPolarization,
+    element_types=(IPolarizationVersion,
+                   ),
+    item_type=IPolarizationVersion,
+    use_autonaming=True,
+    autonaming_prefix='polarization_',
+    permission_create='create_comment',
```
joka: We should set ìmplicit_addable to True, otherwise we cannot add the polarization to the relations service pool.

pallix:
@joka the "Cannot create process content" for moderator functional test is failing because moderators can create comments, and relations have the create_comment permission. What do you think about it?

xi:
We may have the same issue as with badges here: Should the item or the version be polarizable? Or is this handled by our versioning system?

xi:
We need a backreference on the `ICanPolarize` sheet (that is the one on the comment, right?)

xi:
I should note that the frontend will probably ignore the `object` of the polarization as we have no concept for displaying it. We might check whether it corresponds with the `refers_to` from `IComment` and ignore the whole polarization if they don't.

BTW, is it veryfied that there is only ever one polarization for each pair of subject/object?

joka:
> @joka the "Cannot create process content" for moderator functional test is failing because moderators can create comments, and relations have the create_comment permission. What do you think about it?

This error might be related to my pull request that removes empty first versions in batch requests, sorry.

joka:
> BTW, is it verified that there is only ever one polarization for each pair of subject/object?

no, maybe for other use cases we want to allow multiple polarization relations.
For now  I would only care if there is one or None polarization

joka:
mmh, but  the PolarizationSubjectReference backreference is not ordered, so we should add a "there is only one" verification when we have time.

joka:
> I should note that the frontend will probably ignore the object of the polarization as we have no concept for displaying it. We might check whether it corresponds with the refers_to from IComment and ignore the whole polarization if they don't.

For a general solution it would be good to not depend on IComment in anyway. Any resource type can have the ICanPolarize sheet, not only comment.s

pallix:
Last commits fixes errors on the functional tests.

joka:
LGTM

Issue #1661, opened 2015-09-18
Adhocracy.min.ts does not contain all imports


xi:
The relatively recent https://spd-perspektivdebatte.liqd.net/static/js/Adhocracy.min.js (0.3.9) does not contain all (none?) imports. This still worked in 0.8.3.

xi:
This seems to be cause by the switch to UMD in d5e12f0.

xi:
Upstream fix has been merged (https://github.com/Microsoft/TypeScript/pull/4980). This should be fixed with the next TypeScript release.

xi:
TypeScript 1.7 has been released: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#typescript-17

PR #1667, opened 2015-09-21
update typescript and require.js

xi:
While debugging #1661 I tried to update TypeScript to [1.6](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#typescript-16) and require.js to [2.1.20](http://requirejs.org/docs/download.html#2-1-20). It did not help with that problem but is still useful.

Major change: TypeScript now checks some more types.


nidico:
Nice.

PR #1671, opened 2015-09-22
workaround #1661

xi:
TypeScript UMD and the require.js optimizer do not yet play well with each other. I submitted a fix at https://github.com/jrburke/r.js/pull/857. Until that fix is available, here is a workaround in adhocracy.


nidico:
Works.

PR #1674, opened 2015-09-24
Use polytester as full test suite command in INSTALLATION

nidico:


xi:
Maybe we should expand this section sometime, but should be fine for now.

Issue #1663, opened 2015-09-18
Improvements on import scripts


slomo:
I'd like to propose a few improvements to the import scripts

`bin/import_users`:
 - [x] add default values for groups and roles (currently they need to be given as empty list)
 - [x] add badges to user, even when existed before
 - shouldn't be badges be given by their path rather than their name (is that even unique?)
 - [x] what about permissions? (are they updated for existing users?)
 - [x] choose random password if no initial-password is given or even set no password (if that is possible)

`bin/import_ressources`:
- update sheet data (could be dangerous)

In general I don't know how much work should be put into this, given that most resources will hopefully edited through an admin interface in the future.

slomo:
@pallix Isn't that a think you like to do ;)

pallix:
@slomo "choose random password if no initial-password is give": how do you log? Or do you want to print the password on the console when it is generated? or this would be for the case when an invitation is sent?

pallix:
"shouldn't be badges be given by their path rather than their name (is that even unique?)" => badges are created during the import for this script, so technically they don't have a path.

pallix:
"what about permissions? (are they updated for existing users?)" => roles and groups get updated

pallix:
Related to #1390

PR #1678, opened 2015-09-24
Extend import_user script to generate password

pallix:
Allow to not specify initial-password field in import_users script.


pallix:
Related to #1663

PR #1681, opened 2015-09-25
Untie adhocracy_spd and adhocracy_kit config

slomo:
This is motivated by problems with having the spd and the kit subscriber running at the same time.


joka:
LGTM

PR #1683, opened 2015-09-25
avoid calling jasmine.execute() twice

xi:
blanket already calls `jasmine.execute()` event if require.js is not done loading the testdata. This is a potential cause for #1278.


local-girl:
LGTM

Issue #583, opened 2015-01-26
Duplicate tpc_begin calls for same transaction


xi:
When running `bin/py.test src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/test_user_login.py` I have the following error in the backend log on the first test (`test_register`):

```
2015-01-26 19:04:38,167 ERROR [adhocracy_core.rest.exceptions][MainThread] internal
Traceback (most recent call last):
  File "/home/tobib/a3/a3/eggs/pyramid-1.5.1-py3.4.egg/pyramid/tweens.py", line 21, in excview_tween
    response = handler(request)
  File "/home/tobib/a3/a3/eggs/pyramid_exclog-0.7-py3.4.egg/pyramid_exclog/__init__.py", line 115, in exclog_tween
    _handle_error(request, getLogger, get_message)
  File "/home/tobib/a3/a3/eggs/pyramid_exclog-0.7-py3.4.egg/pyramid_exclog/__init__.py", line 111, in exclog_tween
    return handler(request)
  File "/home/tobib/a3/a3/eggs/pyramid_tm-0.7-py3.4.egg/pyramid_tm/__init__.py", line 82, in tm_tween
    reraise(*exc_info)
  File "/home/tobib/a3/a3/eggs/pyramid_tm-0.7-py3.4.egg/pyramid_tm/compat.py", line 13, in reraise
    raise value
  File "/home/tobib/a3/a3/eggs/pyramid_tm-0.7-py3.4.egg/pyramid_tm/__init__.py", line 70, in tm_tween
    manager.commit()
  File "/home/tobib/a3/a3/eggs/transaction-1.4.3-py3.4.egg/transaction/_manager.py", line 111, in commit
    return self.get().commit()
  File "/home/tobib/a3/a3/eggs/transaction-1.4.3-py3.4.egg/transaction/_transaction.py", line 280, in commit
    reraise(t, v, tb)
  File "/home/tobib/a3/a3/eggs/transaction-1.4.3-py3.4.egg/transaction/_compat.py", line 55, in reraise
    raise value
  File "/home/tobib/a3/a3/eggs/transaction-1.4.3-py3.4.egg/transaction/_transaction.py", line 271, in commit
    self._commitResources()
  File "/home/tobib/a3/a3/eggs/transaction-1.4.3-py3.4.egg/transaction/_transaction.py", line 417, in _commitResources
    reraise(t, v, tb)
[...]

PR #1684, opened 2015-09-25
Update frontend dependencies

### LINKED ISSUE #1665: Upgrade jasmine-node to 2.0.0-beta4
We currently use jasmine-node version `2.0.0`, which makes one assume that it's newer than `2.0.0-beta4`. It isn't though. We should upgrade in order to clarify the mess.

xi:
this updates many frontend dependencies, e.g.:

-   [angular](https://github.com/angular/angular.js/blob/master/CHANGELOG.md)
-   jquery (drops [support for IE8](https://jquery.com/browser-support/))
-   [protractor](https://github.com/angular/protractor/blob/master/CHANGELOG.md) (major version jump to 2.0.2)
-   jasmine-node (fixes #1665)
-   [sass](http://sass-lang.com/documentation/file.SASS_CHANGELOG.html)
-   [tslint](https://github.com/palantir/tslint/blob/master/CHANGELOG.md)

The travis gem was removed because it is only really used to initially setup a travis setup.

I did not find any major changes in the changelogs.

While making the protractor tests work with the new version I also re-enabled some tests that were disabled due to #583. @nidico could you please check whether that is fixed now?

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/JasmineHelpers.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/JasmineHelpers.ts
@@ -3,7 +3,7 @@
 
 import * as _ from "lodash";
 
-export var customMatchers = {
+export var customMatchers : jasmine.CustomMatcherFactories = {
     toSetEqual: function(util, customEqualityTesters) {
 
         return {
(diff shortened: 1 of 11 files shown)


nidico:
LGTM. Issue #583 seems not to be present anymore.

PR #1686, opened 2015-09-25
use catalog search  to get  user

joka:
API EXTENSION:

- add search index 'user_name' (to sort user by login name)

INTERNAL:

- refactor to speed up login/activation of users

MANUAL MIGRATION NEEDED:

Restart adhocracy, then reindex all users (@slomo):

    bin/sd_reindex  -c adhocracy etc/development.ini -p "/principals/users"

### DIFF
--- a/src/adhocracy_core/adhocracy_core/catalog/adhocracy.py
+++ b/src/adhocracy_core/adhocracy_core/catalog/adhocracy.py
@@ -53,6 +53,7 @@ class AdhocracyCatalogIndexes:
     reference = Reference()
     user_name = catalog.Field()
     private_user_email = catalog.Field()
+    private_user_activation_path = catalog.Field()
 
 
 def index_creator(resource, default) -> str:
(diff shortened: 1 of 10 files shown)


slomo:
@joka How dangerous would it be to always have that reindex step in the deployment (I am not really into manual deployment steps)

slomo:
Nevermind just saw that this is a really specialised reindex.

joka:
> @joka How dangerous would it be to always have that reindex step in the deployment (I am not really into manual deployment steps)

I tried to add an automatic migrations step, but this is not easy.

joka:
> @joka How dangerous would it be to always have that reindex step in the deployment (I am not really into manual deployment steps)

reindexing all resources is not dangerous, but may take quite a lot time. So we should do it only when needed

pallix:
@joka some tests are failing

pallix:
LGTM, except failing tests

joka:
> @joka some tests are failing

The failing acceptence test is in master and might be related 
The acceptance test are not running on my local machine for some days, so unfortunately did not noticed this.

PR #1649, opened 2015-09-14
2015 09 cc use rem units

local-girl:
Hi this is quite a large overhaul of how we use units. Generally px and em measurements have all been converted to rem units with the fallback of px for outdated browsers. Rem units are relative but only to the base font size - unline em which is relative to any container with a font size, this can make it quite hard to use.

In general I have ignored print and normalize type stylesheets and anything 5px or less.

Eventually we would like to implement horizontal rhythm to aid design. I believ enow that all spacial units run through a mixin we will be able to control this via this central function.

PR #1700, opened 2015-09-29
fix header width

xi:
after #1649, the header and the columns did not have the same width anymore. I have not yet figured out why this is exactly, but this change fixes it.


local-girl:
LGTM

PR #1704, opened 2015-09-29
Add euth frontend package

slomo:
* create new package from scaffold
* extend spd package during merge static
* change brand color to those of AEGEE


xi:
LGTM.

PR #1705, opened 2015-09-29
cleanup requirejs config

xi:
-   adhTemplates was added twice
-   jquery is not loaded by requirejs (there is a script tag directly in root.html)


nidico:
LGTM.

PR #1708, opened 2015-09-29
removed white space from comment detail

### LINKED ISSUE #1630: formatting issues with `white-space: pre-line`
We use `white-space: pre-line` in some places in order to have a very lightweight markdown alternative. However, this has some problems if there are linebreaks in the beginning or end of the container, these are also preserved. This happens for example with `.comment-content`:

    <div class="comment-content" data-ng-switch-default="">                    
        {{data.content}}                                                       
    </div>

There are linebreaks before and after `{{data.content}}` in the HMTL that will be preserved.

There are several alternatives:

-   remove the linebreaks from the HTML
-   do not use `pre-line`
-   use `<br/>` instead of `pre-line` and insert them via JavaScript
-   use markdown instead

local-girl:
Fixes #1630

I've opted for simply removing the white-space in the HTML - it's something we need to keep an eye on but as this css is not used that much, I don't forsee a situation where it wouldn't be clear what the solution is.

In other scenarios (where it's just applied to <p> tags) these would naturally not have a line-break i the HTML - I say we leave this solution for now as it's the simplest.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/CommentDetail.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/CommentDetail.html
@@ -61,9 +61,7 @@
                 <a href="" class="button form-footer-button" data-ng-click="cancel()" data-ng-if="!hideCancel">{{ "TR__CANCEL" | translate }}</a>
             </footer>
         </form>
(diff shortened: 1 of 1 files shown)


xi:
Maybe some additional warning in the documentation would be helpful, but it is not clear where to put it. Merging.

Issue #1618, opened 2015-09-01
Coherent spelling of everything (process types, names, organizations etc.)


nidico:
We have different methods of spelling process types, names, organizations etc. without much logic, e.g.:

- some german terms and names from the problem domain are translated to english (e.g. not_realizeable badge), others are used in the original form (e.g. burgerhaushalt)
- some are plural (e.g. bplaene) some are singular
- different ascii conversions or used (e.g. bplaene, burgerhaushalt)
- some are camel case, some not (e.g. *meinBerlin* in frontend vs. *meinberlin* in backend).

We should investigate the different usages, define a clear policy and migrate everything.

xi:
- singular
- umlaut to ae, ue, …
- for meinberlin, no camelCase

pallix:
For the first point of @xi, do we really want to mix german and english words in the source code?

xi:
My point was more about things like "Stadtform" or "Kiezkasse". Translating common words like "Senatsmitarbeiter" to "office_worker" seems ok to me. But I see that this distinction is not very clear.

xi:
this has been fixed in the frontend. @pallix @joka is more work needed in the backend or should this be closed?

pallix:
Backend is not up to date wrt. to this policy. For example "kiezkassen".

joka:
We cannot update the backend without migrating the database and adapt frontend code.
So I`m in favour of closing this issue. We can use it as a policy for new resources

xi:
It seems like this won't be fixed in the backend, so this issue can be closed. Still I would like to add these guidelines to our documentation.

PR #1709, opened 2015-09-29
Bring meinberlin frontend inline with changes proposed in #1618 (breaking changes to embed API!)

xi:
*implements #1618 for the frontend*

this contains the following breaking changes to the embed API:

-   "kiezkassen" context is renamed to "kiezkasse"
-   "burgerhaushalt" context is renamed to "buergerhaushalt"
-   "mein-berlin-bplaene-proposal-embed" widget is renamed to "meinberlin-bplan-proposal-embed"

The old names still work via an alias system I added to the adhEmbedProvider. However, I would like to have the following deprecation procedure:

-   update the servers
-   inform all embedders about the changes
    -   include updated embed snippets
    -   give them a deadline (~3 month?) until which they can update their embed code
-   remove the aliases some time after the deadline

We need to coordinate this with project management. @slomo, @rukala: what do you think?


slomo:
Why isn't `buergerhaushalt` and `kiezkasse` named `meinberlin-*` ?

I mean after all they are as meinberlin specific as bplan. Or is the difference in naming due to their difference in embed (bplan a single proposal while the other are context embeds).

xi:
> Why isn't buergerhaushalt and kiezkasse named meinberlin-* ?

Good point. This is kind of unrelated, but when we need to inform embedders about breaking changes we could just as well cover this issue at the same time.

I guess we never thought about naming conventions for contexts. Without much thought I would think namespacing them is a good idea.

slomo:
So since this is backwards compatible. I see no issues it being merged (apart of my missing forntend expertise).

Afterwards we can decide on name spacing and afterwards inform our customers.

xi:
I updated the unit tests

nidico:
LGTM.

PR #1710, opened 2015-09-29
Fix paths in adhocracy_frontend scaffold

slomo:
Broken paths in the euth and s1 manfigest lead wheel to their wheel not containing the build directory


xi:
LGTM.

PR #1715, opened 2015-10-02
Add policycompass elastic search sync

slomo:


src/adhocracy_pcompass/adhocracy_pcompass/resources/subscriber.py
```diff
@@ -0,0 +1,72 @@
...
+
+def update_elasticsearch_policycompass(event):
+    """Push comments of IExternalResource ressources to elastic search."""
+    comment = event.object
+    external_resource = find_interface(comment, IExternalResource)
+
+    if external_resource is None:
+        return
+
+    ressource_name = get_sheet(external_resource, IName).get()['name']
```
pallix: you can use get_sheet_field here

src/adhocracy_pcompass/adhocracy_pcompass/resources/subscriber.py
```diff
@@ -0,0 +1,72 @@
...
+    match = re.match(
+        '(?P<type>visualization|event|dataset|metric|model|indicator)'
+        '_(?P<id>[0-9]+)',
+        ressource_name)
+
+    # this is not a known policycompass external resource
+    if (match is None):
```
pallix: parenthesis not necessary

src/adhocracy_pcompass/adhocracy_pcompass/resources/subscriber.py
```diff
@@ -0,0 +1,72 @@
...
+    query = search_query._replace(interfaces=IComment, only_visible=True)
+    comment_count = catalogs.search(query).count
+
+    settings = event.registry.settings
+    es_endpoint = settings.get('adhocracy_pcompass.elasticsearch_endpoint',
+                               'http://localhost:9000')
+    es_index = settings.get('adhocracy_pcompass.elasticsearch_index',
+                            'policycompass_search')
+    r = requests.post(
+        '%s/%s/%s/%s/_update' %
```
pallix: the format method of the string class should be used instead for python3

src/adhocracy_pcompass/adhocracy_pcompass/resources/test_subscriber.py
```diff
@@ -0,0 +1,85 @@
...
+                    'refers_to': context[resoruce_name]
+                }
+            })
+        return comment
+
+    def delete_comment(self, comment, registry, context):
+        from adhocracy_core import utils, resources, sheets
+
+        metadata = utils.get_sheet(comment, sheets.metadata.IMetadata, registry)
+        metadata.set({ "hidden": True })
```
pallix: unecessary whitespaces before/after brackets

src/adhocracy_pcompass/adhocracy_pcompass/resources/test_subscriber.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def requests_post(self, status_code=200):
+        response = requests.Response
+        response.status_code = status_code
+        post = Mock(name="requests.post", return_value=response)
```
pallix: single quote should be used

src/adhocracy_pcompass/adhocracy_pcompass/resources/test_subscriber.py
```diff
@@ -0,0 +1,85 @@
...
+                                           registry,
+                                           pool_with_catalogs)
+
+    @fixture
+    def context_r2(self, registry, pool_with_catalogs):
+        return self.make_external_resource('dataset_2',
+                                           registry,
+                                           pool_with_catalogs)
+
+    def make_comment(self, resoruce_name, registry, context):
```
pallix: typo resoruce_name

src/adhocracy_pcompass/adhocracy_pcompass/resources/subscriber.py
```diff
@@ -0,0 +1,72 @@
+"""Sync external ressources with elastic search."""
+
+import re
+import requests
+import logging
+
+from substanced.util import find_service
+from pyramid.traversal import find_interface
+
+from adhocracy_core.interfaces import IResourceCreatedAndAdded, search_query
```
pallix: one line for import should be preferred (see PEP8)

pallix:
Currently code coverage is excluded from policycompass. If we have to do it seriously we should have 100% test code coverage for pcompass. I guess it's not the case now because tests don't check for all http status errors. The pcompass module needs to be added to `.coveragerc` if we do it properly.

src/adhocracy_pcompass/adhocracy_pcompass/resources/subscriber.py
xi: This is not in pep8 but still a good rule (and I think it is also in our project-internal style guide).

slomo: I am kind of confused by your comments. In my opinion it should be (and also [pep8](https://www.python.org/dev/peps/pep-0008/#imports) is ony my side  :smile:  )

```
from adhocracy_core.interfaces import IResourceCreatedAndAdded, IResourceSheetModified, search_query
```
but p8p complained about it beeing too long. So i split it into

```
from adhocracy_core.interfaces import IResourceCreatedAndAdded,  search_query
from adhocracy_core.interfaces import IResourceSheetModified
```
What are you two suggesting?

slomo:
So your policy is to test really anthing also if branches, which only contain logging? Should i just run them whether and ensure they are not throwing errors or should I even check the log output?

src/adhocracy_pcompass/adhocracy_pcompass/resources/subscriber.py
pallix: Ok PEP8 is on your side, I read it too quickly/wrongly but `python_guidelines.rst` says "one import per line". Maybe because it's easier to grep? Anyway that's what we have everywhere.

nidico: I like it as it's easier to diff.

pallix:
If you thing it's not relevant to test the branch you can exclude it with a pragma but in my opinion it's better to test it. Actually if you do TDD you would have the test for the branch before writing it.

There is `log` fixture in `testing.py` that capture logs. Maybe it could be used to check that the logger is called, I'm searching for an example in the code base...

pallix:
I think here: https://pythonhosted.org/testfixtures/logging.html

nidico:
Let's merge as we need this now and fix the test coverage later.

PR #1719, opened 2015-10-04
Upgrade setuptools, pip, buildout

nidico:
Nothing of particular interest. I was mainly annoyed by warnings from setuptools, which go away with the update.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -25,7 +25,10 @@ import tempfile
 
 from optparse import OptionParser
 
-tmpeggs = tempfile.mkdtemp()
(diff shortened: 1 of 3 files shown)


slomo:
This breaks our very hacky requirements.txt generation, so I put it on hold.

I see two options to solve this:

- fix the requirements generation again
- come up with a smarter and non-hackier solution

nidico:
For me, requirements generation works if I run bootstrap and buildout.

joka:
According to  @nidico  this is general problem and is mostly solved with running buildout twice.

PR #1722, opened 2015-10-05
update angular-translate (fixes #1720)

### LINKED ISSUE #1720: core_en.json request gets canceled
I noticed that the english fallback translations do not load. In chromium, the request shows up as "canceled". In Firefox, the request shows up as successful, but with an empty response.

@local-girl could reproduce the issue. Maybe it was caused by the angular-translate update in #1684.

xi:


local-girl:
LGTM

PR #1724, opened 2015-10-06
2015 10 cc opinion module

local-girl:
Have added a new opinion module to rate <adh-opinion>

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
@@ -271,7 +271,7 @@ Click show results
     text-decoration: none;
 
     &:hover, &:focus {
-        background: $color-button-text-hover;
+        background: $color-brand-one-normal;
         color: $color-text-inverted;
     }
 }
@@ -291,7 +291,7 @@ Click show results
 }
 
 .opinion-bar {
-    background-color: $color-button-text-hover;
+    background-color: $color-brand-one-normal;
 }
 
 .opinion-bar-no {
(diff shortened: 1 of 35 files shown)


xi:
Nice! It is also nice to see that the existing rate code could be reused for this.

I have some remarks, mostly style related

-   [ ] The text below the result bar should be in `text-size-small`
-   [ ] The JA/NEIN should be bold
-   [ ] The order should be JA/NEIN, percentage, absolute number (last to are the wrong way around)
-   [ ] There should be a gab between percentage and absolute number
-   [ ] The buttons should have a have (with the same background color as the result bar
-   [ ] The total number should not appear on small screens
-   [ ] I find it hard to understand which option I chose. I know it is encoded in the coloring of the result bar, but would be nice if we could give an additional hint. Maybe a tooltip (`title="You selected 'YES'"`)?
-  [ ] When I log out, the widget should forget about my rate. That may be a general issue with the rate widget though.
-  [ ] some translation are missing

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
```diff
@@ -248,3 +248,75 @@ Click show results
...
+    display: block;
+}
+
+.opinion-bar-container {
+    @include rem(margin-bottom, 5px);
+    background-color: $color-structure-normal;
+}
+
+.opinion-bar {
+    background-color: $color-button-text-hover;
```
xi: Why `$color-button-text-hover`? This is probably a brand color, isn't it?

local-girl:
The order should be JA/NEIN, percentage, absolute number (last to are the wrong way around) - Sara told me it should be mirrored like this, did she change her mind?

local-girl:
Hi there I've done what I can for now, added en translations only

xi:
Works great!

PR #1727, opened 2015-10-06
Fix CrossWindowMessaging uninitialized login state

nidico:
As long as Adhocracy hasn't properly initialized the user login state, we shall not send any login state through CWM.

This is fixed with this PR by introducing a `adhUser.ready` promise in addition to the already existing `adhCredentials.ready` promise.


nidico:
@MagdaN : could you please have a look at this? we need it for pcompass today.

nidico:
rebased.

nidico:
Unfortunately I cannot test whether the js unit test works (`polytester jsunit` did't fail for me for some reason) and travis has [service unavailable problems](https://travis-ci.org/liqd/adhocracy3.mercator/jobs/84066529), I suggest to merge as is now.

PR #1731, opened 2015-10-09
Add search and hash to camefrom in toplevel state

slomo:
The bug that I am trying to fix can be reproduced using this [embed link]( https://embed-meinberlin-stage.liqd.net/embed/meinberlin-stadtforum-proposal-create?pool-path=https:%2F%2Fembed-meinberlin-stage.liqd.net%2Fapi%2Fstadtentwicklung%2Fstadtforum%2F) from mein berlin stage.

If the user klicks (while not being logged in) on the any of the rate widgets (or are the called directives?) she will be redirected to the login page. After logging in the user will be redirected to the embed widget without the query parameters.

After poking around a bit I would say this is an issue for all embed widgets based wrapping directly a  directives (like the discussion in policycompass). Embed widgets that are based on context or plain are not affected, since they don't use the query parameters of a request.

I am now also copying the hash of the url, that wasn't needed for my issue but seems to me a sane thing to do.

**Disclaimer:** I know that is a change on a pretty important component, hence I want to stress that I have barely a clue what I am doing when it comes to the frontend. I setup a `burgerhaushalt` instance and that seems still to work after the change. I didn't yet think about url escaping when building the came from.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
@@ -427,8 +427,7 @@ export class Service {
             "/activate"
         ];
 
-        // FIXME: DefinitelyTyped
-        if (!(<any>_).includes(denylist, url)) {
+        if (!_.includes(denylist, url)) {
             this.cameFrom = url;
             return true;
         } else {
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
@@ -414,24 +414,9 @@ export class Service {
 
     private cameFrom : string;
 
-    public setCameFrom(path? : string) : boolean {
-        if (typeof path === "undefined") {
-            path = this.$location.path();
-
-            var search = this.$location.search();
-            if (Object.keys(search).length > 0) {
-                path += "?";
-                for (var key in search) {
-                    if (me.hasOwnProperty(key)) {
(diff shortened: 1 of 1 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
```diff
@@ -417,6 +417,21 @@ export class Service {
...
+                path += "?";
+                for (var key in search) {
+                    if (me.hasOwnProperty(prop)) {
+                        path += "&" + key + "=" + search[key];
+                    }
+                }
+            }
+
+            var hash = this.$location.hash();
+            if (hash === "") {
```
xi: `!==`

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
```diff
@@ -417,6 +417,21 @@ export class Service {
     public setCameFrom(path? : string) : boolean {
         if (typeof path === "undefined") {
             path = this.$location.path();
+
+            var search = this.$location.search();
+            if (Object.keys(search).length > 0) {
+                path += "?";
+                for (var key in search) {
+                    if (me.hasOwnProperty(prop)) {
```
xi: `if (search.hasOwnProperty(key)) {`

xi:
I think the whole commit can be shortened to `$location.url()` (see <https://docs.angularjs.org/api/ng/service/$location#url>)

Apart from that, I do not see bigger issues.

slomo:
Ah, of course I didn't see that one before.

slomo:
What about the deny list checking? Shouldn't those be matched against the path only?

xi:
> What about the deny list checking?

I guess that's not relevant atm.

Issue #1729, opened 2015-10-08
BPlaene are broken


pallix:
Probably due to #1657 are bplaene broken, no email is sent currently. In `resources/subscriber.py` the function `_is_proposal_creation_finished` checks that two versions are created (because bevore #1657 version0 was empty) before sending the email.

Now only one version is created when in batch mode [1] but when checking that only one version is created and executing the sending of the email after, then the proposal version has empty values.

I suspect, but need more investigation, this is due to the way events are generated:

We are in a batch,
empty version0 is created -> event is sent
bplan subscriber receives event
accessed version0 has empty values
later in the REST logic the version0 is updated with the relevant values

[1] but not in the test, which is why the test didn't break (here more functional tests would be helpful)

pallix:
So I debugged and my intuition is correct. This mean we don't get any meaningful value for version0 with subscribers.

I'm thinking more and more to revert https://github.com/liqd/adhocracy3.mercator/commit/e8aaac010aee9f5d91fd7f9fd4e2be6f1ca29102 to fix this and #1723

slomo:
Why isn't that discovered by the unittests?

joka:
> I'm thinking more and more to revert e8aaac0 to fix this and #1723

Thats the quickest work around

joka:
> Why isn't that discovered by the unittests?

we have no acceptance for this and the unit/functional tests pass because the don't use batch requests

joka:
> We are in a batch,
> empty version0 is created -> event is sent
> bplan subscriber receives event
> accessed version0 has empty values
> later in the REST logic the version0 is updated with the relevant values

We need a subscriber for  "FinishedCreatingResourceinBatchRequest". 
We can create such an event or throw the exiting "ResourceCreatedAndAdded" add the right time or  try to reuse existing pyramid events http://docs.pylonsproject.org/projects/pyramid/en/latest/api/events.html#pyramid.events.NewResponse.

Throw "ResourceCreatedAndAdded" event at the right time sounds like the best solution. 
We also prevent needless catalog indexing then.

pallix:
We reverted the optimization PR because of lack of time for the release to implement proper fixes.

Issue #1723, opened 2015-10-05
Comments can be edited by other users


xi:
A user can edit the comments of another user. The relevant acceptance tests fails and I also verified it manually. I guess some permission is incorrect.

pallix:
`creator` field in the metadata is `null`

pallix:
This is due to the suppression of the automatic creation of version0, and the permissions check related to version0 in `content/__init__.py`. Still investigating...

joka:
This is due a the nasty hack for bpläne  (annoymous can create but not edit proposal).  
One solution would be to make this hack only for bplän proposals and not for all resources.

joka:
To fix the exiting hack we can do:

if version_0  == last_version_created_in_transaction:
        allow create item_version

PR #1733, opened 2015-10-12
Revert first version optimization

pallix:


xi:
The jsunit test "CrossWindowMessaging Service sendAuthMessages returns true if embedderOrigin is trusted" also fails on master (probably caused by #1727) so you can ignore it for now.

pallix:
Temporary fixes #1729 and #1723

Issue #1711, opened 2015-09-29
remove ResourceWidget


xi:
The resourceWidget concept was a failed attempt to get more structure into the frontend code. It is overly complex and has only been used for comments and mercator proposals. We should refactor this to use the newer `bind_path`/`post_create`/`post_edit` concept and remove the  resourceWidget module.

xi:
I started refactoring comment in 2015-10-tb-refactor-comment

der-john:
I've considered it and decided the cost was too high. @xi and I think the `resourceWidget` module should be moved to `mercator`, though.

der-john:
In my opinion, this is not worth doing unless AE2017 uses a3.

der-john:
That's as far as we should go.

PR #1743, opened 2015-10-13
Do not use resourceWidget for comments

xi:
In the spirit of #1711, this refactors the comment code to not use resource widgets. Sorry for the messy commits.


nidico:
LGTM and everything still seems to works fine.

PR #1744, opened 2015-10-13
Fix stadtforum workflow

pallix:
On hold since we don't know the semantic we want for stadtforum.

Multiple processes, one process with multiple proposals having each one workflow?


joka:
Is this still relevant?

pallix:
nö.

PR #1745, opened 2015-10-13
Upgrade py, beautifulsoup4 and pytest

pallix:
Upgraded versions are python3.5 compatible.

### DIFF
--- a/src/adhocracy_core/versions.cfg
+++ b/src/adhocracy_core/versions.cfg
@@ -36,7 +36,7 @@ pip = 6.0.8
 plone.recipe.command = 1.1
 profilehooks = 1.7.1
 pudb = 2015.2
-py = 1.4.26
+py = 1.4.30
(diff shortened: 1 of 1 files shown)


nidico:
Still works on python 3.4, merging.

src/adhocracy_core/versions.cfg
```diff
@@ -181,7 +181,7 @@ pyramid-tm = 0.11
 # Required by:
 # pytest-pyramid==0.1.1
 # pytest-splinter==1.2.9
-pytest = 2.6.4
+pytest = 2.7.3
```
joka: why don't you update to pytest 2.8?

pallix: true, it should be 2.8

nidico: There's 2.8.2 now.

PR #1749, opened 2015-10-14
Add UI for creating bplan processes

xi:
-   office workers need to be entered by full URL
-   on success, an embed snippet is shown. This makes it very useful in a specific usecase but not useable in some others.
-   I did end up not using `date` inputs because only chrome seems to support them and the polyfill is kinda broken. This was an issue before, can't find the issue right now.


nidico:
Works nicely. Some remarks:

- It can be used as the following:

http://localhost:6551/embed/meinberlin-bplan-process-create?pool-path=http:%2F%2Flocalhost:6541%2Fadhocracy%2F

- It doesn't set the workflow state to `participate`. #1760 introduces a generic workflow state switcher, which can be used for that.

- For some reason, the embed code fails for me locally due to the trailing slash in the created bplan resource. I've created #1761 for this.

Issue #1732, opened 2015-10-09
Standard workflow might not be suited for Stadtforum


slomo:
The standard workflow allows all users to create proposals (== questions) in Stadtforum. That might not be what is intended with that process.

pallix:
This was fixed.

Issue #1347, opened 2015-06-22
rate button not disbaled for anonymous user 


MagdaN:
when i am not logged in the rate button is not disabled in result phase

xi:
The desired situation is something like this:

-   If the user is not logged in, we show functionality available to "standard" users. If users try to use the functionality, they are redirected to login.
-   If the user is logged in, they see the actual availability of the functionality.
-   If a feature is not available, it is either marked as disabled or is hidden completely.

The implementation of this is not very consistent and not very complete. Often, a feature is visible by default. In #1164 I implemented a method to disable features if a user is logged in but does not have the required permission. This does not cover the case where a user is not logged in and a "standard" user would not have the permission.

For this, we need an extension of that backend API that allows us to get permissions of a "standard" user.

I talked to @rukala about this and we decided to not fix this now and do a proper fix in one of the upcoming sprints.

MagdaN:
Should we not at least hide it than also based on the phase?

xi:
Today I thought a bit more about this and collected some issues connected to this:

1.   We use `cameFrom` to get back after redirecting to login. So #1137 and #1061 apply.
2.   If the functionality that is shown to the anonymous user has state (e.g. `textarea`) that state is lost after redirecting to login.
3.  The frontend does not know the permissions of a "standard" user.
   -  The current workaround for this is to hardcode / guess the permissions. This approach works reasonably well for special cases but is not applicable for reusable features (e.g. comments).
4.   After successful login, should the requested action be automatically executed or should I just be redirected to the same place?
5.   Maybe related: If functionality is visible in the frontend even though there is no permission, this will result in backend errors. Unfortunately, there is no concept for showing useful messages on backend errors. Even the error parsing is sometimes broken (e.g. #748)

(1) is a frontend bug tracked in separate issues. (2) can be mitigated by replacing the actual functionality with a button for anonymous users. (3) can be fixed by extending the API as described above. (4) is an open question and (5) is IMHO only related to specific workarounds.

So I guess if we use this issue to track (3) and the respective issues to track (1) this can all be fixed.

pallix:
Can we remove backend tag on this?

xi:
> Can we remove backend tag on this?

No. In order for this to be fixed the API needs to be extended (see (3) from my previous comment).

joka:
Just to clear some responsibilities regarding the backend, it:

- cannot say what the user can do until the user is logged in (we don't know about local roles, groups)

[...]

PR #1750, opened 2015-10-14
show fake comment form for not logged-in users in stadtforum

xi:
some notes:

-   [ ] translation is missing
-   [ ] the disabled textarea is not visually distinguishable from a non-disabled one
-   [ ] the disabled textarea is resizable
-   This is implemented in stadtforum instead of comment because the permission check is hardcoded (it is not possible to check "what would a 'normal' user be able to do"; see #1347). If the scope is limited to stadtforum, it is reasonable that we can guess the correct permissions.
-   As there is no proper workflow for stadtforum yet (see #1732) there are no permission checks in place. Once we have a proper workflow we might want to show/hide the fake form based on workflow state.


local-girl:
LGTM

local-girl:
Other than missing features above. Have made a separate pull request for disabled textareas in #1752

PR #1752, opened 2015-10-14
2015 10 cc disabled textareas

local-girl:
as requested by #1750. Stops textareas from resizing when disabled. They still look the same however, to discuss with designers as this is an input-wide concern.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_forms.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_forms.scss
@@ -30,6 +30,21 @@ it is also dirty or the form has been submitted.
 
 /*doc
 ---
+title: textareas
+name: input.textarea
+category: Inputs
+---
+
+Textareas behave exactly like text inputs however they are resizeable when enabled in modern browsers.
+
+```html_example
+<textarea></textarea>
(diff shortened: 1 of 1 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_forms.scss
```diff
@@ -59,6 +59,21 @@ They should all look and behave similarily.
...
+parent: textinput
+---
+
+Textareas behave exactly like text inputs however they are resizeable when enabled in modern browsers.
+
+```html_example
+<textarea></textarea>
+<textarea disabled="disabled" placeholder="disabled"></textarea>
+```
+*/
```
xi: There is already a textarea in the textinput documentation. That should probably be removed.

xi:
There is still the issue that disabled textareas are not visually distinguishable from non-disabled ones.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_forms.scss
local-girl: But that's true of all inputs, the designers need to think about this and it should be a separate pull request I think

local-girl:
But that's true of all inputs, the designers need to think about this and it should be a separate pull request I think

PR #1754, opened 2015-10-14
Make mercator proposals items badgeable

pallix:


pallix:
A migration is probably needed

joka:
> A migration is probably needed

indeed,

Note: its part of the refactoring list to get rid of this kind of migrations

PR #1755, opened 2015-10-14
Show meaningful message on external resource comment list if logged out

nidico:
Fixes policycompass/policycompass#294.


xi:
a bit hacky, but seems to work.

nidico:
Indeed.

Issue #1757, opened 2015-10-14
mercator migrations broken


MagdaN:
the migrations that add badgeables to items does not work:

```
Traceback (most recent call last):
  File "bin/sd_evolve", line 117, in <module>
    sys.exit(substanced.scripts.evolve.main())
  File
"/home/magda/Documents/dev/adhocracy3/src/substanced/substanced/scripts/
evolve.py",
line 83, in main
    complete = manager.evolve(latest)
  File
"/home/magda/Documents/dev/adhocracy3/src/substanced/substanced/evolutio
n/__init__.py",
line 75, in evolve
    func(self.context)
  File
"/home/magda/Documents/dev/adhocracy3/src/adhocracy_core/adhocracy_core/
evolution/__init__.py",
line 148, in logger_decorator
    func(*args, **kwargs)
  File
"/home/magda/Documents/dev/adhocracy3/src/adhocracy_mercator/adhocracy_m
ercator/evolution/__init__.py",
line 46, in evolve1_add_ititle_sheet_to_proposals
    if 'title' not in introduction._sheets[IIntroduction.__identifier__]
:
AttributeError: 'Base' object has no attribute '_sheets'
```

PR #1759, opened 2015-10-14
update categories

MagdaN:
I updated the categories but could not get the Blog code running:

1. the migration script that adds badgebales to all proposals did not work #1757


xi:
LGTM

PR #1765, opened 2015-10-15
2015 10 mh fix build on mac against master

dwt:
Fix build for mac.
    
The problem is that this script uses gnu coreutil specific switches - so coreutils needs to be installed on the mac and prepended to the path for this script. Also added an error message to make it easier to debug why things went wrong.


nidico:
LGTM.

PR #1770, opened 2015-10-15
2015 10 cc new report ui

local-girl:
Added new abuse overlay as per Nadines requests. There is a new flag icon and color. Also the overlay now has help text with an optional link to nettiquitte url (only used in Mercator for now)


MagdaN:
I get a sass errror

local-girl:
I think this is in master

local-girl:
I fixed it in this branch by adding the @optional like it says but I never discovered what the underlying problem was

local-girl:
This seems to have been fixed upstream

MagdaN:
works

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Column.html
```diff
@@ -1,8 +1,14 @@
 <div data-ng-switch="transclusionId">
     <div data-ng-switch-when="menu">
-        <div class="moving-column-tab">
-            <i class="icon-speechbubbles moving-column-icon"></i>
-            {{ "TR__COMMENTS" | translate }}
+        <div class="moving-column-tab" data-ng-switch on="overlay">
+            <span ng-switch-when="abuse">
```
xi: `data-` prefix missing

PR #1772, opened 2015-10-16
2015 10 nd s1: user activity summary widget

dwt:
Seems I can't yet push to liqd directly - needs debugging. Thus the pull request with how far I came.

Still to do: Styling and localization


dwt:
Anything else you need to merge this?

nidico:
Is there any reason to base this on S-1 instead of on master apart from the fact that it's not ready yet? I'd rather have this ready and merged into master directly otherwise, so we don't enlarge the s-1 branch even more and have separate features already available on master.

dwt:
Sure, let’s do that -> Thursday.

Regards,
Martin

PR #1775, opened 2015-10-19
updated ZODB && dependencies (zodbpickle has fixed python 3.4 support!)

joka:


pallix:
one failing pyunit test is trying to hide behind the mess of outputs

joka:
> one failing pyunit test is trying to hide behind the mess of outputs

The failing python unit tests are in master

PR #1778, opened 2015-10-20
Skip header validation if authentication is done by Thentos

fhartwig:
I've added a check for `adhocracy.validate_user_token` that's necessary to use A3 with Thentos and that #1511 seems to have missed.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/rest/test_batchview.py
+++ b/src/adhocracy_core/adhocracy_core/rest/test_batchview.py
@@ -46,6 +46,7 @@ class TestBatchView:
     @fixture
     def request_(self, request_, changelog, mock_invoke_subrequest):
         request_.registry.changelog = changelog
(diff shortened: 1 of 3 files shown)


joka:
LGTM

ChristianSi:
I think this PR just hides a different problem. Thentos sends a 'X-User-Path' header that should point to an existing user, so if that header is present, `get_user(request)` should NOT be None. @fhartwig is on it.

joka:
> I think this PR just hides a different problem. Thentos sends a 'X-User-Path' header that should point to an existing user, so if that header is present, get_user(request) should NOT be None. @fhartwig is on it.

If you spend time refactor/fix this mind the "proper" pyramid way to do this:
https://github.com/liqd/adhocracy3.mercator/pull/1511#issuecomment-134127195

This would  also fix the "get_user(request)" should not be none problem

PR #1779, opened 2015-10-20
Fixed close button

### LINKED ISSUE #1777: close button on comment column in mercator not working

local-girl:
Fixes #1777

I'm not really sure why this works but it does - possibly breaks something else? Also tested in meinberlin, spd to make sure it doesnt break other sites.


MagdaN:
tried it and works with mercator

xi:
I don't think this is a proper fix because it will break closing the column everywhere else. The proper fix would be to not use `adhParentPath` when `commentCloseUrl` is set in the routing.

MagdaN:
@xi : could you check this please. We were not sure how to do it properly.

PR #1786, opened 2015-10-22
Add frontend and backend modules for the redesign

pallix:
Caroline asked that to help her beginning working on the redesign.


pallix:
maybe we don't need the backend module at all?

PR #1787, opened 2015-10-23
make header optional

MagdaN:
very simple way to make the header for spd optional. I did this because I could not figure out how to get the parameters from the embed snippet.


local-girl:
LGTM

PR #1790, opened 2015-10-26
Fix mercator migrations

pallix:


src/adhocracy_core/adhocracy_core/evolution/__init__.py
```diff
@@ -101,6 +101,7 @@ def migrate_new_sheet(context: IPool,
         if remove_isheet_old:
             logger.info('Remove {0} sheet'.format(isheet_old))
             noLongerProvides(resource, isheet_old)
+        catalogs.reindex_index(resource, 'interfaces')
```
joka: are the tests still running?

pallix: yes

PR #1795, opened 2015-10-28
fix policy compass  subscriber to reindex external resources

joka:


src/adhocracy_pcompass/adhocracy_pcompass/resources/subscriber.py
```diff
@@ -32,31 +35,27 @@ def notify_policycompass(event):
 
     # this is not a known policycompass external resource
     if match is None:  # pragma: no cover
-        return
+        return ''
```
pallix: I would rather through an ValueError exception here no? Because it's used in `requests.post` after, what would be the effect if the argument is an empty string?

joka: Yes, I just did not want tp change to old behavior

joka:
Did rebase to cleanup History

PR #1798, opened 2015-10-29
Parse URLs instead of doing string slicing that breaks in the presence of small variations in the URL

### LINKED ISSUE #1797: Deserialisation of urls is broken when using virtual hosting
https://github.com/liqd/adhocracy3.mercator/blob/21d7e4649ee486f6eed72738f189f074ebef5e5a/src/adhocracy_core/adhocracy_core/schema/__init__.py#L474-L478

The `Fixme` is biting me right now, the path gets completely mangled, so `find_resource` fails. I think this can be fixed by just parsing the url properly (i.e. using `urllib.parse`), but Matthias suggested I ask if this is the right approach. I'll send a PR later unless someone tells me this is not the right way to go.

fhartwig:
Fixes #1797 
With this patch, A3 seems to work when using virtual hosting. I hope I didn't miss any problems with this. Tests still pass. I might add a regression test once I figure out how.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/schema/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/schema/__init__.py
@@ -6,6 +6,7 @@ import decimal
 import io
 import os
 import re
+import urllib
 
 from pyramid.path import DottedNameResolver
 from pyramid.traversal import find_resource
@@ -474,8 +475,7 @@ class ResourceObject(colander.SchemaType):
             application_url_len = len(request.application_url)
             if application_url_len > len(str(value)):
                 raise KeyError
-            # Fixme: This does not work with :term:`virtual hosting`
-            path = value[application_url_len:]
+            path = urllib.parse.urlparse(value).path
             return find_resource(request.root, path)
 
 
--- a/src/adhocracy_core/adhocracy_core/schema/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/schema/__init__.py
@@ -475,6 +475,7 @@ class ResourceObject(colander.SchemaType):
             application_url_len = len(request.application_url)
             if application_url_len > len(str(value)):
                 raise KeyError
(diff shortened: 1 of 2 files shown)


fhartwig:
I've pushed another commit that fixes a similar problem in a different place. In the first commit I've deleted a `FIXME` because I assume that this is fixed by my change, please let me know if you disagree.

joka:
> With this patch, A3 seems to work when using virtual hosting.

What exactly is your virtual hosting use case? 
Do you want to hide or add some parts of the resource paths?

fhartwig:
Sorry, the "virtual hosting" stuff was me being confused. The problem that this solves is that Thentos and A3 can have slightly different (but equivalent) ideas what the backend's URL is. E.g. Thentos thinks the backend lives at `https://example.com:443` and A3 thinks the backend is at `https://example.com`. If this happens, the code that extracts the path from an URL mangles the path because its its assumptions about the length of the first part of the domain are invalid.

src/adhocracy_core/adhocracy_core/schema/__init__.py
```diff
@@ -474,8 +475,7 @@ def _deserialize_location_or_url(self, node, value):
             application_url_len = len(request.application_url)
             if application_url_len > len(str(value)):
                 raise KeyError
-            # Fixme: This does not work with :term:`virtual hosting`
-            path = value[application_url_len:]
+            path = urllib.parse.urlparse(value).path
             return find_resource(request.root, path)
```
joka: The Fixme is not solved,  virtual hosting is still broken: http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/vhosting.html

fhartwig: Sorry about that. I just put it back.

joka:
lgtm

PR #1802, opened 2015-10-29
user activity overview

dwt:
This is foremost for s1 - but as it is general enough to go into core proper it's built in core.

It can be used in the UserDetailColumn.html like this:

```patch
@@ -17,6 +17,12 @@
     <adh-recompile-on-change
         data-ng-switch-when="body"
         data-value="{{userUrl}}">
+        <adh-user-activity-overview
+         data-show-comments="true"
+         data-show-proposals="true"
+         data-show-ratings="true"
+        >
+        </adh-user-activity-overview>
         <adh-user-profile
             data-path="{{userUrl}}">
             <section class="s1-user-profile-proposal-list">
```

Does this work for you or do you require some further changes?

Many thanks,
Martin

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserActivityOverview.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserActivityOverview.html
@@ -1,9 +1,9 @@
 <table class="user-activity-overview">
   <tbody>
     <tr>
-      <td class="comments">{{ commentCount }} Beiträge</td>
-      <td class="proposals">{{ proposalCount }} Eigene Vorschläge</td>
-      <td class="ratings">{{ rateCount }} Anhänger</td>
+      <td class="comments" data-ng-if="commentCount">{{ commentCount }} Beiträge</td>
+      <td class="proposals" data-ng-if="proposalCount">{{ proposalCount }} Eigene Vorschläge</td>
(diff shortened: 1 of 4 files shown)


nidico:
Nice and works.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserActivityOverview.html
```diff
@@ -0,0 +1,7 @@
+<div class="user-activity-overview">
+  <div class="wrapper">
+    <div class="comments" data-ng-if="commentCount">{{ commentCount }} {{ "TR__COMMENTS" | translate}}</div>
+    <div class="proposals" data-ng-if="proposalCount">{{ proposalCount }} {{ "TR__PROPOSALS" | translate }}</div>
+    <div class="ratings" data-ng-if="rateCount">{{ rateCount }} {{ "TR__RATES" | translate}}</div>
```
xi: These classes should be prefixed with widget name (see [CSS guidelines](https://adhocracy3mercator.readthedocs.org/en/latest/coding_guides/css_guidelines.html#selectors))

dwt: will do.

dwt: fixed in c09bbc2d57abebd4bd37447f7f8c708f3408379a

PR #1805, opened 2015-10-30
add picture reference to user resource

joka:
API Extensions:

* add IHasAssetService sheet to users service (/principals/users)
* add IImageReference sheet to all user resources

This is needed by S1 process and mercator.


pallix:
some pyfunc tests are failing

pallix:
actually they pass, I had a problem in my setup. LGTM.

PR #1807, opened 2015-11-02
move functionality from column to directive

MagdaN:
This could be the first step to move the functionalities from the column to the page (which is not yet implemented)


local-girl:
Hi there, I stupidly stated working on the same thing - I'm now in the process of merging your branch with mine 2015-11-cc-remove-cancel-from-column

local-girl:
Ideally I would like to make the resource actions bar also refactor a lot of the code. EG Cancel functions are repeated all over the place and mostly do exactly the same thing, I think this should be a directive.

same goes for editing, creating. reporting and all of that - i think they should be directives all part of this new module. I'm going a bit over my own head here but so far testing in MeinBerlin seems to work - I will need help though !

xi:
Replaced by #1821.

PR #1808, opened 2015-11-03
Tweak export users script

fisx:
See commit messages.

If you need a command line flag before you can merge this, please let me know.

Thanks @joka for talking me through this!


pallix:
Flag would be nice, have a look at `set_workflow_state.py` for example of how to parse flags

fisx:
@pallix fixed (i think).  thanks for the review!  (-:

PR #1810, opened 2015-11-04
2015 11 cc map switch to list

local-girl:
Please note this is an interim solution while and needs to be refactored after the redesign. Also it will clash with 5fc7910bfd6c51b776c979995200cb69d8feb190 but I'll fix this when it comes.

Firstly I don't like facets.scss - it needs to become filtering.scss
I'm not sure if listing-action-bar is the right name

This is basically something to look nice before we really implement the redesign, all the javascript should remain the same or similar, the styling is temporary for now.


xi:
Continued in #1824

PR #1811, opened 2015-11-04
Create styles for listing action bar

local-girl:
I need these styles over and over again today so I will merge to master, this will be refactored through the redesign very soon so it's not that important and it's new styles so wont damage anything. Will probably conflict with 5fc7910 and is duplicated by 9714fa58d27999bcf72d6b42366c0a6c259178df


xi:
Could you please add an example in the comment?

local-girl:
This is now part of #1807 where the scss has been moved from facets to action_bar. Please continue here.

PR #1812, opened 2015-11-05
fix statsd metric authenticationuser

joka:


joka:
merged in other pull request

PR #1814, opened 2015-11-07
Fix: manual upgrade step need for new index user_email/private_user_name

joka:
This updated migration step ist testet with the mercator productive database 
Migration run fine.

But I had internal errors when: 

- creating commnets 
- vote for existing proposal. 

Also the autoname generation for new use was not working.
More testing/Fixes neeed before we update productive.


pallix:
What is the cause of the errors?

PR #1816, opened 2015-11-09
[RFC] Login indicator can switch between `built-in` and `thentos`.

fisx:
[I'm opening this so I can refer to it in discussions, which I can still do if you close it immediately.  Sorry for the noise!]

These few changes demonstrate how a login widget can be embedded from thentos-proxy.  The embedding is "shallow" in the sense that it shares the DOM with the surrounding angular event loop.  Since neither of ng, thentos know (or should care) about the other's DOM element, my hypothesis is that this is a viable approach.

You need to run adhocracy in thentos-proxy mode for this to work (poorly documented).

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Indicator.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Indicator.html
@@ -1,4 +1,10 @@
 <div class="user-indicator" data-ng-switch="credentials.loggedIn">
+
+
+    <div id="thentos-indicator"></div>  <!-- thentos: render here! -->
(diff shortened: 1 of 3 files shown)


fisx:
(As an interesting aside, the thentos widget is loaded many times instead of just once.  Before I built in a test that makes all but the first calls do nothing, I saw many copies of the same widget popping up.  That looks to me like a bug either in angular or in the a3 code that uses it.  Could be performance relevant.)

xi:
@fisx and me had a meeting about this last week and we discussed some options. We will most likely stick with the approach presented in this PR (thentos provides a 3rd party library; thentos devs maintain thin wrapper directives around that library in adhocracy code).

fisx:
closing this; I will re-open something hopefully more useful soon!

PR #1818, opened 2015-11-09
Fixups to "new report ui"

xi:
This are some fixups to #1770. It should fix some iconfont weirdness.

I also fixed a typo (nettiquette -> netiquette). This has a breaking change in the config: `adhocracy.frontend.netiquette_url`.


xi:
As this fixes the CSS builds for most platforms, I added the "high prio" tag.

pallix:
@joka configuration needs to be updated

PR #1819, opened 2015-11-09
Fix close button

### LINKED ISSUE #1777: close button on comment column in mercator not working

xi:
This fixes #1777.

The actual issue was caused in #1743. There, `CommentColumn.html` was renamed to `Column.html`. However, I forgot to also rename the overwrites in mercator and spd.

The mercator overwrite implements exactly the same solution as #1779. However, this version is specific to mercator. Note that mercator is the only platform that has URLs based on versions rather than items.


nidico:
LGTM!

PR #1825, opened 2015-11-13
Fix: version.py only adds annotated tags

joka:
Its confusing that you have to add an annotated tag in order to make a new release version.
This pull request changes version.py to also work with non annonated tags.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/scaffolds/adhocracy/version.py
+++ b/src/adhocracy_core/adhocracy_core/scaffolds/adhocracy/version.py
@@ -37,7 +37,7 @@ from subprocess import Popen, PIPE
 def call_git_describe(abbrev=4):
     """Call git describe to get the current version number."""
     try:
-        p = Popen(['git', 'describe', '--abbrev=%d' % abbrev],
+        p = Popen(['git', 'describe', '--tags', '--abbrev=%d' % abbrev],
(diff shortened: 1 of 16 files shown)


nidico:
So what does this PR change exactly? That I don't have to use annotated tags anymore?

joka:
> So what does this PR change exactly? That I don't have to use annotated tags anymore?

yes

xi:
IIRC @slomo wanted to enforce annotated tags because some tools ignore non-annotated tags (can't remember which tools though)

slomo:
It is exactly that tool which ignores not annotated tags. But what I really want to archive is the usage of signed tags (also some release note like annotations also doesn't hurt).

joka:
> It is exactly that tool which ignores not annotated tags. But what I really want to archive is the usage of signed tags (also some release note like annotations also doesn't hurt).

Release notes would be very helpfull  but this should be part of the code, like the CHANGELOG files.

If we really need annotated tags the build script should enforce this.
Else we should merge this pull request to easy the release process.

joka:
@slomo may i write an email to you if I need information?

slomo:
Well just merge this. There is nothing else than those scripts actually dealing with the git tags.

joka:
> Well just merge this. There is nothing else than those scripts actually dealing with the git tags.

How can we contact you?

PR #1829, opened 2015-11-18
Fix sources for pep8

pallix:


xi:
The `__init__.py` files are generated from `src/adhocracy_frontend/adhocracy_frontend/scaffolds/adhocracy/+package+/__init__.py_tmpl`. So the issue should be fixed there, too.

The `version.py` is third party code and should probably be excluded from linting.

pallix:
Should we exclude `version.py` from the linter or it is sometimes modified manually?

joka:
> Should we exclude version.py from the linter or it is sometimes modified manually?

version.py passed the pep8 test, so why should we exclude it?

xi:
> > Should we exclude version.py from the linter or it is sometimes modified manually?
>
> version.py passed the pep8 test, so why should we exclude it?

It did *not* pass the pep8 test (this is what this PR is about) and it is third party code. We should not modify that code so there is no need to lint it.

joka:
> It did not pass the pep8 test (this is what this PR is about) and it is third party code. 
anyway, if its easy fix we should do it, then we don't depend on special configuration to make the pep8 test pass.

joka:
lgtm

joka:
> Should we exclude version.py from the linter or it is sometimes modified manually?

It was modified twice until  now.

PR #1821, opened 2015-11-09
2015 11 mn move column functions to directive+1

local-girl:
Please note this is work in progress. It basically works in Mercator, I will add functionality for other sites as I come across it rather than pre-empting functionality.

Here are places we need to put the action bar
# Mercator
- [x] proposal detail column
- [ ] proposal list column (filter and sort) Functionality not available in this version of action bar
- [x] new proposal (cancel only)
- [x] new proposal (cancel only)

# SPD
- [ ] list documents (create document)
- [ ] Document detail (edit document)

# MeinBerlin
## kiezkasse
- [ ] List documents (mapSwitcher)
- [ ] Proposal detail (edit)
- [ ] Proposal edit (cancel)
- [ ] MISSING FROM CURRENT? Proposal create (cancel)

## Alex
- [ ] List documents (mapSwitcher)
- [ ] List documents / document tab (create document)
- [ ] Create Proposal (cancel)
- [ ] Proposal Detail (edit)
- [ ] Proposal Edit (cancel)
- [ ] Create Document (cancel)
- [ ] Document Detail (edit, cancel)
- [ ] Document Edit (cancel)

## BPlan
?? Can't run it ?

## Stadtforum
?? Can't run it ?

## Buergerhaushalt
- [ ] List Proposals (Map switch)
- [ ] MISSING Create Proposal (cancel)
- [ ] Proposal Detail (edit)
- [ ] Proposal Edit (cancel)

# Policy Compass
- [ ] Process Detail (create)
- [ ] Proposal Detail (edit, create badge assignments (does not exist in actionbar!!))
- [ ] Proposal Edit (cancel)

PR #1832, opened 2015-11-19
2015 11 cc add filters to actionbar

local-girl:
Added filter and sort to new actionbar code


local-girl:
Will continue working in this branch

local-girl:
See #1821

local-girl:
This should be in the listing package instead

PR #1834, opened 2015-11-19
2015 11 cc add mapswitcher to actionbar

local-girl:
Not complete yet, just pushing for help
See it in Meinberlin Kiezkassen proposal list


local-girl:
ABORT (discussed with @MagdaN )

PR #1835, opened 2015-11-19
Add aula packages

pallix:


src/adhocracy_aula/adhocracy_aula/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def includeme(config):
+    """Setup adhocracy extension."""
+    # include adhocracy_core
+    config.include('adhocracy_core')
+    # commit to allow overriding pyramid config
+    config.commit()
+    # add translations
+    config.add_translation_dirs('adhocracy_core:locale/')
```
joka: We also have to add the translations from the core package:
   
    config.add_translation_dirs('adhocracy_core:locale/',
                                              'adhocracy_meinberlin:locale/')

src/adhocracy_aula/setup.cfg
```diff
@@ -0,0 +1,20 @@
...
+add_comments = TRANSLATORS:
+output_file = adhocracy_aula/locale/adhocracy_aula.pot
+width = 80
+
+[init_catalog]
+domain = adhocracy_aula
+input_file = adhocracy_aula/locale/adhocracy_aula.pot
+output_dir = adhocracy_aula/locale
+
+[update_catalog]
```
joka: We use the script bin/i18n to manage translation file and  have no setuptool integration based on babel:
https://github.com/liqd/adhocracy3/blob/master/docs/development.rst#update-translations-backend

So setup.cfg ist not used at all and we can remove it from the scaffold files, or did I miss something?

joka:
LGTM, the comment is a general remark and not related to this pull request

PR #1839, opened 2015-11-20
show results in resultsphase also to anonymous user

MagdaN:
does not work yet (at least with my test data).


MagdaN:
I tried to get this working, but apparently  adhHttp.get(processUrl) is undefined. It seems to me that currentSpace in adhToplevelstate is not set.

xi:
Should work now.

MagdaN:
Works!

xi:
> if users are not logged in, we don't know which part is supposed to be red and which one is grey.

fixed

PR #1841, opened 2015-11-23
rm hint about adhocracy_sample in docs

xi:
not required anymore since the integration tests have been removed.


pallix:
I think it was merged from and old branch, the numerations of the items is wrong in master

PR #1842, opened 2015-11-23
2015 11 cc add mapswitcher filters to actionbar

local-girl:
Design wise listing actions must be split from resource actions, this is the first step of this.

Couldn't figure out how to make the map switcher work without removing the scope (again) can someone please fix this @xi :D thanks

This is spread across meinberlin (map switch) and mercator (filters) will ad to all sites once we get this merged.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
```diff
@@ -294,6 +288,60 @@ export class Listing<Container extends ResourcesBase.Resource> {
...
+                scope.data.showFilter = !scope.showFilter;
+            };
+
+            scope.toggleSort = () => {
+                scope.data.showFilter = false;
+                scope.data.showSort = !scope.showSort;
+            };
+        }
+    };
+};
```
xi: The actual filter and sort options should be included in this directive (currently it is only the buttons that toggle showing them)

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
```diff
@@ -294,6 +288,60 @@ export class Listing<Container extends ResourcesBase.Resource> {
...
+            filterAndSort: "=?",
+            data: "=?"
+        },
+        link: (scope) => {
+            scope.showMap = (isShowMap) => {
+                scope.model = isShowMap;
+            };
+        }
+    };
+};
```
xi: I wonder whether it would be better to just include this functionality in the listing directly. It is not very likely that we will want to reuse it somewhere else.

local-girl:
Cherry-picked to #1845

PR #1846, opened 2015-11-24
add new proposal

MagdaN:


xi:
Should be fine as a base for further development.

PR #1847, opened 2015-11-24
Add resources, sheets and workflow for mercator2

pallix:
Modelling for mercator 2 so far. I assumed that "road to impact" and "selection criteria" are each commented in block. This may still change as the requirements for that were not clear.

A JSON resource file is available in `src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator2/resources.json`.

### DIFF
--- a/src/adhocracy_mercator/adhocracy_mercator/resources/mercator2.py
+++ b/src/adhocracy_mercator/adhocracy_mercator/resources/mercator2.py
@@ -1,0 +1,155 @@
+"""Mercator 2 proposal."""
+
+from adhocracy_core.resources import add_resource_type_to_registry
+from adhocracy_core.resources import process
+from adhocracy_core.resources import proposal
+from adhocracy_core.interfaces import ISimple
+from adhocracy_core.resources.logbook import add_logbook_service
+from adhocracy_core.resources.proposal import IProposal
+from adhocracy_core.resources.simple import simple_meta
+from adhocracy_core.sheets.comment import ICommentable
+from adhocracy_core.sheets.description import IDescription
+from adhocracy_core.sheets.rate import IRateable
+from adhocracy_core.sheets.title import ITitle
+from adhocracy_core.sheets.image import IImageReference
+from adhocracy_core.resources.comment import add_commentsservice
+from adhocracy_core.resources.rate import add_ratesservice
+from adhocracy_core.resources.badge import add_badge_assignments_service
+
+import adhocracy_mercator.sheets.mercator2
+import adhocracy_core.sheets
+
+
+class IPitch(ISimple):
+    """Proposal's pitch."""
+
+
+pitch_meta = simple_meta._replace(
+    content_name='Pitch',
+    iresource=IPitch,
+    permission_create='create_proposal',
+    use_autonaming=True,
+    autonaming_prefix='pitch',
+    extended_sheets=(
+        adhocracy_mercator.sheets.mercator2.IPitch,
+        adhocracy_core.sheets.description.IDescription,
+        adhocracy_core.sheets.comment.ICommentable),
+)
+
+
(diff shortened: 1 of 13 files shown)


src/adhocracy_mercator/adhocracy_mercator/resources/mercator2.py
```diff
@@ -0,0 +1,155 @@
+"""Mercator 2 proposal."""
+
```
joka: "Mercator 2" is not very descriptive. We should explain what is the difference to "mercator", like  "Proposal forms for the  2016 participation process."

xi: FYI: In the frontend we moved all the old code to a 2015 namespace and added a 2016 namespace for all new code.

src/adhocracy_mercator/adhocracy_mercator/sheets/test_mercator2.py
```diff
@@ -0,0 +1,778 @@
...
+        return testing.DummyResource(__provides__=IItem)
+
+    def test_create_valid(self, meta, context):
+        from zope.interface.verify import verifyObject
+        from adhocracy_core.interfaces import IResourceSheet
+        from adhocracy_mercator.sheets.mercator2 import IOrganizationInfo
+        from adhocracy_mercator.sheets.mercator2 import OrganizationInfoSchema
+        inst = meta.sheet_class(meta, context)
+        assert IResourceSheet.providedBy(inst)
+        assert verifyObject(IResourceSheet, inst)
```
joka: Test to verify  the IResouceSheet  interface is not necessary if we use the default sheet class (adhocracy_core.sheets.AnnotationSheet)

src/adhocracy_mercator/adhocracy_mercator/sheets/test_mercator2.py
```diff
@@ -0,0 +1,778 @@
...
+        assert inst.meta.isheet == IWinnerInfo
+        assert inst.meta.schema_class == WinnerInfoSchema
+        assert inst.meta.permission_view == 'view_mercator2_winnerinfo'
+        assert inst.meta.permission_edit == 'edit_mercator2_winnerinfo'
+
+    @mark.usefixtures('integration')
+    def test_includeme(self, meta):
+        from adhocracy_core.utils import get_sheet
+        context = testing.DummyResource(__provides__=meta.isheet)
+        assert get_sheet(context, meta.isheet)
```
joka: fine, all these "check the plug" tests produces al lot of  overhead, but refactoring this would be another task.

src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator2.py
```diff
@@ -0,0 +1,283 @@
...
+                                process_url,
+                                'participate')
+        assert resp.status_code == 200
+
+    def test_participate_participant_creates_proposal(self,
+                                                      registry,
+                                                      app,
+                                                      process_url,
+                                                      app_participant):
+        resp = _post_proposal(app_participant, path=process_url)
```
joka: Up to now we added all parameters for test methods in one line instead of seperated lines to 
save same space. To we want to change this?

joka:
Fine, i just added some general notes.

src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator2.py
pallix: I would be more enclined to be compliant with the code checker as the code quality of the tests should tend to be as good as the non-test code

pallix:
Mercator 2 resource needs to be Likeable and not Rateable once #1855 is merged

PR #1849, opened 2015-11-24
Revert "Parse URLs instead of doing string slicing that breaks in the presence of small variations in the URL"

nidico:
Reverts liqd/adhocracy3#1798

Unfortunately this breaks installations which make use of `SCRIPT_NAME`, which all our productive installation do (`SCRIPT_NAME = /api`).

@joka proposes to revert instead of fixing as this requires some further thoughts (e.g. the websocket server also uses the code in `schema`); also it might not be necessary for Thentos anymore.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/authentication/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/authentication/__init__.py
@@ -1,7 +1,6 @@
 """Authentication with support for token http headers."""
 import hashlib
 from datetime import datetime
-import urllib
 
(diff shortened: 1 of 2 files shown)


joka:
We already deploy the broken code on some server and the quick solution is this revert.
@fhartwig   is this causing problems with thentos?

fhartwig:
Yeah, that will break the thentos proxy for a3. But since there are apparently good reasons for keeping it this way, we'll just have to fix thentos to work with the status quo.

PR #1850, opened 2015-11-24
fix meinberlin i18n

xi:
meinberlin is missing some translation strings. I extracted them and already deployed the german translations on the server manually.


nidico:
LGTM.

PR #1853, opened 2015-11-25
Fix acceptance test "cannot login with short password"

nidico:
After #1794, the password field is validated in the frontend. This
removes the broken acceptance test.


xi:
I would prefer to check the frontend validation instead of removing the test.

nidico:
> I would prefer to check the frontend validation instead of removing the test.

ok, done.

xi:
Would be even better if we checked for the `.input-error`. But feel free to merge if travis reports success.

PR #1854, opened 2015-11-25
Restrict images types to JPEG

pallix:
This is done for performance reason and to reduce the size of the DB.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/sheets/image.py
+++ b/src/adhocracy_core/adhocracy_core/sheets/image.py
@@ -19,7 +19,7 @@ class IImageMetadata(IAssetMetadata):
     """Marker interface for images."""
 
 
-image_mime_type_validator = OneOf(('image/gif', 'image/jpeg', 'image/png'))
+image_mime_type_validator = OneOf(('image/jpeg'))
 
 
 class ImageMetadataSchema(AssetMetadataSchema):
--- a/src/adhocracy_core/adhocracy_core/sheets/test_image.py
+++ b/src/adhocracy_core/adhocracy_core/sheets/test_image.py
@@ -27,7 +27,7 @@ class TestImageMetadataSheet:
     def test_validate_mime_type(self, meta, context):
         inst = meta.sheet_class(meta, context)
         validator = inst.schema['mime_type'].validator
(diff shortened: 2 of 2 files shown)


pallix:
We have images that are megabytes big on mercator and we want to avoid that for mercator2 and other platforms.

nidico:
2 thoughts:

- AFAIK jpgs are not smaller than pngs for user-generated graphics such as diagrams (but mainly better for pictures)
- I think that not allowing png is not user-friendly, I would be annoyed... So if we really want to restrict this, we should transform to jpg in the scaled images.

joka:
> AFAIK jpgs are not smaller than pngs for user-generated graphics such as diagrams (but mainly better for pictures)

I did no more detail reasearch but on mercator there are big differences, for example:

   https://frontend.advocate-europe.eu/api/mercator/assets/0001586/0000000/  png, non-foto, scaled,  220kb
   https://frontend.advocate-europe.eu/api/mercator/assets/0001586/0000000/ jpg, foto, scaled, 45kb

Allowing only jpeg is the simplest solution and the project manage said its no problem for the customer.
If we want to deploy a platform without this restriction its quit easy to undo.

> I think that not allowing png is not user-friendly, I would be annoyed... So if we really want to restrict this, we should transform to jpg in the scaled images.

Yes this would be the better solution and even usefull if we allow only jpeg, then we could migrate old scaled images.

joka:
For me its ready to merge

joka:
I`m working on a branch to migrate png image scales to jepg image scales.
I the code change was easy, but  i need quite a lot of time to test the migration.

joka:
> I`m working on a branch to migrate png image scales to jepg image scales.
> I the code change was easy, but i need quite a lot of time to test the migration.

I did not mange to make the migration work due to problems with the zodb.

joka:
> I did not mange to make the migration work due to problems with the zodb.

ah, it all works with mercator if I disable filter_by_view_permission

joka:
With #1859 (force image scales to jpeg) on its way,  its also fine for me to allow jpeg and png uploads. 
But i would should definitly remove gif support, its not testet and i guess we don't want animated gifs

joka:
superseed by #1859

PR #1855, opened 2015-11-26
Separate IRate and ILike types FRONTEND NEEDS ADAPTATION

pallix:
Separate IRate and ILike resource to allow both kind of ratings on the same resource for AULA.

API CHANGE/EXTENSIONS:

- Postpool of Ilikable sheet only accepts the new ILike instead of IRead  resources 

- add ILike resource with field `like` to store the rate value (Should we rename this to `rate` to make ILike and IRate more similar??=

- add likes service pool

- add like/likes search index parameter

Internal Changes:

- migration script to migrate IRate to ILike resources on mercator and remove 
 not used  IRate versions

- remove IRateValidator to simplify backend

TODO:

- [ ] test migration with mercator production database

- [ ] adapt frontend code for mercator to post ILike  instead of IRate resources und use the likes instead of rates search index.


src/adhocracy_core/adhocracy_core/sheets/rate.py
```diff
@@ -189,20 +188,122 @@ class RateableSchema(colander.MappingSchema):
...
+        self.context = context
+
+    def validate(self, rate: int) -> bool:
+        """Validate the rate."""
+        return rate in self._allowed_values
+
+    def helpful_error_message(self) -> str:
+        """Return error message."""
+        return 'rate must be one of {}'.format(self._allowed_values)
+
```
joka: With to sepated resource types/sheets we don't need the special validator adapter for rates anymore.
Its much easier to understand if we check allowed values in the colander schema.

src/adhocracy_mercator/adhocracy_mercator/evolution/__init__.py
```diff
@@ -197,3 +254,4 @@ def includeme(config):  # pragma: no cover
     config.add_evolution_step(remove_mercator_workflow_assignment_sheet)
     config.add_evolution_step(make_mercator_proposals_badgeable)
     config.add_evolution_step(reindex_requested_funding)
+    config.add_evolution_step(convert_rates_to_likes)
```
joka: Should we migrate all rates to likes or do we have installations with "rates" instead of likes?

pallix: We have installations with rates, for example Alexanderplatz.

pallix:
I removed the like and rate validator.

src/adhocracy_mercator/adhocracy_mercator/evolution/__init__.py
pallix: and Kiezkassen

pallix: Likeable are only used in Mercator

joka:
Can you merge again?

pallix:
remerged

pallix:
@joka Tobias things it will not scale if we add other types of voting later. I have mixed feeling, I'm not sure having every "votable" things under the same interface help.

joka:
> @joka: @xi things it will not scale if we add other types of voting later. I have mixed feeling, I'm not sure having every "votable" things under the same interface help.

I m not concened about this. Having a clear separtion between votes and rates makes sense and what the future brings we don't know.

pallix:
The merge is not trivial now that the rate have been changed again

joka:
> The merge is not trivial now that the rate have been changed again

yes, merging master is now more work.
But we are still waiting for the frontend adaption.

pallix:
Converting this old PR into issue #2208 since we don't have time to work on it.

PR #1857, opened 2015-11-27
Add profile images to the frontend

dwt:
Hi there,

here's the pull request that adds profile image support to adhocracy3.

Please notice: This is mainly tested with s1, so I wouldn't be surprised if it doesn't work perfectly for the other projects. I expect that a) some small css tweaks are required to make it look perfect with other deployments and b) other projects would want to show the profile image in even more places. I haven't tested that yet, as I don't have detailed knowledge of the other projects and would like some feedback first / help doing that.

This includes that the flow library needs to be active for a project that wants to use profile image upload (without which the avatar feature doesn't really make sense). Not sure how to best handle that for other projects?

Also per an email conversation with Rouven, I added a new setting to the buildout config files to disable the avatar feature completely. The reasoning is that it has the potential to greatly slow down big adhocracy installations as big lists that show avatar images can potentially lead to linear increases in the number of requests. For example: If a list of 100 items shows an avatar image for each entry, up to three (one for the user, one for the asset, one for the image) additional requests to the server will have to be made for each entry. Thus avatar images are currently disabled for mercator - but not for all the other projects.

That being said, I think the avatar feature should be usable and quite easy to get working quite nicely for other deployments. 

I'd love to see some feedback on this.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
```diff
@@ -640,6 +642,69 @@ export var adhUserActivityOverviewDirective = (
...
+
+                adhHttp.get(scope.path).then((user) => {
+                    scope.assetPath = user.data[SIImageReference.nick].picture;
+                    scope.userName = user.data[SIUserBasic.nick].name;
+                    if ( ! scope.assetPath) {
+                        handleImageMissing();
+                    }
+               },
+               handleImageMissing
+               );
```
xi: This indentation looks off. First of all, there is a space missing for the last 3 lines. And I would also combine them into one: `}, handleImageMissing);`

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Meta.html
```diff
@@ -1,5 +1,7 @@
 <!-- as this is an inline directive, whitespace is relevant, including no newline at end of file
---><span data-ng-if="noLink">{{name || userBasic.name}}</span><a data-ng-href="{{ path | adhResourceUrl }}" data-ng-if="!noLink">{{name || userBasic.name}}</a>
+--><adh-user-profile-image data-css-class="user-meta" data-path="{{path}}"></adh-user-profile-image
```
xi: IMHO the profile image should be optional here because the meta directive is used in so many places.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
```diff
@@ -122,20 +136,41 @@ export var showImageDirective = (
...
+            isImageMissing: "=?"
         },
         link: (scope) => {
-            if ( ! scope.imageMetadataNick) { scope.imageMetadataNick = SIImageMetadata.nick; }
-            if ( ! scope.format) { scope.format = "detail"; }
-
-            var fallbackUrl = "/static/fallback_" + scope.format + ".jpg";
+            var imageMetadataNick = () =>
+                scope.imageMetadataNick ? scope.imageMetadataNick : SIImageMetadata.nick;
+            var format = () => scope.format ? scope.format : "detail";
```
xi: Why not simply `scope.format || "detail"`

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
```diff
@@ -122,20 +136,41 @@ export var showImageDirective = (
...
-                    (error) => { scope.imageUrl = fallbackUrl; }
+                    (asset) => {
+                        var imageUrl = asset.data[imageMetadataNick()][format()];
+                        if ( ! imageUrl) {
+                            return handleError();
+                        }
+
+                        scope.imageUrl = imageUrl;
+                    },
+                    handleError
```
xi: I don't really get the changes to `showImageDirective`. I understand that it is now possible to change most values after initialisation of the directive. But why would someone want to set `isImageMissing` from the outside?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
```diff
@@ -122,20 +136,41 @@ export var showImageDirective = (
...
 
             scope.$watch("path", (path) => {
                 // often instantiated before the path can be provided by the surrounding dom
                 if ( ! path) { return; }
                 adhHttp.get(scope.path).then(
-                    (asset) => { scope.imageUrl = asset.data[scope.imageMetadataNick][scope.format]; },
-                    (error) => { scope.imageUrl = fallbackUrl; }
+                    (asset) => {
+                        var imageUrl = asset.data[imageMetadataNick()][format()];
+                        if ( ! imageUrl) {
```
xi: This is the case only if the image is not available in the requested format. IMHO that's an error and should be logged to console, right?

buildout-mercator.cfg
```diff
@@ -9,6 +9,7 @@ dump-requirements-file = ${make_wheels:platform}-requirements.txt
 frontend.static_dir = src/mercator/mercator/build
 frontend_package_name = mercator
 backend_package_name = adhocracy_mercator
+frontend.profile_images_enabled = false
```
xi: @rukala this enables profile images for all projects except mercator. Is this desired from our part?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
```diff
@@ -93,9 +103,12 @@ export var uploadImageDirective = (
         templateUrl: adhConfig.pkg_path + pkgLocation + "/Upload.html",
         scope: {
             poolPath: "@",
-            path: "@"
+            path: "@",
+            didCompleteUpload: "&?",
+            didCancelUpload: "&?"
         },
         link: (scope) => {
+            scope.isUploadDone = false;
```
xi: This is not used.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserProfileImage.html
```diff
@@ -0,0 +1,9 @@
+<adh-show-image 
+    data-ng-if="config.profile_images_enabled"
+    data-path="{{assetPath}}" 
+    data-format="{{format || 'thumbnail' }}" 
+    data-css-class="{{ cssClass }} user-profile-image" 
+    data-alt="{{ 'TR__USER_PROFILE_IMAGE_OF' |translate:{userName: userName} }}"
+    data-fallback-url="/static/images/avatar_fallback.svg"
```
xi: What is the brower compatibility of svg images?

xi:
@local-girl could you please review the css in this branch?

xi:
This looks very good! Some small remarks, but I am looking forward to having this in master.

dwt:
@xi Thanks!

Regarding the questions I asked in the pull request, could you answer to those too please?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
dwt: The problem I found is that if angular updates the data-format property at any point later, the default set this way is overridden.

So the scope.format || 'detail' way to set a default only works if you do that in a template, but not in the link function that is not re-run when this happens.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_user_profile.scss
```diff
@@ -58,6 +58,64 @@
...
+    }
+
+    .user-profile-image-edit-button.user-profile-image-edit-button-force-visible {
+        @extend .button:hover;
+        visibility: visible;
+    }
+}
+
 .user-activity-overview {
     @include box-sizing(border-box);
```
local-girl: box-sizing is border-box by default (I know that's not this PR)

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
dwt: Oh, I set that as a =? mainly so someone can sense from the outside if that attribute has changed by binding to it. It's sorta a callback that way that you can still use to change rendering directly. (at least as far as I understand scopes).

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_user_profile.scss
```diff
@@ -58,6 +58,64 @@
...
+        @include border-radius(50%);
+
+        position: absolute;
+        top: 0;
+        left: 0;
+        background-color: rgba(0, 0, 0, 0.5);
+
+        visibility: hidden;
+
+        > span {
```
local-girl: We always use class names never elements

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_user_meta.scss
```diff
@@ -0,0 +1,6 @@
+.user-meta.user-profile-image {
```
local-girl: No need for a new file, I think this can be part of user_profile.scss

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_user_profile.scss
```diff
@@ -58,6 +58,64 @@
...
+    @include rem(width, $image-base-size);
+    position: relative;
+
+    // position nicely on top of the username
+    margin-top: 60px;
+    margin-bottom: -60px;
+
+    // center
+    margin-right: auto;
+    margin-left: auto;
```
local-girl: shorthand: margin: 60px auto -60px;

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_user_profile.scss
```diff
@@ -83,3 +141,7 @@
         > div:first-child { text-align: left; }
```
local-girl: please use class names not elements

xi:
> some small css tweaks are required to make it look perfect with other deployments

Looks fine for me (tried mercator).

> other projects would want to show the profile image in even more places

Couldn't think of any. But we can always add this later.

> This includes that the flow library needs to be active for a project that wants to use profile image upload 

Is this already the case? What happens if it is not available?

buildout-mercator.cfg
dwt: It is, according to an email conversation I had with him (as long as @rukala is Rouven Brüess)

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
dwt: I could switch to the callback format for this, it seemed more complicated that way, but you are right that the isolation would be better.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Meta.html
dwt: That is actually the whole idea for embedding the image in the user meta directive. If the username is shown, his profile image should be there too. I.e. many little splashes of color to the platform to make it more humane.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserProfileImage.html
dwt: http://caniuse.com/#feat=svg

I have found that we can use them safely with IE > 8 in other projects. Do you have other experiences? Should I add a fallback to a png?

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_user_profile.scss
dwt: It seems this default is in master already? I'll remove this and rebase the pull request then.

dwt:
@xi I've just had a talk with @nidico and he agreed to review these changes in the context of the other projects to check if the CSS holds up and activate the flow library in them. I think that right now the image upload directive will bomb if flow is not available, as the dependency injector will fail. So activating it in all projects is probably wise. :)

One last thing to do: change the way isImageMissing is handled to become a callback for better widget isolation.

dwt:
There's some work left for @nidico to activate and test the flow library in the remaining projects. But from my reading of the comments, this should be all that's required from me to get this merged.

The build server complains about some failures that I don't quite understand how I might have triggered. Could someone please elaborate?

Is there anything else in here that needs attention?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Meta.html
xi: I would still prefer to have this optional. It can default to `true`, but we should be able to deactivate it if needed.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
xi: I understand why you used a function. But `scope.format || "detail"` is shorter and less redundant than `scope.format ? scope.format : "detail"`.

xi:
> The build server complains about some failures that I don't quite understand how I might have triggered. Could someone please elaborate?

It looks like they are not related to this branch. I will investigate.

buildout-mercator.cfg
xi: I would like to disable this for all liqd projects for now. This will make it fully backwards compatible and allow us to activate it as we see fit. @nidico can you please take care of this?

nidico: Yes, I'll take care. I'm okay with disabling by default. We just had conversations in the past that it'd be good if contributions from external partners should be enabled in order to make liqd aware that they exist, but that can also be done in a separate PR.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Meta.html
nidico: For now, the whole feature will be disabled by default as you requested. We can add a local switch later if this is really needed.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
nidico: Done in a1e3a163ec635beed9d77af52a2d6c6e73d216a1.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
nidico: Done in dd297506a4dfef8af4b79c4b39ebcb88a83aeab3.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Image.ts
nidico: Fixed in 7a74d056 respective 04068e16f733ed5289245268a130db1d008bf470.

buildout-mercator.cfg
nidico: Done in 4722cadc134a13074d38fab731744847816826e8.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/UserProfileImage.html
nidico: Should be fine, we only support IE >= 10.

nidico:
Let's merge!

xi:
@dwt: I have some more feedback:

-  wouldn't it be nice if users could delete their profile images (go back to fallback)
-  The profile image edit button is not accessible via keyboard

PR #1858, opened 2015-11-28
revert revert empty first versions for batch request,  Fix #1735

### LINKED ISSUE #1735: Restore version optimization in backend
Restore version optimization in backend with

- [ ] proper event handling

      We are missing an event for "Create item versions finished".
      A general solution for this is complicated: 
     There is the empty first version  and in batch request there can be multiple 
     requests creating item version".  In this context we should 
     also  remove the empty first version that is needless anyway.
     But at the moment we need this only to send notifications about created bplan 
     versions and this is working now. So i would not invest time in this now.
   
- [x] bplaene working #1729 

- [x] permissions checks working #1723

- [x] tests working with batchmode?
   - more unit/integration tests added

Also related to #1733

joka:
API EXTENSION:

- Update first version instead of creating a new version within a batch request. This prevents creating an empty first version for every item we create.

- migration script to delete existing first versions hat have no sheet data.

INTERNAL CHANGES:

- fixed issues (#1729 and #1723) caused be reverted merge #1657 
- add test fixture 'mailer' to access send emails in functional test
- refactor code allowing anonymous to post one bplan item/version, more tests

### DIFF
--- a/src/adhocracy_core/adhocracy_core/changelog/subscriber.py
+++ b/src/adhocracy_core/adhocracy_core/changelog/subscriber.py
@@ -81,7 +81,7 @@ def add_changelog_followed(event):
         return
     _add_changelog(event.registry, event.object, key='followed_by',
                    value=event.new_version)
-    item = find_interface(event.object, IItem)
+    item = find_interface(event.new_version, IItem)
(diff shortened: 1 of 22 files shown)


pallix:
Can you merge master?

PR #1859, opened 2015-11-29
optimize file size of png and jpeg image scales

joka:
Motivation:

Optimize  file size of jpeg and png image scale downloads

Internal Changes:

- jepg scales are saved with optimize and interlace flag
- jepge scales image quality  increased from 75% to 80%
- png scales are reduces to 128 colors (reduces files size ~ 60%) 
- migration script to recreate existing image scales

Tested with mercator database.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/evolution/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/evolution/__init__.py
@@ -465,7 +465,8 @@ def remove_empty_first_versions(root):  # pragma: no cover
             del item['VERSION_0000000']
 
 
-def _is_version_without_data(version: IItemVersion) -> bool:
+def _is_version_without_data(version: IItemVersion)\
+        -> bool:  # pragma: no cover
     for attribute in version.__dict__:
         if attribute.startswith('_sheet_'):
             return False
@@ -475,7 +476,8 @@ def _is_version_without_data(version: IItemVersion) -> bool:
         return True
 
 
-def _has_follower(version: IItemVersion, registry: Registry) -> bool:
+def _has_follower(version: IItemVersion,
+                  registry: Registry) -> bool:  # pragma: no cover
     followed_by = get_sheet_field(version, IVersionable, 'followed_by',
                                   registry=registry)
     return followed_by != []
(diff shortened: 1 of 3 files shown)


xi:
I agree with https://github.com/liqd/adhocracy3/pull/1854#issuecomment-159675986 that this has significant impact on usability. We should make sure with designers and PO that this is really what we want.

nidico:
In contrast to #1854, this PR still allows users to upload pngs, they're just converted to jpgs. The original is still stored as-uploaded (if I see correctly). I'm fine with this.

I thought that pngs may be smaller than jpgs in corner cases, but I guess we can ignore this.

xi:
I still see two isses:

-  Users can no longer upload gifs (most users do not know how to convert images)
-  some images (e.g. line based graphs) have low quality when encoded as jpeg

nidico:
> Users can no longer upload gifs

Why not?

joka:
> Users can no longer upload gifs (most users do not know how to convert images)

We should discuss this  here: https://github.com/liqd/adhocracy3/pull/1854#issuecomment-160442419
This pull request is about converting images to jpeg when they are scaled/cropped by the backend.
Up to now we have now uploaded gifs and i don't even now how good the backend support gifs.

> some images (e.g. line based graphs) have low quality when encoded as jpeg

This could happen, but is this really important for our productive installations?   
We could suggest the user to upload an JPEG instead of png/gif as a workaround.

joka:
> I agree with #1854 (comment) that this has significant impact on usability. We should make sure with designers and PO that this is really what we want.

PO said allowing only jpegs is not problem.

nidico:
I'm a bit confused now. Doesn't this PR supersede #1854?

joka:
> some images (e.g. line based graphs) have low quality when encoded as jpeg

I test this again with the images in the current mercator database

joka:
> I'm a bit confused now. Doesn't this PR supersede #1854?

This is not clear yet.

src/adhocracy_core/adhocracy_core/evolution/__init__.py
```diff
@@ -506,6 +508,30 @@ def update_asset_download_children(root):  # pragma: no cover
...
+    This happens automatically if we recreate them.
+    """
+    from adhocracy_core.sheets.asset import IAssetMetadata
+    from adhocracy_core.sheets.image import IImageMetadata
+    from adhocracy_core.resources.image import add_image_size_downloads
+    from adhocracy_core.resources.image import IImageDownload
+    registry = get_current_registry(root)
+    catalogs = find_service(root, 'catalogs')
+    assets = _search_for_interfaces(catalogs, IAssetMetadata)
+    images = [x for x in assets if IImageMetadata.providedBy(x)]
```
pallix: why not search for IImageMetadata in the `_search_for_interfaces` function?

joka:
Example images from mercator:

joka:
This quality loss does not look so bad,  but but the transparent backround should not be replaced with black (fixable).

src/adhocracy_core/adhocracy_core/evolution/__init__.py
joka: > why not search for IImageMetadata in the _search_for_interfaces function?

We missed reindexing the interfaces index, wo the search for IImageMetadata resturns no result on the mercator platform.

joka:
I just tried an alternative solution, reduce the numer of colors for png files to 128 (might also lead to data loss):

joka:
Examle for lost data with 128 colors png:
0001586 detail
0001586 detail 256-colors

pallix:
Does the solution for PNG works on more complex images, like a photo?

joka:
> Does the solution for PNG works on more complex images, like a photo?

No, we can 

> Does the solution for PNG works on more complex images, like a photo?

No, it only works with PNG files that have max 128 colors

pallix:
One pydoc test fails

PR #1862, opened 2015-11-30
add maxwidth to proposal item image

### LINKED ISSUE #1852: Pictures in AE are too big for planned column-width
bildschirmfoto 2015-11-25 um 11 47 41

MagdaN:
fixes #1852


xi:
I think this is not the best solution. There is a class called `mercator-proposal-detail-view-image` that actually has `width: 100%` set. It was replaced by `proposal-list-item-image` in b442c116, probably by accident. I would propose to use the proper class instead.

PR #1864, opened 2015-12-01
refactor listing (part 1)

### LINKED ISSUE #1376: Frontend erroneously sends `count=true` to non-pool requests
Symptom: Requests with unexpected query parameters may lead to an outdated varnish cache, as the backend sends wrong purge requests.

@joka wrote a workaround in d4e3b689a443fae57e0e6cea35d3e9ffafb9a8d9, which basically purges "too much", but avoids the symptom.

Possible solution: in `Listing ... $scope.update`, call `getElements` with `warmup` as second parameter. This makes sure, count is only called for containers, which `canWarmup`, i.e. pools. This is a bit confusing, but should work.

xi:
*fixes #1376*

The original plan for adhListing was that it can show any list of resources, e.g. for pools (list children) and commentables (list comments). But later we added a lot of pool specific functionality resulting in complexity and issues like #1376.

In this pull request I converted the comment listing to not use adhListing. It still uses the same CSS though.

Not that I had thought that the createForm feature of the listing could be removed with the comment case gone. However, this feature is still used in adhBlog. I am not sure how to procede with that aspect.

In future pull requests I want to further refactor adhListing by

-   fixing "load more"
-   simplify directive interface (params, sort, sorts, …)
-   combine adhListing and adhMapListing into a single directive and move map switcher inside (completing the listing action bar)

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Adapter.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Adapter.ts
@@ -1,5 +1,4 @@
 import * as AdhComment from "./Comment";
-import * as AdhListing from "../Listing/Listing";
 import * as AdhUtil from "../Util/Util";
 
 import * as ResourcesBase from "../../ResourcesBase";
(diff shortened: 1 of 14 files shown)


MagdaN:
looks good

PR #1865, opened 2015-12-02
Add filter toggle buttons

local-girl:
Hi have added filter toggle buttons as described in https://tree.taiga.io/project/a3/task/557

They aren't the same color as we don't have that in the Mercator palette, have asked sara to provide another color, minor for now.


MagdaN:
this works nicely, just found a minor issue (there is no gap between the two filter buttons):
filters

xi:
Just took a quick look, but I think I found an issue: What happens if there are many facets selected? WIll it wrap? We should also add some `max-width` in combination with `text-overflow: ellipsis` (I think this is in the designs).

xi:
> there is no gap between the two filter buttons

fixed in  d019687

> What happens if there are many facets selected? WIll it wrap?

yes.

> We should also add some max-width in combination with text-overflow: ellipsis (I think this is in the designs).

done in 02a115b

PR #1866, opened 2015-12-03
Fix INSTALLATION: prefer virtualenv rather than compiling python

nidico:
This is what everybody does already anyway, so it's really a fix, not a change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,7 +29,7 @@ Softwarestack
 
 Server (backend):
 
-- `Python 3 <http://www.python.org>`_ (programming language)
+- `Python 3 <https://www.python.org>`_ (programming language)
 
 - `Pyramid <http://pylonsproject.org>`_  (web framework)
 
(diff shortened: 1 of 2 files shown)


dwt:
Looks good from my side - and it's what I did too anyway.

pallix:
LGTM

pallix:
I also have success with pyenv.

joka:
> I also have success with pyenv.

I would also prefere pyvenv, its good to use the "standard" way  python provides to do things

joka:
Can we close the taiga ticket regarding the same issue now? https://tree.taiga.io/project/a3/us/624

nidico:
> Can we close the taiga ticket regarding the same issue now?

Yes, done.

nidico:
> I would also prefere pyvenv, its good to use the "standard" way python provides to do things

Feel free to submit a PR, I agree pyvenv is better nowadays, I just forgot about it when I wrote the text.

xi:
> Can we close the taiga ticket regarding the same issue now? https://tree.taiga.io/project/a3/us/624
@xi I guess you created the taiga ticket

I did not create the ticket, so I don't mind if you close it.

PR #1871, opened 2015-12-07
check assign badge permission

joka:
API Extensions:

- If a user creates an IBadgeAssignment he need the `badge_assignment` permission for the referenced 
IBadge resource.


joka:
> can't we combine the validators here instead of combining them in the function above?

We could combine the two validators in one colander.ALL validation object.
But this does not work because one of them is "deferred" and colander.ALL does not support "deferred" validators yet.

PR #1872, opened 2015-12-08
2015 12 as extend and fix protractor tests

2e2a:
Fix and extend protractor frontend tests

* Create a test for meinberlin bplan
* Fix mercator tests
* Introduce a general approach to execute project specific tests depending on the current build


src/meinberlin/meinberlin/tests/acceptance/setup_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
pallix: On the detail page bash is not always in '/bin', it's often in /usr/bin/bash.

It would be better to use a Python script for the setup as it allows more powerful things to be done in the setup and reduce the dependencies on the bash or sh interpreter.
 
We also need one setup_tests (with 's'!) for each buildable frontend module, otherwise protractor will encounter an error trying to execute a script that is not there.

2e2a: Can change the script to python.

A missing setup_tests script does not lead to a protractor error, thus it could be added only to modules where its needed.

pallix:
Very nice. Tests are running okay on master locally after this merge.

PR #1760, opened 2015-10-14
Add agenda s-1 frontend

nidico:
Sorry for the huge PR with many merges, unfortunately this has turned out like that.

Review comments may best be added to the overall diff.

PR #1873, opened 2015-12-08
refactor listing (part 2)

xi:
*This continues #1864*

`adhListing` has many parameters for filtering and sorting. In this pull request I tried to find out how these are actually used and how they can be simplified. I came up with the following changes:

-   `params` was mostly used to set `tag: 'LAST'` which is implicitly added for versions anyway
-   Before, there were `sort` and `reverse` and than there was also `sorts` which was used when there was an UI to select between different sorts. The structure of `sorts` was slightly changed. `sort` can now be used to set an initial option from `sorts` (i.e. it can no longer be used on its own). If `sort` is not provided, the first option is used.
-   There was an additional frontend-based sorting system (`frontendOrderPredicate`/`frontendOrderReverse`). This was only used to have some reproducible sorting in case backend-based sorting was not available. So now the parameters are gone and this sorting happens automativally if `sorts` is empty.

I did not do extensive testing yet.


MagdaN:
looks good - I tested mercator

xi:
Some new listings have been merged in the meantime (e.g. #1760 and 1882). They need to be adapterd.

xi:
sorting in S1 still needs to be adapted.

PR #1875, opened 2015-12-08
2015 12 as increase btrees version

2e2a:
Increase BTrees version to 4.2.0


joka:
The pull request should be named in a way that everone (frontend/backend) has an impression what is this about, for example: "Update BTree dependency".
You can  also add the "backend" label to mark that this change is only backend related.
Further information about changes should  go to the description.

PR #1876, opened 2015-12-08
2015 12 cc mercator 2016 tests fixes

local-girl:
This has tests for validation but not for anything else as the form isn't finished yet.


xi:
I did some cleanup work. We should revert/fix protractorConf.js because this disables all core tests. Currently, two assertions are failing. Apart from that, this looks very good!

pallix:
ah!

It should be
```
core: "../src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/*Spec.js"
current: "../src/current/current/tests/acceptance/*.js",
```

xi:
I am done with my changes. @local-girl can you please review soon?

PR #1882, opened 2015-12-09
add listitem with dummydata

MagdaN:
I will also add the list.


MagdaN:
I also added the listing but could not test it because of missing data

PR #1883, opened 2015-12-09
Fix bplan confirmation email sending

2e2a:
The confirmation email sending is triggered upon the sheet
modification event, due to a change in the first version handling.


pallix:
Is the subscriber for the `IResourceCreatedAndAdded` still needed?

joka:
> Is the subscriber for the IResourceCreatedAndAdded event still needed?
No to make the productive system work.
But yes to be on the save side and make the functional backend tests work.

PR #1884, opened 2015-12-09
Make heard_from fields selectable multi-time

### LINKED ISSUE #1880: wrong type of heard_from
The `heard_from` field in `adhocracy_mercator.sheets.mercator2.ICommunity` should be a set, not just a single value.

pallix:
Fixes #1880


joka:
good,

Travis acceptance test are not working at all, but i guess this is not related.
I cannot test the acceptance test locally, they are very unstable on my system again.

joka:
> Travis acceptance test are not working at all, but i guess this is not related.

wrong, it is related, so we have to fix this as soon as possible

xi:
Fixed in #1889

PR #1885, opened 2015-12-09
mercator 2016 proposal basic http

xi:
This extends the mercator 2016 proposal form so that it actually sends a post requests. Unfortunately, it does not work yet.

I think one of the reasons is that there is a circular dependency between the proposal and its subresources (subresources are posted into proposal, `IMercatorSubResources` sheet on proposal references subresources). The proper fix is probably to first post the proposal without `IMercatorSubResources`, then all subresources, and then update the proposal to contain the actual `IMercatorSubResources` sheet. This could all be wrapped in a badge request (see mercator 2015 proposal for reference).


pallix:
@joka means it's normal and like mercator 2015. First POST an empty proposal then PUT.

It will work better within a batch request instead of a badge request :-)

xi:
Merging myself because this now containes many small fixes

PR #1888, opened 2015-12-10
2015 12 cc mercator 2016 form detail view

local-girl:
I have made the bare bones for the detail form complete with dummy data. I had to comment out some code (see FIXMEs) as the typescript wouldn't compile and I couldn't figure it out.

There are some translations missing and I have no idea what section needs comments and not.


MagdaN:
looks good

PR #1890, opened 2015-12-11
I thought I tested this, but apparently angular doesn't synthesize an…

dwt:
… empty callback if an optional callback isn't specified. Ah well.

Found when testing some more. :/


xi:
LGTM

PR #1896, opened 2015-12-11
remove tags resource type

joka:
API CHANGE:

   - resource type tag.ITag is gone, instead  tags.ITags sheet of IItem resources have the fields 'LAST' and 'FIRST' referencing the last and first itemversion.
    
   - 'LAST' tag is now a single references instead of  a reference list (multiple versions with the LAST tag only make sense  for Forkable resource types,  but this is only partly implemented in the backend.
Also its questionable if we need it all, so I should use this refactoring for  API simplification)

Open Questions:

   -  the item  "first_version_path"  is now completly useless overhead that we should remove.
      But i did not manage to adapt the frontend without help.

FRONTEND CHANGES:

   - adapt frontend to changed API 

BACKEND  CHANGES:

   - migration script 
   - remove helpers to get first or follow versions (you can use the normal get_sheet/field functions intead)
 
TESTING:

     - python, junit tests pass,
     - acceptances test pass on travis
     - manual acceptances test: create, edit, vote, comment mercator proposal, pass


pallix:
LGTM for the backend part.

joka:
@xi or @MagdaN  can you have a look at the frontend adaption commit  https://github.com/liqd/adhocracy3/commit/398c34261b0d3786fd3c3fde4ecbf23c416525af before everybody is leaving for Holidays?

nidico:
398c34261b0d3786fd3c3fde4ecbf23c416525af LGTM, tests pass, mercator-2016-proposal can be created (tested manually) after merge of master.

joka:
fine, @pallix  can you merge this and #1912

PR #1898, opened 2015-12-12
[RFC] thentos-adhocracy eggs for a3

fisx:
It is *probably* ok to use this in production as is (perhaps except for the thentos egg version that contains known bugs), but I would like to talk about it a bit more first.

In particular:

- instead of the top-level `thentos.cfg`, would you prefer a snippet that can be used elsewhere in `extends` clauses?
- which directory should `thentos.cfg` live in?
- is it ok that the platform (amd64) is not explicit in the egg, i.e. if you try to install the egg on other platforms, you will get runtime errors from supervisor about broken executables?
- the supervisor entry does not belong to any group.  is that ok and should be configured on a per-deployment basis?  if not, how do you extend existing groups in the supervisor builout recipe?

### DIFF
--- a/base.cfg
+++ b/base.cfg
@@ -6,6 +6,7 @@ extends =
     src/adhocracy_core/checkcode.cfg
     src/adhocracy_core/sphinx.cfg
     src/adhocracy_core/wheels.cfg
+    src/adhocracy_core/thentos_adhocracy.cfg
 #    src/adhocracy_core/varnish.cfg
     src/adhocracy_frontend/sources.cfg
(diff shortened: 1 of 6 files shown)


fisx:
(solved.  i'll comment here if i need more input.  sorry for the noise.)

fisx:
I think this is obsolete and can be closed.  See also: https://github.com/liqd/thentos/pull/444

@ssadler?

joka:
We deploy thentos now without python egge, i close this

PR #1899, opened 2015-12-14
Disable project specific acceptance tests on travis

2e2a:
As the tests are currently failing on Travis, while passing locally.


xi:
Now another test failed. But this is still an improvement.

xi:
btw I can verify that the tests work fine locally (except for the one for 2016)

Issue #1893, opened 2015-12-11
IName required on IPitch


xi:
`data.adhocracy_core.sheets.name.IName` is required on `adhocracy_mercator.resources.mercator2.IPitch` (it should not be). I guess this is also true for other mercator2 subresources.

pallix:
Solved by #1900

Issue #1891, opened 2015-12-11
Typo in ProjectStatusEnum: developping


xi:
On value of `adhocracy_mercator.sheets.mercator2.ProjectStatusEnum` is  `developping` instead of `developing`. I would fix it myself, but I guess a migration is needed.

pallix:
no need for migration as long as we don't have data in prod.

pallix:
Solved by #1900

PR #1900, opened 2015-12-14
Fix mercator resources

pallix:


pallix:
Fix #1893 and #1891

pallix:
I tested with @xi 'branch. It seems ok except `the practicalrelevance` field is set to null

PR #1904, opened 2015-12-14
2015 12 mn fix mercator2 form

MagdaN:
image is not uploaded. Any idea, @xi ?
Also practical_relevance is not set.


MagdaN:
when I create a proposal and it has a adhocracy_mercator.sheets.mercator2.IMercatorSubResources and a adhocracy_core.sheets.pool.IPool which contain basically links to the same ressources. Is this meant to be like that?

xi:
The IPool sheet simply contains references to all child resources of the pool. They have no further semantics. The fields in IMercatorSubResources hav actual semanatics.

PR #1905, opened 2015-12-14
Improve search performances

pallix:
Improve search performance when searching for references

- Resolve oids during search as late as possible (not sure if this tweak was efficient or not)
- Do not search for all IResources when searching for references without using the interfaces parameter

### DIFF
--- a/src/adhocracy_core/adhocracy_core/catalog/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/catalog/__init__.py
@@ -73,11 +73,11 @@ class CatalogsServiceAdhocracy(CatalogsService):
 
     def _get_interfaces_index_query(self, query):
         interfaces_value = self._get_query_value(query.interfaces)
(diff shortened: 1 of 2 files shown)


src/adhocracy_core/adhocracy_core/catalog/__init__.py
```diff
@@ -89,23 +111,76 @@ def _search_elements(self, query) -> IResultSet:
...
-            index = self.get_index('reference')
-            for reference in query.references:
-                referencence_elements = index.search_with_order(reference)
-                referencence_elements.resolver = elements.resolver
-                elements = referencence_elements.intersect(elements)
+                indexes.append(index_comparator(index_value))
+        return indexes
+
+    def _get_private_visibility_index_query(self, query):
+        if not query.only_visible:
```
joka: Return type is missing

pallix:
Return types added

PR #1906, opened 2015-12-14
Remove hardcoded meinberlin sample processes

2e2a:
The bplan and kiezkasse sample processes are imported during the tests
from a resource file


pallix:
Very nice.

Somehow the bplan is not listed in `http://localhost:6541/organisation/`. Is it still the case?

pallix:
I mean, what is the case with the hardcoded version?

2e2a:
The hardcoded bplan example process was removed. Thus organisation is now empty and needs to be populated via import_resources.

pallix:
Yes I understand, the problem ist that once imported, the bplan ist not listed within the organisation.

2e2a:
The bplan process is listed under organisation after the workflow is set to participate. This behavior is the same on the master branch.

pallix:
Ok! maybe I had a cache problem.

PR #1911, opened 2015-12-15
Fix error in search with unbound workflows

### LINKED ISSUE #1908: API call for list of proposals returns 500
GET http://localhost:6541/organisation/advocate-europe2?content_type=adhocracy_mercator.resources.mercator2.IMercatorProposal&count=true&elements=content

returns

{"errors":[{"location":"internal","name":"","description":"KeyError: 'workflow'; time: 2015-12-15 09:33:38,576"}],"status":"error"}

It works, if I remove the elements=content parameter

pallix:
Ensures the workflow is bound when using Colander serializations.

Fix #1908


joka:
test is missing, buts its quite complicated..

PR #1912, opened 2015-12-15
fix 1887 first version is  missing creator

joka:


joka:
> Can't we have just this commit in master to fix the bug? 
>Or does this commit also depends of the changes on the Tags?

both pull request change the same code/tests

PR #1913, opened 2015-12-16
Reorder fields in mercator 2016 proposal detail

xi:
Apart from some styling and replacing the dummy data by actual data, the detail view is pretty much done.


xi:
merging because urgent.

PR #1915, opened 2015-12-16
integrate mercator 2016 workbench

xi:
This fixes some smaller issues and makes it finally possible to see the create, listitem and detail directives in an integrated UI. To also get the right header you need to use the embed context defined in #1895.


xi:
merging because urgent

PR #1916, opened 2015-12-16
mercator 2015 comments

xi:
add some comment counts and such. Note that our 2015 approach for getting recursive comment counts does not work here because all comments for a proposal are in one single pool.


xi:
merging because urgent.

PR #1918, opened 2015-12-16
Added topics to filters, tidied a bit

local-girl:


xi:
small fixup in 5d4ae2b

PR #1919, opened 2015-12-16
implement image upload

xi:


xi:
merging because urgent

PR #1927, opened 2015-12-22
Fix #1921 search by mercator_topic not working

### LINKED ISSUE #1921: topic filter don't work on mercator
test here: https://plone-a3-stage.liqd.net/en/idea-space#!/r/advocate-europe/2016/

joka:


MagdaN:
Works

PR #1930, opened 2015-12-22
2015 12 mn fix mercator bugs

### LINKED ISSUE #1923: proposal detail ae partners section is broken
in the about section, even if no partners were added, the country of the organisation is shown three times

MagdaN:
fixes #1923 and also shows title in listitem


local-girl:
LGTM

PR #1936, opened 2016-01-04
Forced English on form

local-girl:


xi:
This will change the language to english for the whole application, not only the form. IMHO the is the wrong place to do that. If we want to force english on mercator2016, we should do that in the process package.

PR #1937, opened 2016-01-05
added hide-user attribute for space switcher, implemented in mercator…

local-girl:
Hi I don't know how to activate the 2016 context locally (can someone show me how please?) but I sufficiently tested this by overwriting the core header (and reverting)

it's quite simple so this should be self-explanatory


MagdaN:
In order to get this work, all links to users should removed. Instead of this I would suggest, to remove the userlist from the userspace but keep the detail column of the user, so that when someone clicks on a userlink he/she can still get the information abaout that user and write a message. @xi what do you think? Is this doable?

xi:
> Instead of this I would suggest, to remove the userlist from the userspace but keep the detail column of the user

I implemented this in #1967. That replaces this pull request.

PR #1938, opened 2016-01-05
Added cancel button

local-girl:


src/mercator/mercator/static/js/Packages/Mercator/2016/Workbench/ProposalCreateColumn.html
```diff
@@ -2,7 +2,9 @@
     <div data-ng-switch-when="menu">
         <i class="icon-document moving-column-icon"></i>
     </div>
-
+    <adh-resource-actions
+            data-resource-path="{{processUrl}}"
+            data-cancel="true"></adh-resource-actions>
```
xi: This needs a `data-ng-switch-when="body"`

xi: Fixed in #1962

PR #1939, opened 2016-01-05
Fix migrations errors

pallix:
The elements returned by the catalogs search are now returned as a
generator which do not support the len function on them so passing resolve=True solves the problem.

Alternatively if it would be more optimal to use the element with a generator but then we cannot indicate to the user in the console the total number of elements subject to migration.


ChristianSi:
I don't know why the checks failed, but I'd appreciate it if this could be merged soon-ish so that the A3 master becomes usable again.

pallix:
Tests are failing only because of #1934 so this PR can be merged

@joka can you review?

joka:
@pallix you forgot to run the unit test, i'll fix them

PR #1941, opened 2016-01-05
Caught number back end error

local-girl:
In the unlikely event someone put in a total budget or duration of more that 26 characters the FE will block it


local-girl:
Adding more to this

xi:
`maxlength` is not valid on number inputs ([spec](http://www.w3.org/TR/html5/forms.html#concept-input-apply))

PR #1942, opened 2016-01-05
2016 01 mn fix mercator 2015

MagdaN:
This fixes the problem that the edit and delete buttons are shown in mercator 2015 proposals to anonymous users

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
@@ -24,7 +24,7 @@ export var resourceActionsDirective = (
         },
         templateUrl: adhConfig.pkg_path + pkgLocation + "/ResourceActions.html",
         link: (scope, element) => {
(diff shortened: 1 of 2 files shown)


MagdaN:
Looks good to me - tested it a bit, but will test more when working on the frontend.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
```diff
@@ -24,7 +24,7 @@ export var resourceActionsDirective = (
         },
         templateUrl: adhConfig.pkg_path + pkgLocation + "/ResourceActions.html",
         link: (scope, element) => {
-            adhPermissions.bindScope(scope, () => scope.resourcePath && AdhUtil.parentPath(scope.resourcePath), "proposalItemOptions");
+            adhPermissions.bindScope(scope, scope.resourcePath, "proposalItemOptions");
```
xi: the second argument should be a function so that the permissions will be updated when `resourcePath` changes.

xi:
The implemented solution only works for non-versionable resources. I suggest to add a parameter for this and check the POST on the parent in that case.

xi:
Review no longer relevant

PR #1951, opened 2016-01-06
2015 12 mn enable edit

MagdaN:


local-girl:
Hi I made some progress tonight - mainly editing kind of works ! No console errors but there are some validation problems mainly I fixed Heard From checkboxes so they work in Edit but now not in create :-/ Actually I think I might have broken checkbox validation in general when moving this stuff to directive. I noticed when creating topics and heard from are not marked as required... Hopefully it's an easy fix cause I just have to leave right now, sorry.

local-girl:
Also best to clean up the controller - lots of unused things in there now

MagdaN:
At the moment this PR breaks the listing

xi:
replaced by #1961

PR #1953, opened 2016-01-07
Updated some help texts

local-girl:


xi:
Is this in sync with transifex?

xi:
I updated the translations in #1963

PR #1958, opened 2016-01-10
 add comments_count  field  to ICommentable sheet

joka:
API Extensions:

- add field ICommentable:comment_count: 
 
  number of all comments referenced by ISubresource (IDocument/IMercator2Subresources) and       IComment references. Example traverse:
document -elements-> paragraph  <-referes_to- comment1 <-refers_to- comment2

  visiblity is checked but not view permissions

Internal Changes:

- refactor rate validation

- 'traverse' search comparator to so traveres all references of a given type starting by the given resource.

- 'send_reference_event' option for the sheet set method (defaults True) 
  (This is more like a workaround.  For testing I need to create References without raising IBacksheetAdded events. The exiting similar option 'send_event'  did not provided this, but fixing 
  this would be quite complicated)

Migration:

- short running migration script to set initial  'comment_count' field values

TODO: 

- [x]  Make the fronted use the new comment_count field for the Mercator2Proposal

### DIFF
--- a/src/adhocracy_core/adhocracy_core/catalog/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/catalog/__init__.py
@@ -19,6 +19,7 @@ from adhocracy_core.interfaces import FieldComparator
 from adhocracy_core.interfaces import FieldSequenceComparator
 from adhocracy_core.interfaces import KeywordComparator
 from adhocracy_core.interfaces import KeywordSequenceComparator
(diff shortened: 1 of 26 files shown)


pallix:
LGTM for the backend part

MagdaN:
@xi: could you check the FE part?

xi:
FE seems to work fine. We could probably use the new API in many more places, but that can be done separately.

pallix:
youhoo!

Issue #1932, opened 2015-12-22
Simple API-CHANGES to improve performance


joka:
The follwoing proposals work without changing the frontend and are easy in 
the backend. They are meant as workarounds until we have a proper REST-API concept,
                                                   
1. needless frontend requests with huge performance impact:
                                                   
- /principals/users                                
- ../process/assets                                                                                         
- ../process/proposal/version/rates                
- ../process/proposal/version/comments             
- ../process/proposal/version/assets               
- ../process/proposal/version/badge_assignments    
                                                   
Proposal:                                          
                                                   
a) remove these resources from listing sheet.pool.IPool:elements (works except for /principal/users)
b) reduce response size: return search result with query parameter 'elements=omit' and 'count=True'.
                                                   
2. needless sheet fields with huge performance impact:
                                                   
- ..sheets.rate.IRateable:rates                    
- ..sheets.comment.ICommentable:comments           
- (..sheets.version.IVersionable:following, no big impact, but also needless)
                                                   
Proposa

c): remove these fields

pallix:
Can we close this? Everything has been implemented or?

xi:
+1. We should wait until #2020 is merged though.

pallix:
It's merged. Congrats!

PR #1959, opened 2016-01-10
by default don't list elements of IServicePool (performance workaround)

joka:
API Change

- IServie pool like (../comments, ..../rates, ../users/ don't list elements by default.
 Instead the elements count is shown. If you need all resources do a search query, 
 and limit search results, for example with .` ../comments/?serialization_form=content&limit=10`

Part of https://github.com/liqd/adhocracy3/issues/1932, 
Assumes: https://github.com/liqd/adhocracy3/pull/1958 is merged.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/sheets/pool.py
+++ b/src/adhocracy_core/adhocracy_core/sheets/pool.py
@@ -6,6 +6,7 @@ import colander
 
 from adhocracy_core.interfaces import ISheet
 from adhocracy_core.interfaces import SheetToSheet
+from adhocracy_core.interfaces import IServicePool
 from adhocracy_core.sheets import AnnotationRessourceSheet
 from adhocracy_core.sheets import sheet_meta
 from adhocracy_core.sheets import add_sheet_to_registry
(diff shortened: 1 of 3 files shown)


pallix:
LGTM. Can you document it in the REST API document since it impact the API?

pallix:
@xi or @MagdaN will proof manually that the frontend does not break

xi:
I did not have the time to look at this in detail, but it seems to be introduce major inconsitency into our API. I would really prefer to fix this in the frontend (I have some ideas how this could work)

joka:
> I did not have the time to look at this in detail, but it seems to be introduce major inconsitency into our API. I would really prefer to fix this in the frontend (I have some ideas how this could work)

This improves alot performance with the exinting frontend and and other clients that might do get request to our backend. On the ohter hand it does not brake anything and its documented .
Regarding "inconsitency"  i don't thing its worse that the current state of the API.
Improving the frontend is also possible but we have more important  issues at the moment.

For a proper solution for listiing large numbers of references i suggested we find a solution based on the json/hydra API standard.

xi:
I talked to @joka and here is what we came up with: The solution proposed here should be extended to all pools. This way the API does not loose consistency. The frontend does not use the IPool sheet a lot so there is no huge amount of adaption needed. And the performance improvement is even bigger because it applies to more cases.

Closing this PR in favor of the new concept.

Issue #1960, opened 2016-01-11
Cannot switch to user space AE Staging


local-girl:
This doesn't seem to happen locally. I have no idea how to start with this.

pallix:
Is this still relevant now that the user space is deleted?

xi:
fixed by #1966 and #1967

Issue #1965, opened 2016-01-12
New "home" concept


xi:
_This may not fit well as a github issue. Feel free to transfer this to some other place._

Most classic websites have a home page that can always be reached by clicking on the logo or title in the top left of the screen ([1. good deed in web design](https://www.nngroup.com/articles/ten-good-deeds-in-web-design/)). This way it is ensured that users can always go back to a familiar place in case they get lost.

In adhocracy, we have a variation of that where each space has its own home page. You can get there using the space switcher. In addition, the user is sometimes redirected to `/` which in turn redirects to a hardcoded location.

As with many things in adhocracy, this is half-baked and too complicated (see e.g. https://github.com/liqd/adhocracy3/issues/1640). In an effort to make things better we want to remove the space concept. The question is: How will it be replaced? This is what we came up with after some discussion:                

As said before, you usually reach home by clicking on the logo in the top left of the screen. In our case, this space is occupied by the embedding CMS. So that CMS is in charge of providing a home button. However, adhocracy can optionally provide its own home button to make embedding as simple as possible. 

For the case of redirecting to `/`, the following locations are used in decreasing preference:
-   `initialUrl` (plain/context embeds)
-   process (on single process setups)
-   process overview
-   If none if these is available, we have a problem.

Tasks
-   [x] remove space switch
-   [x] remove space concept
-   [x] add optional home button
-   [ ] change `/` redirection logic
-   [X] implement process overview
-   [x] define single process setups

pallix:
Do we need to distinguish between:

process (on single process setups)
and
process overview (= organisation containing processes, no?)

since both are contained in an organisation?

xi:
Consider the following case: If a platform only contains one process, there are two candidates for home: `/organisation/process/` (process) and `/organisation/` (process overview). If the platform is considered a "single process setup", the process should be used directly. Otherwise, the process overview should be used.

der-john:
This is as good as it will get.

Issue #1640, opened 2015-09-11
Make initial redirect configurable


slomo:
I have a few installations where the default redirect points to bogus locations, while it is any expected to be replace by a list of process in the future I'd like to have finer control for now.

xi:
I think there are 3 distinct issues here:

- #1378 introduced a config option that controls whether we link to a platform on register/password-reset or not. Maybe some link was overlooked there.

- In some places, the default for `adhTopLevelState.goToCameFrom()` is `"/"`. On `/` we usually redirect somewhere. This is configured in Adhocracy.ts with `adhTopLevelStateProvider.when("", …)`.

- When a user clicks on the "content" icon in the space switcher while already being in the content space, they are send to some url. This url is configured in Adhocracy.ts with `adhTopLevelStateProvider.space("content", …)`.

I propose to add config options for the last two bulletpoints.

xi:
> In some places, the default for adhTopLevelState.goToCameFrom() is "/". On / we usually redirect somewhere. This is configured in Adhocracy.ts with adhTopLevelStateProvider.when("", …).

would it help to redirect to `initialUrl` (set in plain/context embeds) instead?

PR #1966, opened 2016-01-12
rm space switch

xi:
*see #1965. fixes part of #1640.*

The space switcher (and therefore all visible traces of the space concept in the UI) are removed. The space logic itself is not touched.


xi:
*closes #1960*

PR #1967, opened 2016-01-12
remove user list

xi:
This removes the user listing column. The user listing itself is still in the code because we might want to use it again some time (I can remove it if you think that is a good idea).

This should be merged **after** #1966 because otherwise the user space homes will break.


MagdaN:
when I click on a user link the user details are not shown

xi:
> when I click on a user link the user details are not shown

I guess you checked with mercator2016. For some reason, there was no user space in the template. I added it in  29fc592, but I cannot actually remember why it was originally omitted. Is it right to add it now?

MagdaN:
works now excepts, that it says:
* {name}'s proposals 
* proposal of user are not shown

PR #1969, opened 2016-01-12
2016 01 16 thentos buildout

libscott:
Hi, could someone give me a hand to test/finish the buildout script for thentos?

It should be nearly ready just one or two tweaks :)

Please come grab me. Cheers,

Scott

### DIFF
--- a/src/adhocracy_core/thentos.cfg
+++ b/src/adhocracy_core/thentos.cfg
@@ -7,9 +7,9 @@ parts +=
 [thentos_download]
 # Download and extract thentos binary tarball
 recipe = hexagonit.recipe.download
-url = TODO
+url = https://github.com/liqd/thentos/releases/download/v0.1.0/thentos-captcha-v0.1.0.bin.tar.gz
 strip-top-level-dir = true
-md5sum = TODO
+md5sum = 6a863a985a6ff27f319674d214e221ca
 ignore-existing = true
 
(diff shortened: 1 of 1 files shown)


src/adhocracy_core/thentos.cfg
```diff
@@ -0,0 +1,27 @@
...
+strip-top-level-dir = true
+md5sum = 6a863a985a6ff27f319674d214e221ca
+ignore-existing = true
+
+[thentos_config]
+# Configure thentos
+recipe = collective.recipe.cmd
+on_install = true
+on_update = true
+cmds = [ -f ../thentos-captcha.config ] || cp thentos-captcha.config ..
```
joka: we use the inline templates for this, like here: https://github.com/liqd/adhocracy3/blob/master/src/adhocracy_core/checkcode.cfg#L74

..and sorry for the late answer

joka:
@fisx ping pong

fisx:
@joka, sorry for the low bandwidth.

I have a suggestion: since the development process only works when connected anyway, a simple alternative would be to use the production captcha server on development machines.  then it's just a matter of entering the resp. urls in the ini files.

If you decide there is a good reason to supported it locally in buildout, please let me know, and I'll look into this as soon as I can.

joka:
ok, i close this for now.

PR #1974, opened 2016-01-13
update text and fix typos

MagdaN:


src/mercator/mercator/static/i18n/mercator_en.json
```diff
@@ -189,6 +189,13 @@
...
-    "TR__USERS_PROPOSALS": "{name}'s proposals"
+    "TR__TWITTER": "Twitter",
+    "TR__USERS_PROPOSALS": "{name}'s proposals",
+    "TR__MERCATOR_FIRSTNAME": "First name",
+    "TR__MERCATOR_LASTNAME": "Last name",
+    "TR__MERCATOR_16_OTHER": "Other",
+    "TR__MERCATOR_16_WEBSITE": "Website",
+    "TR__MERCATOR_16_COUNTRY": "Country",
+    "TR__MERCATOR_16_ONLINE": "Online",
+    "TR__MERCATOR_16_NEWSLETTER": "Newsletter"
```
xi: you should use the extract_messages script instead of manually appending fields. This way the correct order is preserved.

xi: Also: I would prefere `TR__MERCATOR_2016_...`

MagdaN: Is the usage of the extract_message script documented somewhere?

xi:
merged and pushed the changes to transifex

PR #1979, opened 2016-01-14
Change audio captcha content type to make chromium/chrome happy.

### LINKED ISSUE #1977: html5 audio in chrome / chromium
`<audio src="...>` does not play; the control behaves like the audio is invalid.  firefox works fine.

Observed in the registration widget with captchas enabled.  Follow #1950 on how to reproduce this.  

@MagdaN: should I add any more details here, or can you work with this?

fisx:
Fixes #1977


fisx:
Thanks @fhartwig!

MagdaN:
Audio captcha is playing in chrome

PR #1980, opened 2016-01-16
extract frontend stats

xi:
After reading an article about the [Evolution of the Firefox Codebase](http://almossawi.com/firefox/) I spent some of my free time analysing the adhocracy3 frontend. Some time ago I had already created a script called `deps2dot` that extracts the graph of angular module dependencies from our code. This pull requests extends that script to provide more output options.

-  `--matrix`: Output dependency matrix as PBM. If you want to only include direct dependencies, use the `--direct` switch.

     

   Note that the order of modules has a huge impact on the outcome of this. I have experimented a bit, but I am not yet sure which order should be used.

-  `--stats`: Output some general stats
   -   Total number of modules.
   -   A categorization of modules into "core", "shared", "peripheral", and "control"
   -   Propagation cost
   -   Maximum rank

-  `--dump`: Dump data as JSON. Can later be loaded using the `-i` option.

In addition to extending the script, I collected data from different states of our git history and produced these graphs (note that the 2nd and 3rd graphs only cover the last 3 month because our import structure has changed at that time and deps2dot does not work on older code):

Get the [raw data](https://github.com/liqd/adhocracy3/files/92942/stats.txt) as csv.

## size

size_2014-11_2016-01

Both LOC and total number of modules have grown significantly in the last year. However, the number of modules is growing faster, which means smaller modules on average.

## modules

A module is categorized by the number of modules it depends on (fan-in) and the number of module that depend on it (fan-out).

fan-in | fan-out | category
--------|----------|---------------
high | high | core
high | low | control
low | hight | shared
low | low | peripheral

High/low means above/below average.

modules_2015-09_2016-01

We have next to no core modules (which is good!) and the biggest increase is in control modules. This is expected because we recently mainly added high level stuff like new workbenches. But this also shows that our abstractions actually work because we do not need to add core/shared modules all the time.

## general metrics

metrics_2015-09_2016-01

Code size, propagation cost and maxiumum rank have grown about the same in the past 3 month. Note however that this is a very small dataset (e.g. maximum rank changes from 11 to 12).

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/scripts/deps2dot.py
+++ b/src/adhocracy_frontend/adhocracy_frontend/scripts/deps2dot.py
@@ -261,7 +261,7 @@ def main():
         print('digraph adhocracy_frontend {')
         print('  graph [splines=ortho];')
 
-        for rank in range(args.min_rank, args.max_rank):
+        for rank in range(args.min_rank, args.max_rank + 1):
             print('  subgraph rank%i {' % rank)
             print('    rank = same;')
             for path, module in modules.items():
--- a/src/adhocracy_frontend/adhocracy_frontend/scripts/deps2dot.py
+++ b/src/adhocracy_frontend/adhocracy_frontend/scripts/deps2dot.py
@@ -55,13 +55,16 @@ def get_modules():
 
 def add_counts(modules):
     """Add import_count/export_count to an existing dict of modules."""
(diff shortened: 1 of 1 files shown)


joka:
interesting. 
All those commits are not helpfull, you could squash them to one. 
Also the code does not follow pep8, @xi can you fix this?

PR #1981, opened 2016-01-18
fix frontend linting errors

Johnschmidfront:
- fixed all linting errors, BUT
- value of img's alt-attribute is going to be changed by the Thentos team/ @fisx
- Travis is still going to fail because of other issues


xi:
Looks good to me. Feel free to merge once the alt attribute has been fixed.

fisx:
**tl;dr** decision should be made by a3 team, but I recommend [0] below.

@xi pointed out that the accessibility docs ask for alt fields in img elements that contain either describe / specify contents of the image (in our case, that would just be the solution and thus render the captcha pointless), or remain empty if the image has no function besides aesthetics (not applicable here).

I think captchas are a special case and not covered by the guidelines, so we have to make something up.  What about this?

- [0] "To circumvent visual anti-bot verification, please select audio captcha." / "Um das graphische Captcha zu umgehen, schalten Sie bitte auf Audio-Captcha um."

what other people do:

- [1] "captcha image" https://publixphere.net/i/publixphere-de/register?came_from=https%3A%2F%2Fpublixphere.net%2Fi%2Fpublixphere-de%2Finstance%2Fpublixphere-de
- [2] "visual verification" https://gmail.com/SignUp

I think [2] is even less helpful than [1].  I like [0] best.

PR #1992, opened 2016-01-20
Change organisation info fields in mercator 2016

pallix:
Remove contact email to avoid crawlers / spammers.


xi:
Frontend needs adaption, right?

pallix:
Yes

PR #1994, opened 2016-01-20
two more small mercator fixes

xi:


MagdaN:
@xi how is this supposed to work?

PR #1996, opened 2016-01-20
Omit elements in routing requests

xi:
This is similar to #1959, but it fixes the issue (at least one instance) in the frontend.

Unfortunately I only noticed that they solve the same issue after writing the code so I wanted to put it up for discussion. We should probably decide whether we prefer #1959 or this.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
@@ -236,9 +236,14 @@ export class Service implements AdhTopLevelState.IAreaInput {
             path = AdhUtil.parentPath(path);
         }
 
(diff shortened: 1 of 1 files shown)


joka:
> Unfortunately I only noticed that they solve the same issue after writing the code so I wanted to put it up for discussion

I guess the problem was we had to start making the API work  somehow,  without having much help for frontend team.

Regarding this pull request:

Are  you adding the get parameter  'omit'  to all get requests or only specific ones? 

Also I wonder  why the frontend does "non-search" get request to the service pools  like rates/users/comments at all..

xi:
> Are you adding the get parameter 'omit' to all get requests or only specific ones?

This PR is about a specific set of requests: When the user visits some URL, we traverse the content hierarchy to find the next process ancestor. These are the requests where `elements: 'omit'` is added.

xi:
Closing in favor of a newer concept (see https://github.com/liqd/adhocracy3/pull/1959#issuecomment-174509851)

PR #2002, opened 2016-01-21
2016 01 tb mercator fixes+1

local-girl:
Sorry I screwed this up earlier by merging it to master without realising it causes a 404 so this is a resubmission of #1994 - and I don't know what we're supposed to do about the 404


xi:
The 404 is caused by the fact that the initial redirect now points to `/r/advocate-europe/2016/`. On production, this is where the 2016 process lives. However, the development resource defined in  `src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator2/resources.json` lives in `/r/organisation/advocate-europe2/`. So there is no real issue, it is only that our development setup is different from production.

A proper fix would be to fix #1640. @Johnschmidfront could you please look into that?

Issue #1986, opened 2016-01-18
mercator2..extrainfo should be optional


xi:
When I try to submit a mercator2016 proposal without extrainfo I get the following error message:

```
{
  description: "Required",
  location: "body",
  name: "data.adhocracy_mercator.sheets.mercator2.IExtraInfo.extrainfo"
}
```

I am especially confused because the meta_api reports this field as `create_mandatory: false`.

xi:
Frontend has already been fixed in #1987

pallix:
I think there is no backend problem here. 

Extrainfo is a reference in the  `sheets.mercator2.IMercatorSubResources` and is not optional. 

What is optional is to post an extrainfo subresources with a `adhocracy_mercator.sheets.mercator2.IExtraInfo` a sheet.

Does that make sense?

xi:
How can I remove that sheet?

local-girl:
Any update on this ? This field needs to be optional

pallix:
Ok i think I understand better and tried to fix by #2004.

Can you retry?

local-girl:
I still get the required error
```
description: "Required"
name: "data.adhocracy_mercator.sheets.mercator2.IExtraInfo.extrainfo"
```

pallix:
It was an environment problem, it's working now.

PR #2004, opened 2016-01-21
Make extrainfo sheet optional

pallix:


local-girl:
@pallix I still get the required error

pallix:
See https://github.com/liqd/adhocracy3/issues/1986#issuecomment-173574736

PR #2005, opened 2016-01-21
exclude otherText from result

MagdaN:
@Johnschmidfront please check if the error still occurs


xi:
What issue does this fix?

xi:
Now I get it: `otherText` is send in the field `heard_from_other` and should therefore be excluded from `heard_froms`.

PR #2006, opened 2016-01-23
remove space concept

xi:
This continues on #1966 by removing most traces of the space concept from the code. They are still used to decide which set of columns to display (content/user/error), but that is the only place.

**Note:** Before this, callbacks registered via `adhTopLevelState.on()` were only triggered when something changed in the space that was active at the time of registration. Now, they are triggered on every change. This is important consequences for space switch: Before, when changing space, it was not important whether we first rendered the new space DOM and then triggered callbacks or the other way around, because the directives on the old space would not receive the events for the new space. 

One effect of this change can be seen in 7fc87b8: `adhProcessView` is only used in the content space. However, when switching to user space, it now receives the changes to `processType` *before* being removed from the DOM.


local-girl:
If you're happy I'm happy this is merged. It somewhat bothers me there's no easy way to get back to the proposal list, I suppose this will be handled in the redesign.

xi:
The translations look fine. But I found a lot of issues in [travis](https://travis-ci.org/liqd/adhocracy3/jobs/104590401). We should look into that before merging.

xi:
I fixed the tests and ran them with a different runner that seems to be more reliable. I will open a separate pull request for switching the runner.

xi:
Related to #1965

PR #2009, opened 2016-01-25
Mercator fixes

local-girl:
Rename awards and fix static links on create form. Fixing static links is hacky see #2011


xi:
LGTM.

PR #2013, opened 2016-01-25
Fix stadforum workflows

pallix:
Two workflows are implemented:
- one for the process
- one for each proposal

As before poll are modelled with a proposal. They are renamed in
poll (IPoll) since it was anyway necessary to add a new type to set the
workflow.

It may be necessary to refine permissions once we know more about the requirements. Also we should check if a migration is needed or not (it may not be needed if the old poll is not archived/accessible).

Initiator has the right to create a poll.


pallix:
@xo @MagdaN frontend needs small adaptation to post an IPoll instead of an IProposal

pallix:
@joka review?

@xi I tried the frontend and I don't see the title, even so the data from the backend seems correct with an ITitle sheet.

``` 

http://localhost:6551/embed/meinberlin-stadtforum-proposal-detail?path=http:%2F%2Flocalhost:6541%2Fstadtentwicklung%2Fstadtforum%2Fproposal_0000004%2FVERSION_0000000

```

xi:
> I tried the frontend and I don't see the title, even so the data from the backend seems correct with an ITitle sheet.
> 
> ```
> http://localhost:6551/embed/meinberlin-stadtforum-proposal-detail?path=http:%2F%2Flocalhost:6541%2Fstadtentwicklung%2Fstadtforum%2Fproposal_0000004%2FVERSION_0000000

> ```

I guess the path parameter was incorrect. You probably have to replace "proposal_0000004" by "poll_0000004".

pallix:
@xi It must be something else, the names of the objects weren't changed, just their types and interfaces.

xi:
> I tried the frontend and I don't see the title 

fixed in c26088b

pallix:
@xi cool, thanks!

pallix:
@joka review?

pallix:
@2e2a told me this is needed for next monday @joka

pallix:
Please review / merge ; it's already the second time I need to remerge master into it.

joka:
LGTM

PR #1957, opened 2016-01-07
Remove unused sheet fields: comments

pallix:
Workaround to improve performance. This pull request is part of issue #1932

API CHANGE:

- rm field: ICommentable:comments

TODO:

- [x]  frontend: all commentables should uses the provided comment_count field, see #1958
   (IDocument, IProposal, IParagraph,  mercator/meinberlin/spd proposal resources)

PR #2017, opened 2016-01-26
Use comments_count

### LINKED ISSUE #1972: Use comments_count provided by backend in frontend
#1958 introduced a new `comment_count` value that is provided by the backend. We should remove all custom queries from the frontend and use that value instead.

xi:
*fixes #1972*

This refactors the frontend code to use the `ICommentable.comments_count` field introduced in #1958. 

The only regression that I know of is that for mercator 2015, the list item only shows the comment count of the proposal itself, without the sub-resources. IMHO this is acceptable.

**Note:** I only tested mercator (2015/2016).

**Note:** This eliminates some uses of `ICommentable.comments`, but not all of them. So #1957 is still blocked.


local-girl:
Seems to work although I think this caused some Travis problems

PR #2019, opened 2016-01-28
made paragraphs deletable

Johnschmidfront:
Paragraphs of documents (e.g. of the spd workflow) can be deleted. There are two open issues:

1. Frontend: in Document/Create.html, I gave two elements misleading names to improve styling; the class names currently are `commentable-section chapter` and `commentable-section-show-comments`. @local-girl: Can you add classes to our stylesheets and replace these names please?

2. Backend: what the `postEdit` function now does is to ignore paragraphs with `deleted` set to true. Hereby they are really deleted in the newest version, but are still readable in older versions of the document. @xi @pallix Do we want this to stay like this?


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Document/Document.ts
```diff
@@ -72,6 +73,7 @@ export interface IScope extends angular.IScope {
 export interface IFormScope extends IScope {
     showError;
     addParagraph() : void;
+    deleteParagraph(number) : void;
```
xi: Needs to be `deleteParagraph(index : number) : void;`. Otherwise `number` is interpreted as a name rather than a type.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Document/Create.html
```diff
@@ -103,4 +107,4 @@ <h2 class="print-only section-jump-cover-header" data-aria-hidden="true">{{ data
     </div><!-- /.section-jump-body -->
 </div><!-- /.section-jump -->
 <!-- DATA SUBMIT -->
-</form>
\ No newline at end of file
+</form>
```
xi: Please avoid adding/removing final EOL (probably an editor setting)

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Document/Document.ts
```diff
@@ -262,7 +265,8 @@ export var postEdit = (
     var paragraphVersion : RIParagraphVersion;
 
     _.forEach(scope.data.paragraphs, (paragraph : IParagraph, index : number) => {
-        if (index >= oldParagraphVersions.length) {
+        // currently, if a paragraph has been deleted, it doesn't get posted at all.
+        if (index >= oldParagraphVersions.length && !paragraph.deleted) {
```
xi: I think it would be easier to read if you wrap an `if (!paragraph.deleted)` around the whole block.

xi:
I would have preferred if the UI changes were in a separate commit. But that is really only my personal preference.

xi:
Edit works great! However, create needs to be adapted, too.

local-girl:
We should have vaildation, currently you can delete all the paragraphs then submit the form in edit mode

Johnschmidfront:
@local-girl: Thank you for pointing that out! We (@xi, @rukala and I) talked about it and currently don't want to prevent documents without paragraphs.

local-girl:
Can we prevent it as a simple front end validation ? It seems really silly to me - I can take it on if you like...

local-girl:
PS I terms of UI I don't really see a problem with what we have except the icon is a bit lame, I would perhaps use a trash can, thought ? @xi @Johnschmidfront ?

local-girl:
Merge master pls

local-girl:
Please merge #2019 first @xi or @Johnschmidfront

xi:
The issues with 3ebdc48 is that sometimes you may want to delete the first paragraph, e.g. if you have 5 paragraphs and in the end realize that the first one is not necessary at all.

local-girl:
@xi done

xi:
LGTM

PR #2018, opened 2016-01-26
Added terms link URL for mercator on register

local-girl:
Adds terms and conditions to Mercator. This is ready for other sites should they need it

PR #2023, opened 2016-02-01
Add terms static links to buildout

local-girl:
Fixes buildout which I broke in #2018 (except mercator)


xi:
Wouldn't it be easier to add this to `src/adhocracy_frontend/base.cfg`?

local-girl:
I don't understand how :(

PR #2026, opened 2016-02-01
2016 02 cc rename commentable section

local-girl:
Commentable section can also be a deleteable section # #2019 so renamed to make it more generic and updated help


xi:
LGTM.

PR #1982, opened 2016-01-18
Extend organisation and process sheets

pallix:

PR #2027, opened 2016-02-02
(very) basic process overview

xi:
This adds an extremely basic process overview. The following issues still exist:

-  [ ] proper list items
-  [ ] tile layout for listings
-  [ ] do not automatically create `/adhocracy/` process
-  [ ] Many processes rely on an embed context to work properly. This is mostly used for replacing the header. When using a process overview, all processes are in the same context.


xi:
#1982 is needed for proper list items.

local-girl:
How do I see this ?

xi:
Its on `/r/`.

PR #2028, opened 2016-02-02
remove IVersion sheet followed_by field

joka:
part of https://github.com/liqd/adhocracy3/issues/1932

API CHANGE:

- rm 'followed_by" backreference field of IVersions sheet

This  has a small performance impact and this field is actually not used. 
If you want to traverse all versions you can use the 'first'  tag of the item and follow the 'follows' field, use the `elements`field or  do a search query.


pallix:
LGTM

joka:
> LGTM

@pallix  according to @xi we may merge this

xi:
I just realize that we **do** use this in the frontend. I will try to fix it first thing tomorrow.

PR #2030, opened 2016-02-02
tests needed to be adapted because the user space was removed

Johnschmidfront:


src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/UserPageSpec.js
```diff
@@ -9,15 +9,14 @@ var _ = require("lodash");
 
 describe("user page", function() {
     it("displays the correct name for each user", function() {
-        var usersListing = new UserPages.UsersListing().get();
+        var annotatorPage = new UserPages.UserPage().get("0000001");
```
xi: this line should be removed

src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/UserPages.js
```diff
@@ -166,6 +171,7 @@ module.exports = {
     ResetPasswordCreatePage: ResetPasswordCreatePage,
     ResetPasswordPage: ResetPasswordPage,
     UsersListing: UsersListing,
```
xi: `UsersListing` is not used anymore, so it should be removed.

Johnschmidfront:
@xi done.

PR #1993, opened 2016-01-20
mercator 2016 moderator fields

xi:
This adds a section to the proposal detail view displaying moderator fields. This also adds a new view (`@moderate`) where you can set these fields.

**Note:** Editing funding already works. But creating/editing the badge is not yet finished.

TODO:

* [ ] Validation
* [ ] create/edit Badge
* [x] add link to resouce actions directive

PR #2035, opened 2016-02-03
mercator 2016 moderator fields

xi:
*replaces #1993*

This adds a section to the proposal detail view displaying moderator fields. This also adds a new view (`@moderate`) where you can set these fields.

I did not implement any validation. It can be added later.


MagdaN:
there is no description near the dropdown:
screenshot - 09 02 2016 - 12 24 15

MagdaN:
can't submit: 

description: "This resource path does not exist."
location: "body"
name: "data.adhocracy_core.sheets.badge.IBadgeAssignment.badge"

xi:
I forgot to mention that the functionality depends on the existance of the badges defined in `src/adhocracy_mercator/adhocracy_mercator/scripts/badges.json`, but in the current process instead of `/mercator/`

MagdaN:
OK, now its working, but only the jury statement is shown - should there not be this yellow corner badge? Also, it is kind of weird, that the text field are grey and the labels white - is this supposed to be like that?

xi:
> should there not be this yellow corner badge

I did not yet implement that. As it is not urgent I would like to do it in a speparate pull request.

> Also, it is kind of weird, that the text field are grey and the labels white - is this supposed to be like that?

This is the general style of inputs on mercator, or am I missing something?

PR #2036, opened 2016-02-03
fix displaying creation date

xi:


xi:
Will open a new pull requests that combines multiple fixes

PR #2037, opened 2016-02-03
small mercator frontend fixes

xi:
-  fix creation date in detail and list item
-  change links to static pages in form
-  add missing validation


Johnschmidfront:
LGTM - links to static pages aren't working yet, but @2e2a will try them on staging.

PR #2007, opened 2016-01-25
Move SPD workflow to core as 'debate' workflow

pallix:

PR #2040, opened 2016-02-08
move spd workbench to core

xi:
This moves the former spd workbench to core and renames it "debate" (see also #2007). There was very little custom code, so it is pretty straight forward.


Johnschmidfront:
As @MagdaN said in #2059, now, comments aren't displayed when you click on the speechbubble.

PR #2042, opened 2016-02-09
Advocado Europe 2016 export script

pallix:


pallix:
@joka besser so?

joka:
> @joka besser so?

yup, 
To be more consistent we should also rename  the local variable 'is_in'  in `is_included`

pallix:
true, I missed one

pallix:
done

PR #2044, opened 2016-02-09
change implementation of conditionallyRedirectVersionToLast

xi:
`conditionallyRedirectVersionToLast` checks whether a resource URL points to a version that is not the last one. For this, it uses `IVersionable.followed_by`. This field was removed in #2028. So I changed the implementation. Now it checks whether the resource URL is different from `ITag.LAST` on the parent resource.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
@@ -255,7 +255,13 @@ export class Service implements AdhTopLevelState.IAreaInput {
 
         return self.adhHttp.get(AdhUtil.parentPath(resourceUrl)).then((item : ResourcesBase.Resource) => {
             if (item.data.hasOwnProperty(SITags.nick)) {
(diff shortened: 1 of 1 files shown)


joka:
yeah less code lines

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
```diff
@@ -253,18 +253,9 @@ export class Service implements AdhTopLevelState.IAreaInput {
...
-                    var itemUrl = AdhUtil.parentPath(resourceUrl);
-                    return self.adhHttp.getNewestVersionPathNoFork(itemUrl).then((lastUrl) => {
-                        self.$location.path(self.adhResourceUrlFilter(lastUrl));
-                        return true;
-                    });
-
-                }
+        return self.adhHttp.get(AdhUtil.parentPath(resourceUrl)).then((item : ResourcesBase.Resource) => {
+            if (item.data.hasOwnProperty(SITags.nick)) {
+                return item.data[SITags.nick].LAST !== resourceUrl;
```
Johnschmidfront: you forgot to reset the URI, too

PR #2045, opened 2016-02-09
2016 02 cc remove spd custom styles

local-girl:
Managed to remove all the SPD custom styles with no repercussions that i can see


xi:
Yeay!

PR #2047, opened 2016-02-09
opinion result toggle

xi:
This adds a toggle button to the opinion widget as used in meinberlin stadtforum. It allows to toggle between results and rate button in the cases in which previously it was only possible to vote. This has the advantage that users to not try to vote only to see the results. It has the disadvantage that users may be biased by seeing the results before voting.

I also added documentation for the opinion widget. The styling was changed quite a bit, so it should be reviewed with extra care.


Johnschmidfront:
LGTM.

PR #2049, opened 2016-02-10
2016 02 cc mercator refactor custom styles

local-girl:
I would like to have done more refactoring to the form, but perhaps its just better if its something we bare in mind when we create forms in the redesign, it's very complicated and messy at the moment.


xi:
Yeay! more simplification!

xi:
@stadine @dimarconicola: with the tinted background in input fields they are less prominent if the background is already tinted (`color-background-base-introvert`):

fieldset1

I will merge this anyway and we can fix that issue separately.

PR #2050, opened 2016-02-10
Make bplan process settings mandatory on create

2e2a:


joka:
This is usefull for the Imperia integration. Certain fields are required to make the backend work, so we should enforce this.

xi:
Bplan is the one frontend for which we have a frontend to create them, and you managed to break it :). Will fix as soon as possible. @Johnschmidfront maybe you could have a look?

PR #2054, opened 2016-02-10
fix captcha register

xi:
This works around the following issue:

If captchas are enabled, `IUser` requires an `ICaptcha` sheet. If captchas are not enabled, this sheet is not even available.

The frontend stripps all unavailable sheets from a resource before it is posted (`exportContent`). For this, it uses a dump of the meta api that is part of the frontend distrubution. If this dump was created on a setup without captchas, registration will fail in a setup with captchas enabled.

A proper fix could be to completely remove the `exportContent`/`importContent` magic.


2e2a:
Tested locally.

PR #2056, opened 2016-02-11
on missing permissions show comment create button

Johnschmidfront:
@stadine Styling can be added / modified.


Johnschmidfront:
thank you for laborious i18n

PR #1856, opened 2015-11-26
Relax AGPL license version

fisx:
See https://github.com/liqd/thentos/pull/393 for analogous thentos-PR, and https://github.com/liqd/thentos/issues/392 for the discussion.

TODO: You should check the first line of the new LICENSE.txt (2nd commit) and decide if that's a good slogan.  I was just making a wild guess.

PR #2057, opened 2016-02-13
Fix a minor typo

pra85:
Remove extra `some`


xi:
Thanks a lot!

Unfortunately, this can not easily be merged because of legal reasons: We are currently in the process of relaxing our license from "AGPLv3" to "AGPLv3 or later". Could you please add a comment that you agree to that change in #1856?

xi:
Thanks for the quick response!

pra85:
No problems :+1:

PR #2058, opened 2016-02-15
Store process start/end time in workflowassignment instead of processsettings

2e2a:


2e2a:
Evolve script and frontend adjustments are still required.

2e2a:
As far as I understand there is only one element in the state_data array so far.
If not, we would need to get the correct index.

We could also discuss, why state_data has to be an array.

xi:
You only changed the `IProcessSettings.participation_*`, right? The only place where this is used is in the frontend is `src/meinberlin/meinberlin/static/js/Packages/Meinberlin/Bplan/Process/Process.ts`.

2e2a:
@xi: Yes IProcessSettings.participation_ is removed and state_data in IWorkflowAssignment is used instead.

2e2a:
@pallix Fixed the state_data handling.

pallix:
LGTM, migration still needed

Johnschmidfront:
LGTM. @MagdaN could you review the frontend part please?

MagdaN:
How can I test it?

joka:
> how do we know that this information is at the first index?

This issue is not fixed yet, or?

2e2a:
>>how do we know that this information is at the first index?

>This issue is not fixed yet, or?

The should be fixed with the last commit. The state_data for 'participate' is used.

xi:
When testing this locally with a fresh database, I received the following error on creating a proposal version:

    description: "KeyError: 'name'; time: 2016-02-22 15:05:51,564"
    location: "internal"
    name: ""

Could this be because I left the `name` field in the `IWorkflowAssignment` sheet blank? (see  d0039d6)

2e2a:
@pallix: Added migration

pallix:
@xi will have a look at the error and then we can merge it

xi:
I am waiting for @2099 as it seems to be related.

xi:
I found a fix. Now it works for me.

pallix:
This PR breaks the 100% test coverage.

@2e2a fixes it and then we can merge it.

2e2a:
@pallix: the test coverage is broken because if merging master

PR #2062, opened 2016-02-15
Fine-tune stadtforum UI

xi:
Important note: The action bar in listings is now **below** the create form/button.


Johnschmidfront:
LGTM

PR #2064, opened 2016-02-16
fix acceptance tests

xi:


Johnschmidfront:
the only test that now fails is "user password reset: email is sent to user".

PR #2067, opened 2016-02-16
Implement debate workbench in meinberlin

Johnschmidfront:
Reverts liqd/adhocracy3#2040

At the moment, comments aren't displayed yet.


xi:
We should fix the issue instead of reverting the feature.

PR #2072, opened 2016-02-16
correct sorting

Johnschmidfront:
Before, when I sorted comments using the sort-by buttons of `Listing`, I got them in reverse order.


xi:
Your solution would revert the sort order for every sorting an all listings, not only for sort by rates in the comment listing. I did a different fix in c76a893 that only changes this specific case.

Johnschmidfront:
you're right. that was Semmelschmarren.

PR #2073, opened 2016-02-16
rm adapters

xi:
Adapters were meant as a layer between directives and resources. They were supposed to allow to use a single directive with different types of resources. Unfortunately, this did not help much and caused a lot of additional complexity.

So now I am removing adapters from the code.

This does not mean that the approach was wrong. Still, removing them now IMHO is the right decision because there are too many half-baked abstractions in our code. So this is a step towards temporary simplification so we can re-introduce better abstractions in the future.

Here are some random thoughts on what a better abstraction could look like:

-   Some calculations can be done in the backend (see for example #1958). This way the frontend does not need much abstraction.
-   Sometimes you do not need a full-blown adapter. See for example my solution for the rate directive, where you can still configure a `template` and a `sheetName`.
-   We intentionally restricted our adapters to be synchronous and stateless. This made them simple, but also very limited in functionality.


Johnschmidfront:
I whole-heartedly approve and confirm that it works.

PR #2074, opened 2016-02-17
fix mercator2016 edit if heard_from contains "other"

xi:
Some users reported the message `Required if "other" in heard_froms` when they tried to edit their proposals.

I fixed this by removing the `ICommunity` sheet from the post data on edit since it can not be changed anyway.

I also added some unrelated frontend validation that was missing.


2e2a:
Tested locally.

PR #2077, opened 2016-02-17
help link

xi:
This adds an optional link to some help page to the right top corner of the window/iframe.


Johnschmidfront:
laeuft.

PR #2082, opened 2016-02-19
Add spreadsheets to document global permissions and workflows standard, stadtforum

joka:
Add opendocument spreadsheets with defintion and description for all
global permission settings and the workflows: standard, stadtforum.

These definition are used to document the permission in a form suitable
for all stackholders.

They might be used to implement a a CSV import for permission/workflow or extend the workflow definition for processes (badges?, addables resource types?)

Todo: 

- establish workflow to keep this spreadsheets in sync with the source code defintions.
- add missing workflows


xi:
I would prefer some file format that is easier to work with in code, e.g. csv.

pallix:
@xi +1

pallix:
Text tables https://www.youtube.com/watch?v=EQAd41VAXWo ?

joka:
csv files are easy to parse but bad to read and modify.

formatted text files are not so easy to parse (at least i don't now how to do it), easy to ready but cumbersom to modify.

odfs files are nice to read and modify but its binary data (zip file). We could convert them to csv.
If we stay with ods files and but don't wont binary files in the repository, we could also store them
somewhere else, I just don't know where.

May be a better solution is to use fixed width data files?
The main down side is fields with long text data is not formatted nice.

pallix:
and something like JSON (or a simplified array in Python) ?

xi:
> I would prefer some file format that is easier to work with in code, e.g. csv.

I think the idea behind this request was that it should be easy to sync the permissions in the code with the permissions in the spreadsheets. So on second thought I am ok with ODS **if** we have scripts to sync them with the code (in both directions).

joka:
> and something like JSON (or a simplified array in Python) ?

Thats is not good to read/modify by non developers.

pallix:
> A fixed width data file would be some kind of simplified python array, or?

Sort of. I was thinking about code as data, like the workflows metadata, but simplified.

But fixed width data file may be better.

We need Markdown but for Excel :)

pallix:
MultiMarkdown tables: http://fletcher.github.io/MultiMarkdown-4/tables.html

joka:
> MultiMarkdown tables: http://fletcher.github.io/MultiMarkdown-4/tables.html
> 
> It can't be that hard to parse.

We could write a plugin for python markdown (supports the same syntax for tables) http://pythonhosted.org/Markdown/extensions/api.html to retun a dictonary with the table data.

joka:
> A fixed width data file would be some kind of simplified python array, or?
> Sort of. I was thinking about code as data, like the workflows metadata, but simplified.
> 
you are write its good to reuse the existing data structure for workflow metadata and 
to use a non binary format with nice diff. To make it easy to edit for non developers may be not 
so important. 

We forgot about YAML, this is an example with our existing workflow data, looks quite nice with spaces instead of tabs:

```yaml
---
id: sample
use_defaults_from:
name: Sample
description: Sample Workflow
initial_state: participate'
states':
    - participate:
        title: Participate,
        description: This phase is to participate,
        acm: principals:              [participant,  moderator,  creator, initiator],
             permissions: 
                 - [create_proposal,   Allow,        ,        ,         ,          ],
                 - [edit_proposal,     ,                  ,        ,         ,          ],
                 - [create_document,   Allow,      ,        ,         ,          ],
```

I also found this (yaml in markown files to json): https://www.npmjs.com/package/markdown-to-json

pallix:
The YAML solution is rather good. 

Maybe we can have another string instead of None? Maybe '_' ? I find Nones are not readable.

joka:
> The YAML solution is rather good.
> 
> Maybe we can have another string instead of None? Maybe '_' ? I find Nones are not readable.

we can use "~", pyaml load this to None

pallix:
that's good!

joka:
supersede by #2132

PR #2085, opened 2016-02-21
Remove authentication header 'X-User-Path'  

joka:
API-CHANGE:

The authentication header 'X-User-Path' is not validated anymore.

There is an exception if you set the config option
'adhocracy.validate_user_token' to false. This is a hack to make
adhocracy work with thentos and should be replaced with a proper setup
for `pyramid.authentication.IExternalAuthentication policy.

Internal Changes:

- cache authenticated_userid method for one request (its called for every permission check)
- refactor adhocracy_core.authentication, more strict  IAuthentictionPolicy implementation, no dependencies to other packages


joka:
> Why is this documentation removed?

ups

2e2a:
LGTM

PR #2088, opened 2016-02-23
2016 02 js add editorconfig

Johnschmidfront:
@xi and I think Editorconfig can help unify the whitespace/ newline settings of the development team. You probably need a plugin to use it in your editor - check editorconfig.org.


xi:
Nice! Could you send an announcement to the developers mailing list so everyone can install a plugin?

PR #2070, opened 2016-02-16
2016 02 mn set up euth proposal

MagdaN:
This adds the pc proposal to euth. To get this running run the following commands

bin/import_resources etc/development.ini src/adhocracy_euth/adhocracy_euth/resources/test_idea_collection/resources.json

bin/set_workflow_state etc/development.ini /opin/idea_collection/ announce participate

Browse to: http://localhost:6551/r/opin/idea_collection/

PR #2090, opened 2016-02-23
rm redundant context templates (breaking)

xi:
The context template in pcompass was used only to remove the header. This can now be achieved simply by using the `noheader` embed option.

Breaking changes:

-   The config option `adhocracy.custom.hide_header` was previously converted to boolean via `custom["hide_header"] === "true"`. This is now done via `custom["hide_header"]`. So, for example the string "false" is now interpreted as truthy.
-   The `pcompass` embed context now has a header by default. You can remove it with the `noheader` embed option.
-   The `noheader` embed option now also works for plain and context embeds. Note however that the standard header is overwritten in many cases, eliminating this effect. We are currently working on removing these overwrites.

**Note:** Everything I wrote about the pcompass context also applies to the euth context. See #2070.

@2e2a @slomo @MagdaN please update pcompass and euth embed codes before deploying.


Johnschmidfront:
works.

Issue #1694, opened 2015-09-28
Create a core proposal form style


localgrr:
We keep on duplicating proposal form styles, this should be in core

xi:
I think as a step it would be good to have some statistics of

-   which kinds of proposals do we have?
-   how similar are they?
-   home much code do we have for each one?

xi:
Current state:

-   `proposal-form` in core
-   `document-form` in core
-   `s1-proposal-form` in s1
-   `mercator-proposal-form` in mercator
-   `meinberlin-bplan-proposal-form` in meinberlin

IMHO we should not have a core proposal form style, but rather components from which a proposal form can be composed (e.g. inputs, erros messages, jump navigation, …)

xi:
As written in the previous comment, we have already made significant improvements, but we do not necessarily work towards a proposal form style, but rather reusable form components. Closing

PR #2091, opened 2016-02-23
2016 02 cc move pc proposal style to core

local-girl:
Simple PC styles are now core form styles this also help EUth. No other site uses this generic form as yet, it's something that I could look in to in the future but as the redesign is looming this is more of a quick fix for now.


xi:
As far as I can see the generic form is also used in meinberlin stadtforum. Could you please check whether it works there?

xi:
LGTM

xi:
I just realized that `.proposal-form` has `position: absolute`. This breaks the layout in stadtforum (not very visible), the style guide (very visible) and potential future usecases. @local-girl could you please have a look?

xi:
related to #1694

PR #2092, opened 2016-02-23
replace context header logic by custom headers

Johnschmidfront:
Up to now, contexts have been used, among others, for one rather confusing purpose, i.e. to provide custom templates for individual projects. We avoid this logic and have created `CustomHeader`s, instead. These are defined in `TopLevelState` and are stored in the individual projects' `Workbench` or `Proposal` folders.


xi:
I would prefer to have one change per commit. Some of your commits leave the system in an unusable state that is only fixed in the next or even a later commit. It is also a bit hard to understand which commits belong together. I really like that you made small commits, but maybe they are a bit too small.

Johnschmidfront:
@xi and I are splitting it up.

Johnschmidfront:
this branch has been thoroughly rebased and turned into #2109

PR #2094, opened 2016-02-24
fixup stadtforum

xi:
This fixes some issues that were reported by some test users of an upcoming stadtforum process.


Johnschmidfront:
I'm trying it out and it doesn't work. The "Create a Comment" button leads me nowhere, both when logged in and logged out.

Johnschmidfront:
Clearly something is wrong with my local stadtforum setup. So @2e2a and I will test the PR on staging.

PR #2098, opened 2016-02-24
add some translations

xi:
The euth project needs adhocracy in some languages that have not been supported until now. They provide the translations themselves. This pull requests adds a partial french translation as a first step.

I also added some incomplete country name translations. They are properly not needed for euth, but its nive to have them already.


Johnschmidfront:
Ça me plaît bien.

PR #2102, opened 2016-02-25
Add users export script

pallix:
@joka what's the syntax for type annotation of generators?

### DIFF
--- a/src/adhocracy_core/adhocracy_core/scripts/export_users.py
+++ b/src/adhocracy_core/adhocracy_core/scripts/export_users.py
@@ -1,0 +1,60 @@
+"""Export adhocracy3 users to CSV.
+
+This is registered as console script 'export_users' in setup.py.
+
+"""
+import argparse
+import csv
+import inspect
+from pyramid.paster import bootstrap
+from substanced.util import find_service
+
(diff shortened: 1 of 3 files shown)


joka:
> @joka what's the syntax for type annotation of generators?

The examples in the code are wrong

> @joka what's the syntax for type annotation of generators?

what we use is "[int]" or "Iterator" ( from collections.abc)
"Iterator" is the correct way but is missing the value type of the iterated objects.
With mypy we could add this information with "Interator[int]"

src/adhocracy_core/adhocracy_core/scripts/export_users.py
```diff
@@ -0,0 +1,60 @@
...
+                        quoting=csv.QUOTE_MINIMAL)
+        _write_users_to_csv(users, wr)
+    print('Users exported to {}'.format(filename))
+
+
+def _get_users(root: IResource) -> [IUser]:
+    users = find_service(root, 'principals', 'users')
+    for user in users.values():
+        if IUser.providedBy(user):
+            yield user
```
joka: Not DRY, we have similar code in various other places

PR #2106, opened 2016-02-25
euth idea collection image upload

xi:
This adds a simple image upload to the euth idea collection process. Note that with this change it is no longer identical to the pcompass process.

Some styling is missing, especially for the detail view. The upload widget does not look to nice either, but I fear there is not much we can do about that at this moment.


local-girl:
Still looks a bit crappy but its better.

Johnschmidfront:
I agree. LGTM.

Issue #2080, opened 2016-02-18
Don't show help button by default


MagdaN:
How can I disable the help button near the user name if there is no help url

xi:
This is actually an issue: If nothing is configured, the help button should not show. Currently, you have to configure a falsey value in order to disable it.

xi:
For reference: The help button was introduced in #2077.

PR #2108, opened 2016-02-26
On default, don't show help link

Johnschmidfront:
This should solve #2080.

I know this fix looks primitive. But I think it basically solves the
issue.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
```diff
@@ -840,13 +840,15 @@ export var adhUserProfileImageEditDirective = (
...
-    };
+    if (adhConfig.support_url.length > 0) {
+        return {
+            restrict: "E",
+            templateUrl: adhConfig.pkg_path + pkgLocation + "/HelpLink.html",
+            link: (scope) => {
+                scope.supportUrl = adhConfig.support_url;
+            }
+        };
+    }
```
xi: This change is not required as there is already a check in `HelpLink.html`

xi:
LGTM.

PR #2109, opened 2016-02-29
replace context template logic with custom headers

Johnschmidfront:


xi:
This is much much better structured, but still a bit much. I think it is better to split this up.

PR #2110, opened 2016-02-29
rm x-user-path from http request header

Johnschmidfront:
This is sensible because of #2085: the X-User-Path is not validated after authentication.


xi:
Works fine.

PR #2020, opened 2016-01-28
dont list elements of  IPool sheet by default (performance workaround)

joka:
API Change:

- IPool sheet elements field is empty by default (instead of listing the resource paths) 
   If you need the resource paths add as  custom search query parameter or set explicit how todisplay elements , like:  `http://../comments/?elements=paths 

- IPool sheet count field (number of resources) is now always shown

This is based on the discussion in https://github.com/liqd/adhocracy3/pull/1996 and https://github.com/liqd/adhocracy3/pull/1959

- [x] check that the frontend works with this api change

Issue #2022, opened 2016-02-01
don't use 'count' parameter for search request


joka:
Since https://github.com/liqd/adhocracy3/pull/2020 the 'count' parameter
is not needed anymore, because the count is always shown. 

Using this parameter is still possible but its deprecated and should be removed.

xi:
I put this on hold because #2020 is not merged yet.

xi:
Fixed by #2112

PR #2112, opened 2016-02-29
rm count parameter from pool queries

Johnschmidfront:
this is sensible because since #2020 the `count` parameter is always shown by default.


Johnschmidfront:
this addresses #2022

xi:
Great!

PR #2113, opened 2016-02-29
custom header infrastructure

xi:
This is the part of #2109 that provides the common infrastructure for custom headers.

The idea is that you can register a template url for a "custom header" for a process. The template will be included in the header.

In the next steps we want to use this to show buttons in the header. These buttons are already there, but the way they are injected requires a lot of boilerplate code and is counter-intuitive (they are bound to context, not process).

Another step would be to harmonize the interfaces of `adhResourceAreaProvider.template()` and `adhResourceAreaProvider.customHeader()`.


Johnschmidfront:
great.

PR #2117, opened 2016-02-29
Breaking: update mercator headers

xi:
*This was split from #2109.*

The "add proposal" buttons in the header now are picked based on process, not context. This means that there is no longer a button in the user space. It also means that the 2016 button is shown in the default context in a 2016 process (instead of the 2015 button that was displayed before).

It is still incomplete because the buttons do not work yet (Some scope variables like `setCameFromAndGo` are not defined and the argument to `setCameFromAndGo` is missing).

@slomo @2e2a: This contains breaking changes to the embed API: The `mercator2016` context is removed. Instead, a plain embed should be used.

Also note that this removes the `adh-busy` widget from the header in mercator. I would like to re-add it in a separate pull request.


xi:
Another reason why this is incomplete: The `mercator-print-header` is removed.

xi:
The remaining tasks have been implemented. However, there will be a separate pull request for getting the process workflow state in the mercator2016 button.

Johnschmidfront:
>This means that there is no longer a button in the user space.

Clarification: We aren't sure whether the "add proposal" button was displayed in the user space before. But now it can't be displayed there anymore because the user space is independent of processes.

Johnschmidfront:
great.

Johnschmidfront:
@xi and I still want to fix jsunit tests that fail due to other stuff.

Johnschmidfront:
merged fixes it in #2124.

Issue #1622, opened 2015-09-01
Load more button not working


MagdaN:
In item list the load more button is not working. If the sort parameter is removed it works nicely - if not the count value that is returned from the backend is wrong (its alsway the same as the limit)

Johnschmidfront:
@2e2a and I verified that, when in a pool query sort is set (to something relevant, like item_creation_date), the count is restricted by the limit parameter and fails to return the total count.

PR #2120, opened 2016-03-01
Fix element count when using sort and limit

2e2a:
Limit must not be applied when sorting, as element count computation and
slicing is done later.


pallix:
Can you write a test for that?

2e2a:
@pallix: Added a test

Johnschmidfront:
thank you! works just fine. 8 pyfunc tests are failing, but I don't think they're related ... ?

pallix:
@joka can you review?

xi:
I assume this fixed #1622?

2e2a:
@xi yes

joka:
cool

PR #2122, opened 2016-03-02
Add support for private processes in euth workbench

slomo:
Extend the workbench of euth to also render idea_collection.IPrivateProcess and collaborative_text.IPrivateProcess.


xi:
LGTM.

PR #2123, opened 2016-03-02
Make euth private example ressource really private

slomo:


pallix:
Thanks. 

This is a remaining of the first implementation try where private processes didn't have their own private workflows (and a separate process was not necessary).

PR #2124, opened 2016-03-02
fix jsUnit tests

Johnschmidfront:
This is a necessary adjustment due to #2110 (rm x-user-path from http headers).


Johnschmidfront:
due to #2113, there are 4 more tests failing in ResourceAreaSpec. I'm looking into it.

Johnschmidfront:
that's it - all jsunit test pass.

xi:
Great!

PR #2125, opened 2016-03-02
Breaking: change meinberlin contexts

Johnschmidfront:
this is made possible by #2113.

so, as of now, contexts do not affect headers or templates in meinberlin anymore, **except** in Alexanderplatz.


Johnschmidfront:
I noticed we don't check whether the process is in 'participate'. I'll add a check analogous to the one #2127. So that PR needs to be merged before this one.

xi:
I don't think we should remove the contexts completely, (1) because it is a breaking change in the embed API which we would have to communicate to the embedders and (2) because these contexts still differs from the default one by not having user routes.

xi:
Instead of not showing a button at all, all these cases had a deactivated version of the button.

xi:
You removed the phase headers. I guess this was not intentional?

Johnschmidfront:
unfortunately, we were a little too ambitious ... a slimmer PR will follow.

PR #2126, opened 2016-03-02
move busy bar styling to mercator

xi:
The busy bar was only ever used and styled for mercator. However, it is defined in core.

I moved the styling to mercator and added documentation in core. We may add styling in core at some later time.


Johnschmidfront:
LGTM

PR #2127, opened 2016-03-02
replace get workflow state

xi:
This saves the workflow state of the containing process in `adhTopLevelState`. This is simpler than the old approach. It also fixes the mercator 2016 add proposal button (see #2117).

**Note** that using the state from this process may not be what we want in all cases. It may be better to explicitly request the state of the next ancestor process of the widget resource instead of the one from routing.


Johnschmidfront:
I'm trying to see whether I can harmonize the usage of `processState`, `workflowState` and `currentPhase`.

Johnschmidfront:
That task is too big to be added here. Ready for review.

Johnschmidfront:
although I'd like to harmonize `processState`, `workflowState` and `currentPhase`, this PR is done and works.

Johnschmidfront:
works.

PR #2129, opened 2016-03-03
minor dependency updates

Johnschmidfront:


local-girl:
Everything seems to work, but when I go to comment on a proposal I get the unhelpful error "Empty Path" confirmed on SPD and Mercator

xi:
Please include a description of the changes in the description of pull requests like this. Also, calling this a "minor" update is an understatement IMHO.

-  TypeScript was updated to [1.8](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#typescript-18), containing many new features, some of them breaking:
    -   unreachable code is now reported as an error (if I understood that correctly) [potentially breaking]
    -   "string literal types" can be used to limit a variable to a fixed set of strings. We may be able to use this for accessing sheets on a resource (e.g. `resource.data[SIComment.nick]` would not compile if the keys in resource.data are restricted in this way).
    -   concatenated output with `--outFile` may allow us to remove r.js from our build toolchain.
    -   Modules are now emitted with a "use strict"; prologue [potentially breaking]
    -   colored output with the `--pretty` option
-  Protractor was updated to the next major release. Last time I tryed this it did not work at all. Did you verify that this works?
-  Lodash was updated to the next major release. Did you verify that this works?

PR #2132, opened 2016-03-06
use yaml files to define workflows and root permissions

joka:
Refector  workflow and permissions definitions:

- use IRootACMExtension adapter to override the permissions for the root
  object. See example in adhocracy_core.authorization.
  This way we can set the subscriber to update the root permission in adhocray_core and we have
  less boilerplate in extension packages.

- move root acm definition to adhocray_core.authorization (better SRP)

- refactor set_acm_for_app_root: use safe way to get the root object

- add yaml to python renderer (activated by default when rendering `.yaml` files)

- allow shortcuts in ACM permission tables: ` 'A' = Allow   'D' = Deny  '~' = None (yaml syntax)`
  This way we are less busy fixing the table layout. (Feel free to suggest better shortcuts)

- import root acm from yaml file (asset: `adhocracy_core.authorization:root_permissions.yaml`)

- import workflow metadata from yaml file

- add config directive 'add_workflow' to register workflow, example: 
     `config.add_workflow('adhocracy_core.workflows:debate.yaml', 'debate')`


src/adhocracy_core/adhocracy_core/workflows/__init__.py
```diff
@@ -36,28 +38,103 @@ def get_next_states(self, context, request: IRequest) -> list:
...
+                updated = updated.transform(['transitions', transition_name],
+                                            transition)
+        elif key == 'states':
+            for state_name, state in value.items():
+                for permission in state.get('acm', {}).get('permissions', []):
+                    name = permission[0]
+                    permissions = \
+                        updated['states'][state_name]['acm']['permissions']
+                    editing = name in [p[0] for p in permissions]
+                    if editing:
```
pallix: would it make more sense to name it `overwriting` ?

pallix:
ooops. I pushed a commit on the wrong branch.

pallix:
Reverted.

joka:
> why is registered on Interface here but on IResource elsewhere?

To make it easy to add a more specialized Adapter in extension packages, like for IResources or even IRootPool.

joka:
> why call f.close? It should be closed by the with construct.

I also rember it that way, but I could not find it  in the python documentation, are you sure?

pallix:
Yes.

"IOBase is also a context manager and therefore supports the with statement. In this example, file is closed after the with statement’s suite is finished—even if an exception occurs:"
https://docs.python.org/3/library/io.html#module-io
https://docs.python.org/3/library/functions.html#open

joka:
> why call f.close? It should be closed by the with construct.
> I also rember it that way, but I could not find it in the python documentation, are you sure?

you are right: https://docs.python.org/2/library/stdtypes.html#file.close

joka:
I rebased master

pallix:
I mean to rebase to remove my unwanted commit?

Otherwise all tests work locally for me, so it's ready to merge.

joka:
> I mean to rebase to remove my unwanted commit?

done

pallix:
LGTM

PR #2137, opened 2016-03-07
update lodash to 4.6.1

xi:
lodash 4.0.0 brought many breaking changes. This is why we did not update until now. I did not do extensive testing, but I thoroughly went through the breaking changes described in the [changelog](https://github.com/lodash/lodash/wiki/Changelog).

PR #2141, opened 2016-03-07
copy DefinitelyTyped definitions to our repo

xi:
*depends on #2137*

We are currently installing [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) with bower. This has some major drawbacks:

-  It contains definitions for over 1500 projects. We only use 5. Downloading everything is really not required.
-  The project is active, but not exactly fast when it comes to contributions. There are currently over 400 open pull requests and over 600 open issues. So if there is a bug, you may have to wait a long time before an upstream fix is merged.

For these reasons I propose to just add the relevant definition files to our repo. This would allow the following workflow:

-  fix a bug locally
-  create an upstream pull request
-  update our copy from DefinitelyTyped once in a while (you may have to manually port local changes)


Johnschmidfront:
works.

PR #2145, opened 2016-03-08
update frontend docs

xi:
This restructures and extends the frontend docs.

-  Most information about the frontend is now bundled under the "Frontend" entry.
-  I left some information where it was, e.g. coding style and testing.
-  I added a short introduction to some basic concepts (angular, typescript, backend api, customization packages, independent widgets and common services).

I still plan to add a "How To" section that explains how to do some common tasks. However, I would like to discuss this first to know whether I am moving in the right direction.

It would be great to get feedback from as many people as possible on this. Do you understand this? Does it help? Is the structure comprehensible? What is missing?

btw, the best way to test this is to checkout the branch and then run

```
$ bin/sphinx_build_adhocracy
$ xdg-open docs/build/html/index.html
```


docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
+Overview
+========
+
+The adhocracy 3 frontend is an opinionated web framework based on
```
pallix: as mentioned in the chat, if we document a3 as a framework, then it is our responsability that external developers can use what we write to develop their participation processes and get support if they find bugs in our software. It is very different from our actual responsabilities.

docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
...
+the building blocks that developers need to implement participation
+processes of all kinds. The user interface should be consistent and
+recognizable while providing enough flexibility for a wide range of
+target audiences as well as branding.
+
+Angular
+-------
+
+`AngularJS <https://angularjs.org/>`__ is an open source JavaScript
+framework mainly developed by google.
```
pallix: Google

docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
...
+
+In contrast to traditional frameworks like jQuery, you do not directly
+interact with the DOM. Instead, you only interact with data structures
+in JavaScript. The DOM is *bound* to these data structures and updates
+automatically.
+
+Other notable features are *services* (singleton objects), *dependency
```
pallix: I think the paragraph ist not helpful. For beginners it's too technical and for non-beginners it's too unclear. What do I know about angular services except that they are singleton? It does not convey much information.

I would suggest either make a small general description of the particularities of angular (directives allow to extend HTML for example) or link to a good intro.

docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
...
+injection* (a way to decouple your code), and *promise based APIs* (as
+opposed to callbacks).
+
+Angular is somewhat similar to other client-side rendering frameworks
+like `ember.js <http://emberjs.com/>`__ or
+`react <http://reactjs.com/>`__.
+
```
pallix: Let's not start a war.

docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
...
+TypeScript
+----------
+
+`TypeScript <http://www.typescriptlang.org/>`__ is an open-source
+language mainly developed by microsoft.
```
pallix: Microsoft

docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
...
+
+All resources that can contain other resources have an ``IPool`` sheet
+and are generally refered to as pools. Pools can be :ref:`queried
+<api-pool-queries>` for their contents. The results can be sorted and
+filtered by several conditions, some of which depend on the available
+sheets. Example: In order to get all comments that refer to resource
+``/my/resource``, you may use the following (simplified) query string::
+
+    ?depth=all&content_type=IComment&IComment:refers_to=/my/resource&sort=creation_date
+
```
pallix: I think this example is too complicate because it implicitely uses references

docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
...
+Batch requests
+++++++++++++++
+
+Sometimes you may want to change data in the backend, but it is not
+possible to do it in a single request. Doing it in two or more requests
+however has the risk that you end up with an inconsistent state because
+one of the later requests fails.
+
+For this case, the adhocracy backend allows to encode several requests
+in a single :ref:`batch request <batch-requests>` that is then processes
```
pallix: processed

docs/frontend/overview.rst
```diff
@@ -0,0 +1,172 @@
...
+frontend ignores all this and is only interested in the result: Is the
+current user allowed to do this action? All information required for
+that can be obtained by sending an :ref:`OPTIONS request
+<meta-api-options>` to the relevant backend endpoint.
+
+Websockets
+++++++++++
+
+The backend uses :ref:`websockets <api-websockets>` to notify the
+frontend whenever a resource changes. This can be use to update the UI
```
pallix: used

docs/frontend/routing.rst
```diff
@@ -102,6 +102,10 @@ In the frontend it also defines which template should be used.  The
 directive ``adhProcessView`` is used to render that template.  It is
 currently used in the resource area's content space.
 
+.. NOTE::
+
+   The UI for a process is sometimes refered to as a *workbench*.
```
pallix: Like in Eclipse :-(

docs/frontend/services.rst
```diff
@@ -0,0 +1,27 @@
...
+========
+
+This section gives a brief introduction to the most important services.
+
+adhConfig
+  Provides access to the configuration. Basically identical with
+  ``/config.json`` on the frontend.
+
+adhHttp
+  All HTTP communication should go through this services. Apart from
```
pallix: communications

pallix:
@Johnschmidfront as far as I know this configuration has no relation to build/buildout but to the `.ini` files and other backend options?

docs/frontend/overview.rst
xi: Do you have an idea for a simpler example?

docs/frontend/services.rst
xi: I think singular is actually right in this case.

xi:
@MagdaN @localgrr could you please give some feedback?

docs/frontend/overview.rst
xi: But what is it then? It is definitely not an application. Maybe a set of applications with a common core?

MagdaN:
Thanks @xi , this is going to help a lot - so yes  its moving in the right direction. What I found confusing, is that there are some parts, that I would not expect to be in the frontend chapter, e.g. the part about the API. Also, all parts are still too short for me to understand and to be useful to me during programming, maybe adding some example code would really help. The How-To-Section is a good idea, maybe we could even provide a tutorial which also includes BE-Modeling

xi:
> so yes its moving in the right direction

Thanks for the feedback! As the feedback was mostly positive and this is work-in-progress anyways I will merge this now. Feel free to comment (or create a pull request) if you want anything changed.

pallix:
If we add code sample to the frontend test, is there something like doctest to ensure that they will be kept up to date and compiling over time?

PR #2147, opened 2016-03-08
refactor sheet schema binding

joka:
Refator sheet schema bindings:
=========================

Changed files:
--------------------------

- interfaces.IResourceSheet
- sheets.BaseResourceSheet
- content.ResourceContentRegistry
- rest.views.RestView

other changes are mainly cleanups and adaptions to the changed IResourceSheet interface. 

Refactoring Goals:
-------------------------

One place to set places to set schema bindings for sheet data:

- interfaces.IResourceSheet.get_schema_with_binding

One place to set schema bindings for resource response/request schemas:

- rest.views.IRestView._create_schema

Fix defintion how to use bindings:

- `context`: Resource to set/get data, find services, traverse object hierarchy, ..

- `registry`: Pyramid registry to get adapters, sheets, metadata, settings, ...

- `request`: The current request or None. Use for extra permission checks, serialize/deserialize Resource-URLs, get authenticated user.

- `creating`: ResourceMetadata of the going to be created resource. Allow different behavior when validation data of new resources. If not None  `context` is the parent of the new resoure.

Fix definition what bindings to rely on:

- deferred validators: `creating`, `registry`, `request` , `context`

- defferred default values for sheet schemas: `context`, `registry`

Simplify serialization/deserialization with  IResourceSheets:

 - make `registry` parameter mandatory

- add optional `request` parameter

- new method `deserialize` (replaces `get_cstruct`)

- new method `serialize`

- updated documentation (`interfaces.IResourceSheet`)

One place to create sheets: ResourceContentRegistry (`registry.content`):

- alway use `get_sheet` method, removed helper `utils.get_sheet`
       (sheet objects are cached for every object after first usage)

- always use `get_sheet_field` method, removed helper `utils.get_sheet_field`

Validate json data when calling scripts.import_resources:

- added dummy request with god user permissions

Other improvements:
----------------------------

Fix: embed_code  serializes to empty string

Remove workarounds to get the right workflow when creating a new resource

Simplify schema for the workflow sheet

Add test fixture 'kw'  with default bindings for schemas

less needless calls of deferred validator/default value functions


joka:
Note: we pass around  both the registry and the request and use them for different porposes. Actually we don't have to do it that way and could just pass the request. The benifit would be to have one parameter less for some functions. The downside would be the request is a big black block with many functions. This is bad for unit testing or caching decorators.

joka:
Test coverage is broken also in master

joka:
> Test coverage is broken also in master

I suggest we readd the test coverage check to travis

pallix:
Very nice.

PR #2149, opened 2016-03-09
rm localsocket package

xi:
localsocket was planned to be used just like websocket. The difference would have been that the events would have been triggered in the frontend itself. However, we never used it.

If we ever need it, we can simply revert this commit.


Johnschmidfront:
LGTM.

PR #2151, opened 2016-03-09
2016 02 mn show login on private process

MagdaN:
This improves the errorhandling in order to handle private processes

@stadine: this needs a bit of styling at the moment it looks like that:

screenshot - 09 03 2016 - 16 02 50


xi:
This is a great improvement, thanks! Btw: this also improves the case when a process is in draft state and you do not have the permission to see it.

xi:
Travis reports failing jsunit tests. Could you please check?

MagdaN:
This thing is broken - when I return a 403 the user is automatically logged out. Do we want to keep it like this? 

Concerning the Jsunit tests - I get the following error (which I can not understand):


/home/magda/Documents/dev/adhocracy3/src/euth/euth/build/js/AdhocracySDKSpec.js:2
    define(["require", "exports"], function (require, exports) {
    ^
    ReferenceError: define is not defined

xi:
I cannot reproduce the issue on euth. Can you provide more context?

The failing tests were caused by a missing lodash import (fixed in 9c2912a). Note that this issue did not surface in the browser because `_` is available as a global object there.

xi:
> but the bigger problem is, that we cant show 403 at all as the user gets logged out 

For me it did not make any difference whether we raise 404 or 403 at this place.

I opened #2160 about the general session issue.

MagdaN:
@xi could you maybe take care of this?

xi:
@MagdaN I think I'm done. could you please review my changes?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/templates/Error.html
```diff
@@ -1,5 +1,4 @@
 <div class="error-page error-page-{{code}}" data-ng-switch="code">
     <img class="error-page-image" data-ng-switch-when="404" src="/static/images/404.png" alt="{{code}} - {{message | translate}}" />
-    <h1 class="error-page-title" data-ng-switch-when="403">{{ 'TR__NO_PERMISSIONS' | translate}}</h1>
     <h1 class="error-page-title" data-ng-switch-default="">{{code}} - {{message | translate}}</h1>
```
MagdaN: Did you test that? It did not work when I tested it.

slomo:
As far as I see the logout issue has not been addressed. Should we open up a separate issues for it?

xi:
> As far as I see the logout issue has not been addressed. Should we open up a separate issues for it?

Yes please. Note that I could not reproduce the issue when testing.

MagdaN:
@xi @slomo can I merge this?

xi:
IMHO this can be merged.

PR #2153, opened 2016-03-09
add mein.berlin.de context

xi:
this adds a context that is different from the default context in two aspects:

-   there are no user routes
-   the header consists only of the `customHeader` which renders the 'add proposal' button.

It is meant to be used on mein.berlin.de, where all other elements of the header (logo, help link, user indicator) are available in the embedding page.


Johnschmidfront:
works.

PR #2154, opened 2016-03-10
fix mercator

MagdaN:


Johnschmidfront:
@MagdaN and I think that adhLike and adhRate are actually responsible for checking whether the support buttons should appear, but they obviously fail to do so. So this is more a hack than an appropriate solution.

Johnschmidfront:
works.

xi:
Note: advocate-europe.eu is running an older version where `processState` was still called `data.currentPhase` (see #2127)

PR #2159, opened 2016-03-10
Add euth invitation email templates

slomo:
- use those for user imports


slomo:
I am not 100% sure that hose are super euth specific. I'd like if we could have some very basic invitation templates somewhere (maybe even core). Also if some one can contribute more languages please add them.

pallix:
@joka can you review this?

PR #2161, opened 2016-03-11
idea collection: keep image on edit

### LINKED ISSUE #2155: euth image upload: image removed after update
When I add an image to a proposal it works fine but when I update the proposal the image is gone and I have to add it again. 

@xi are you going to take care of this since you added this image upload featureß

xi:
This fixes #2155 by using `AdhResourceUtils.derive()`. We wrote this function exactly for this usecase some time ago but then did not use it consistently. It may be worth refactoring the code to use it in more places.


Johnschmidfront:
works.

PR #2163, opened 2016-03-14
Optimised embed for authentication only

xi:
Sometimes we want to display a user indicator on the embedding page. Embedding a complete adhocracy seems like overkill. So I created a custom HTML snippet that is optimised for exactly this usecase.

We need to have access to two distinct data sources:

-  the session token stored in `localStorage`
-  the backend API (access is limited by cross origin policy)

In order to access these two we need an iframe that is running in the adhocracy origin. We can then communicate with this iframe using the `postMessage` API. The code I wrote is similar to the `adhCrossWindowMessaging` service and offers the following messages:

-  on load, it will automatically send the user name (e.g. `{name: "userName", data: "god"}`)
-  when a user logs in, it will also send the user name
-  when a user logs out, it will send the message `{name: "logout"}`
-  when it receives the message `{name: "logout"}` it will log out.

For security reasons, the snippet will refuse to communicate with any website outside of the origin of `canonical_url`.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/__init__.py
+++ b/src/adhocracy_frontend/adhocracy_frontend/__init__.py
@@ -1,5 +1,7 @@
 """Frontend view and simple pyramid app configurations."""
 import pkg_resources
+from urllib.parse import urlparse
+from urllib.parse import urlunparse
 
 from pyramid.renderers import render
 from pyramid.config import Configurator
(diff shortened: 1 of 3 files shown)


xi:
This approach has serious limitations in some browsers (due to different 3rd party cookie handling). I put this on hold for now.

xi:
It turns out that the only way to get reliable access to `localStorage` is if the iframe is embedded on its own domain. If that is the case we do not need to embed it because the only reason for embedding it in the first place is to get code running on that domain.

PR #2169, opened 2016-03-16
Improve check-and-compile buildout steps

slomo:
- ensure templates do proper indenting
- rewrite `merge_static_directories` in python to not use `cp -s` option
- enable checking of script results


joka:
fine

Issue #1782, opened 2015-10-20
Have a specific database for each project


local-girl:
databaseis always written to var/Data.fs could it be say for mercator var/mercator/Data.fs to remove the need to restart supervisorctl and delete the database every time one switches sites - at least in development

pallix:
@slomo made the same suggestion. I would like too.

MagdaN:
me too

joka:
There different solutions, for example:

- just  copy the Data.fs file with wanted test data
- have a separated git checkout for every platform
- generate a  special development.ini  files for every platform.

pallix:
Fourth solution have something like `    path $INSTANCE/var/dbs/$PROJECT/Data.fs` into `zeo.conf`?

joka:
> Fourth solution have something like path $INSTANCE/var/dbs/$PROJECT/Data.fs into zeo.conf?

yes

xi:
Fixed by #2170

PR #2170, opened 2016-03-17
Use on zeo database per project

slomo:
- removes the need to delete/moves databases after running buildout with a different config
- fixes permissions of blob dir


pallix:
Fixes #1782

joka:
lgtm

PR #2172, opened 2016-03-18
refactor: use only colander schema for data validation, SRP

joka:
Internal Changes:

      - RestAPI view classes have no extra_validators hook anymore
        All validation has to happen in schemas defined in the class attributes schema_GET/POST/...

PR #2176, opened 2016-03-22
refactor: add decorator `api_view` to configure api views (more pyramid best practise, ease add prefix '/api'  to rest api later)

joka:
depends on #2172 

Internal changes:

- Remove RESTView

- Add decorator `api_view` to configure view classes\methods\function for
the rest api.
It works like the normal :class:`pyramid.view.view_config` decorator exept
the following defaults are set::

        renderer='json'
        decorator = [validate_user_headers,
                           validate_visibility,
                          set_cache_header]

    In addition it is possible to set a :term:`Schema` to validate therequest
    body or querystring::

        schema=MySchema

    If set the `validate_request_data` decorator is initialized with it and
    added to the decorator settings.

    Example::

        @api_view(context=IMyContentType,
                  schema=MySchema,
                  request_method=POST,
                  )
        def post_rest_view(context:IResource, request: IRequest) -> dict:
            appsturct = request.validated
            ...


joka:
There is a failing python test on master

src/adhocracy_core/adhocracy_core/rest/__init__.py
```diff
@@ -1,4 +1,74 @@
...
+    """ A function\class\method decorator to configure a rest api :term:`view`.
+
+    It works like the normal :class:`pyramid.view.view_config` decorator exept
+    the following defaults are set::
+
+        renderer='json'
+        decorator = [validate_user_headers,
+                     validate_visibility,
+                     set_cache_header]
+
```
joka: We could also add more defaults, like request_method, permission, accepted_content_type.
But I was unsure if we want to make it more implict what going on, what do you think?

pallix:
@joka pls merge master

src/adhocracy_core/adhocracy_core/rest/__init__.py
```diff
@@ -1,4 +1,74 @@
 """Configure rest api packages."""
+from pyramid.view import view_config
+from adhocracy_core.authentication import validate_user_headers
+from adhocracy_core.caching import set_cache_header
+from adhocracy_core.rest.schemas import validate_request_data
+from adhocracy_core.rest.schemas import validate_visibility
+
+
+class api_view(view_config):  # noqa
+    """ A function\class\method decorator to configure a rest api :term:`view`.
```
pallix: I think we should use `from functools import wraps` otherwise the docstring of the wrapped method will disappear when using this decorator.

joka: > I think we should use from functools import wraps otherwise the docstring of the wrapped method will disappear when using this decorator.

I have no strong opinion in this but but I would keep it that way to make it work like the normal view_config decorator.

src/adhocracy_core/adhocracy_core/rest/__init__.py
joka: > We could also add more defaults, like request_method, permission, accepted_content_type.
> But I was unsure if we want to make it more implict whats going on, what do you think?

@pallix any thoughts?

pallix: I'm not very familiar with this part of the code so I'll need more practice to answer your question. Let's leave it like this for now.

joka: > I'm not very familiar with this part of the code so I'll need more practice to answer your question. Let's leave it like this for now.

If you put more logic into the decorator you could write:

   @api_view()
    def get()

instead of 

@api_view(request_method='get', permission='view')
def get()

PR #2184, opened 2016-03-24
Fix merge static directories

2e2a:
Using absolute paths, as links with relative paths are not included in wheels.


Johnschmidfront:
works.

PR #2190, opened 2016-03-24
Add user export script for mercator 2016

2e2a:


2e2a:
@pallix , @joka: Can be merged

pallix:
The export script 2016 is a copy/paste no? It would be better to refactor the code.

2e2a:
@pallix: Yes its copy/paste, but I'm not sure if refactoring such one-time scripts is worth the effort

joka:
> @pallix: Yes its copy/paste, but I'm not sure if refactoring such one-time scripts is worth the effort

Next year we will need a similar script again, so its not really  a one time script.
To avoid code bloat I would use the same module (mercator_export_users), reuse helper methods and extend get_most_rated_proposals to work with different types of MercatorProposal.

2e2a:
> >@pallix: Yes its copy/paste, but I'm not sure if refactoring such one-time scripts is worth the effort

>Next year we will need a similar script again, so its not really a one time script.
>To avoid code bloat I would use the same module (mercator_export_users), reuse helper >methods and extend get_most_rated_proposals to work with different types of >MercatorProposal.

For the other scripts we did a copy for 2016 proposals instead of having the same script handling both resources.
Do we need to decide how to handle this in general to have a consistent scripts structure? Or just do it different for this script?

2e2a:
Moved common functionality into  a single script.

PR #2195, opened 2016-03-29
Fix calculation of commentCountTotal in mercator2016 proposal

### LINKED ISSUE #2189: Mercator 2016: comment_count stays 0
`comments_count` in the API displays the correct number of comments, but in the frontend the number is always 0.

xi:
*Fixes #2189*

`_.sum` does no longer work on objects (see #2137). This caused `commentCountTotal` to always be 0. This is the only place where `_.sum` is used in our code.

#2189 claimed that `commentCountTotal` was displayed correctly in the detail view, but I was not able to reproduce that.


Johnschmidfront:
works & looks good.

PR #2196, opened 2016-03-29
add debate workbench to the mein.berlin.de context

xi:


Johnschmidfront:
looks good. this will get tested when deployed and embedded into mein.berlin.de.

PR #2197, opened 2016-03-29
mercator 2016 badges

xi:
Badges for merctor 2016 are not completely implemented yet. This pull request provides the missing blocks to support the "winner" and "community award" badges that also existed in 2015. The new "shortlist" badge will be added separately.

@2e2a this assumes that the badges described in `src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator2/resources.json` exist.


Johnschmidfront:
the rosette in the listing is missing.

PR #2198, opened 2016-03-29
mercator 2016 shortlist badge

xi:
*Based on #2197*

This adds the shortlist badge to mercator 2016.

**Note:** I simply added the filter to the existing group, but I am not sure if this is the intended behavior: Currently it only shows up in the result phase, together with "winner" and "community award". Maybe it also needs to be selected by default in evaluate or something.


Johnschmidfront:
the comment on my commit is false.

Johnschmidfront:
looks good & works.

Issue #2188, opened 2016-03-24
comment count on meiberlin includes all versions of edited comments


2e2a:

joka:
This would not be the case if propsoal item instead of proposal version are commentable.
As long as we don't do this change we have to fix resources.subsriber.update_comment_count.
It should only  increase if the added comment has the 'FIRST' tag.

slomo:
This is an issue that blocks the release of the platform. How should we deal with it? Ideal would be a work around or solution until the 1st of april.

pallix:
Closed by #2200

PR #2200, opened 2016-03-30
Fix comments count code + fix all platforms with evolution script

### LINKED ISSUE #2194: rate count on meinberlin includes all versions of edited rates

pallix:
Fixes #2194  and #2188

### DIFF
--- a/src/adhocracy_core/adhocracy_core/evolution/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/evolution/__init__.py
@@ -630,6 +630,37 @@ def add_image_reference_to_proposals(root):  # pragma: no cover
     migrate_new_sheet(root, IProposalVersion, IImageReference)
 
 
+def reset_comment_count(root):  # pragma: no cover
+    """Reset comment_count for all ICommentables - See #2194, #2188."""
+    from adhocracy_core.resources.comment import ICommentVersion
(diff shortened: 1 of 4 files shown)


pallix:
@slomo can you have a look?

slomo:
Should be reviewed by someone with more backend experience.

src/adhocracy_core/adhocracy_core/evolution/__init__.py
```diff
@@ -630,6 +630,37 @@ def add_image_reference_to_proposals(root):  # pragma: no cover
...
+                                  resolve=True,
+                                  indexes={'tag': 'FIRST'})
+    comment_versions = catalogs.search(query).elements
+    count = len(comment_versions)
+    for index, comment in enumerate(comment_versions):
+        logger.info('Recalculate comment_count for resource {0} of {1}'
+                    .format(index + 1, count))
+        update_comments_count(comment, 1, registry)
+
+
```
joka: why did you not fix set_comment_count? Or we should add a Note that set_comments_count is broken and should not be used

pallix: I didn't fix it to ease deployment: we don't need to mark some evolution steps unfinished in Salt just to remove them after and to track for which platform it was done.

Issue #2182, opened 2016-03-23
Invalid varnish regex in deployment


pallix:
The regex in the varnish config should start with `^` for proper match.

pallix:
I am reclassifying this issue as a bug. Let me a give an example.

A `POST` request is issued on `/organisation/process/proposal0/comments`.

Backend sends the following URLs for PURGE to Varnish:
`/organisation/process/proposal0/`
`/organisation/process/`
`/organisation/`
`/`

They then constitute the ban list in varnish.

The last one is interessing: `/` is root but it is also is an organisation, so it is normal to PURGE it.

Since the varnish configuration uses this regular expression:
`req.url ~ " + req.url + req.http.X-Purge-Regex`

and that `X-Purge-Regex` is used only to match identical URLs but with the query parameters,
*any URL will be a MISS* since the last ban item `/` will match everything given the the expression in this case is `req.url ~ /<X-Purge-Regex>`.

For exampe `GET /organisation/process/proposal0/comments/c0` will have a MISS, but it should normally not.

slomo:
I am currently fixing this in the deployment scripts. Just one final taught: We are adding caret `^` now on the server but the dollar `$` is added by the regex provided, isn't that a bit wired?

pallix:
True, it would be more consistent to have the '$' in the varnish configuration.

slomo:
This has been fixed in deployment the varnish vcl does now the following:

```
ban("req.http.host == " + req.http.X-Purge-Host + " && req.url ~ ^" + req.url + req.http.X-Purge-Regex + "$");
```

Still a bit wired since everything is generated by backend. But at least consistent.

PR #2201, opened 2016-03-30
Fix purge regex for caching

pallix:
For consistency, since the regex is matched with '^' in the Varnish
configuration, '$' should be put there and not come from the backend.


pallix:
See https://github.com/liqd/adhocracy3/issues/2182#issuecomment-203512018

PR #2210, opened 2016-04-04
show shortlist filter in evaluate state

xi:
Completes #2198

### DIFF
--- a/src/mercator/mercator/static/js/Packages/Mercator/2016/Workbench/Workbench.ts
+++ b/src/mercator/mercator/static/js/Packages/Mercator/2016/Workbench/Workbench.ts
@@ -197,7 +197,15 @@ export var proposalListingColumnDirective = (
                         ]
                     }];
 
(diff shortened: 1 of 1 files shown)


src/mercator/mercator/static/js/Packages/Mercator/2016/Workbench/Workbench.ts
```diff
@@ -197,7 +197,15 @@ export var proposalListingColumnDirective = (
                         ]
                     }];
 
-                    if (currentPhase === "result") {
+                    if (currentPhase === "evaluate") {
```
Johnschmidfront: why not the following?
```
if (currentPhase === "evaluate" || currentPhase === "result")
```

xi: In evaluate, only the "shortlist" filter is availabl. In result, "winner" and "community award" arw added and "winner" is selected by default.

Issue #2209, opened 2016-04-04
Tests database should go in the db folder 


pallix:
The test database should go in the `var/db` folder like the others dbs and not in the `var` directory, if we want to be consequent wrt. #2170

pallix:
Fixed by #2218

PR #2218, opened 2016-04-05
Move test database to var/test

pallix:


pallix:
Can somebody explain to me why zeo cannot create `var/test/Data.fs` automatically whereas it works for normal db of other projects?

pallix:
Fixes #2209

PR #2219, opened 2016-04-06
Fix test coverage

joka:


pallix:
Why did the coverage indication disappeared again?

joka:
> Why did the coverage indication disappeared again?

I guess it only apears if the tests are sucessfull

PR #2225, opened 2016-04-12
only show chapter jump navigation if we have more than one chapter

### LINKED ISSUE #2140: only show chapter jump navigation if we have more than one chapter
As a user I only want to choose a chapter if I can choose different chapters. 

Here's an example:
https://opin.me/en/projects/commenting-texts-testgroup/#!r/test/collaborative_text/

@xi you can delete the issue if it's not a feature request.

Johnschmidfront:
this should fix #2140. plus, there's a resource json edit that hasn't got anything to do with it - sorry.


Johnschmidfront:
the gray little column ist still displayed. Should it be removed if there's only one chapter?

Johnschmidfront:
looks good & works.

PR #2229, opened 2016-04-13
rm phantomjs

### LINKED ISSUE #2226: Build is broken
A3 build is broken, due to a phantomjs dependency that fails to be downloaded. I tried to remove the dependencies but then there is another error with nodejs being invoked with the wrong name ("nodejs" instead of "node").

xi:
*Fixes #2226* I could not reproduce the issues reported by @pallix.

This removes phantomjs from our build scripts. It was used for browser tests some time ago, but we stopped doing this when switching to protractor, which [explicitly recommends against using it](https://github.com/angular/protractor/blob/master/docs/browser-setup.md#setting-up-phantomjs).

**Note** that you have to remove `parts/` and `.installed.cfg` to trigger a full reinstall.

### DIFF
--- a/src/adhocracy_core/versions.cfg
+++ b/src/adhocracy_core/versions.cfg
@@ -21,7 +21,6 @@ flake8 = 2.4.1
 flake8-docstrings = 0.2.1.post1
 flake8-quotes = 0.0.1
 gp.recipe.node = 0.10.28.0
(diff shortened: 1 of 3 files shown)


Johnschmidfront:
works.

pallix:
Thanks! Maybe we need a force-reinstall option in buildout  :)

PR #2230, opened 2016-04-13
rm document workbench

xi:
The document workbench was the first prototype of a workbench we worked on, even before mercator 2015. I was surprised that the code still exists. Tt is completely out of date, the used directives have been removed long ago and it was not even included in most builds.

So removing this code should be completely safe.


Johnschmidfront:
good idea.

PR #2232, opened 2016-04-14
Add comment placeholder to comment creation form so that it works the…

dwt:
… same as other create forms.

Should be a straightforward change - let me know if there are open questions.


Johnschmidfront:
looks good & works. Design (Nico) agrees.

PR #2233, opened 2016-04-14
Custom error pages for Euth

localgrr:
Simple error pages for Euth. Unfortunately it's not quite like Nadines simple design as I can't really make the whole background blue without really unnecessary code, and I'm not going to install the Euth font to adhocracy just for this either, still it's fine enough :)


MagdaN:
LGTM

xi:
Wouldn't it have been simpler to not overwrite the template but only the image?

PR #2234, opened 2016-04-15
fix stadtforum embedding on old mein.berlin.de

### LINKED ISSUE #2231: old mein.berlin.de: Stadtforum embedding is broken
When I navigate to mein.berlin.de and click on a Stadtforum process (e.g. https://www.berlin.de/familie/umfrage), I get a broken (i.e. I can't post comments or rates) embedded process and the error message:

> GET https://mein.berlin.de/apil_0000000/VERSION_0000001 404 (Not Found)

while the URL that should have been accessed is:

> https://mein.berlin.de/api/familienportal/dialog/poll_0000000/VERSION_0000001/

So a part of the path is swallowed by the frontend routing.

Johnschmidfront:
this should fix #2231.

I couldn't test it, unfortunately, because I could never reproduce the error in the issue locally.

But I think I found the problem:

In `ResourceArea.ts`, the function `getProcess` assumed that its input string `resourceUrl` was a complete URL, not a relative path, and then chopped it accordingly.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
@@ -242,7 +242,10 @@ export class Service implements AdhTopLevelState.IAreaInput {
     public getProcess(resourceUrl : string, fail = true) : angular.IPromise<ResourcesBase.Resource> {
         var paths = [];
 
-        var path = resourceUrl.substr(this.adhConfig.rest_url.length);
+        if (resourceUrl.substr(0, 4) === "http") {
+            var path = resourceUrl.substr(this.adhConfig.rest_url.length);
+        }
+
         while (path !== AdhUtil.parentPath(path)) {
             paths.push(path);
             path = AdhUtil.parentPath(path);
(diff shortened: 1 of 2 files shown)


2e2a:
LGTM.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceArea/ResourceArea.ts
```diff
@@ -242,7 +242,12 @@ export class Service implements AdhTopLevelState.IAreaInput {
     public getProcess(resourceUrl : string, fail = true) : angular.IPromise<ResourcesBase.Resource> {
         var paths = [];
 
-        var path = resourceUrl.substr(this.adhConfig.rest_url.length);
+        // FIXME: This if-Statement was added for the special case that resourceUrl is a URL,
+        // not a relative path. It should be erased when the archived Stadtforum processes go offline.
+        if (resourceUrl.substr(0, 4) === "http") {
+            var path = resourceUrl.substr(this.adhConfig.rest_url.length);
+        }
```
xi: `path` is `undefined` if this block is not executed.

Johnschmidfront: omg. horrible mistake.

PR #2236, opened 2016-04-18
Temporary fix for private processes

pallix:
Fixes only 'participate' state. 

All participants have both the 'participant' and 'everyone' roles, so
setting 'Deny' for 'view' permission for 'everyone' result in a 'Deny',
if participant does not have explicity 'Allow' in the private workflow.
Setting explicitely 'Allow' do not allow us to us inheritance with the ~
symbol.

@joka what do you think about the problem?

### DIFF
--- a/src/adhocracy_core/adhocracy_core/workflows/debate_private.yaml
+++ b/src/adhocracy_core/adhocracy_core/workflows/debate_private.yaml
@@ -9,7 +9,7 @@ states:
     acm:
       principals:       [everyone,  participant, moderator, creator, initiator, admin]
       permissions:
-        - [view,         D,         ~,           ~,          ~,      ~,         ~]
+        - [view,         D,         A,           A,         A,       A,         ~]
   evaluate:
     acm:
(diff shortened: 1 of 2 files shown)


src/adhocracy_core/adhocracy_core/workflows/debate_private.yaml
```diff
@@ -9,7 +9,7 @@ states:
     acm:
       principals:       [everyone,  participant, moderator, creator, initiator, admin]
       permissions:
-        - [view,         D,         ~,           ~,          ~,      ~,         ~]
+        - [view,         D,         A,           A,         A,       A,         ~]
```
slomo: This should result in admins not still not beeing able to see those processes? Is that intended?

joka:
> @joka what do you think about the problem?

I don't get understand it yet, 

Setting Deny to everyone should work because participators, moderators, .. have an explicit Allow in the  inherited non private workflow.  
Pratically it does not work because the Allow is overwritten with None?

Can you extend the workflow tests to cover this?

joka:
Please add a regression test

PR #2246, opened 2016-04-20
Use local firefox with xvfb instead of sauce labs

slomo:
More effecting integration testing on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,14 @@ install:
   - "git fetch --unshallow"
   - "python bootstrap.py --setuptools-version=12.0.4"
   - "PATH=${TRAVIS_BUILD_DIR}/bin:${PATH} ./bin/buildout -c buildout-${BUILDOUT_TARGET}.cfg 2>/dev/null"
+before_script:
+  # setup xvfb for protractor tests
+  - "export DISPLAY=:99.0"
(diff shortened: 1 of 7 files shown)


xi:
I think there were reasons why we postponed updating protractor in the past. Do you know about them and if yes, what changed?

xi:
Could you give a brief overview of pros/cons? I guess this is faster (about 20%) and less vulnerable to network timeouts. Anything else?

slomo:
I dont' know why we didn't update protractor. It is though required for this pull-request because the old version seems to have some issues with the firefox driver and the chrome version on travis is currently to old to run with either the new or the old chrome driver.

slomo:
My main motivation was speed. It makes the test setup more similar to the local one and remove dependency to sauce labs as an external service (though increasing coupling with travis a bit more). A big downside is that we can't run tests with other browsers any more (which we weren't doing any more). An other downside is that this might bring us closer to the travis memory limit, which when violating it leads to hard to debug test aborts.

slomo:
I have spoken with @der-john, he didn't update protractor the last time because there were some tests failing afterwards. But this was either a protractor issue or unrelated because now everything keeps working.

slomo:
Rebased on to master an squashed all dummy commits. Some change here are conflicting with #2330.

xi:
Continued in #2353

PR #2247, opened 2016-04-20
2016 04 as add automatic workflow transition script

2e2a:
@joka please review


2e2a:
@xi, @der-john: This adds a new value to workflow_meta in the meta_api.

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
```diff
@@ -0,0 +1,105 @@
...
+def _workflow_auto_transition_needed(context: IResource, registry: Registry,
+                                     workflow_assignment: IWorkflowAssignment):
+    workflow = registry.content.get_workflow(context)
+    next_states = workflow.get_next_states(context,
+                                           create_fake_god_request(registry))
+    if len(next_states) == 1:
+        next_state = next_states[0]
+        end_date = _get_current_state_end_date(workflow_assignment, next_state)
+        if end_date:
+            return now() > end_date
```
pallix: it would be better to have this measured once when the script begin. Otherwise it could have some strange effects, for example workflow w1 and workflow w2 have and end date of t1 and t2 respectively. Now the script is executed at time t1+d where t1+d < t2. The following could occur: updating workflow w1, for any reason, take e time where t1+d+e > t2. When this check occurs, `now()` will be superior or equal to t1+d+e and t2 will also be updated, which breaks the principle of least astonishment: the user would have expected only t1 to be updated.

I know the example seems a bit contrived but I prefer to be on the safe side.

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
```diff
@@ -0,0 +1,105 @@
...
+    for state_data in state_data_list:
+        if state_data['name'] == current_state:
+            current_state_data = state_data
+        if state_data['name'] == next_state:
+            next_state_data = state_data
+    if current_state_data:
+        if next_state_data and \
+                current_state_data['end_date'] != \
+                next_state_data['start_date']:
+            return None
```
pallix: why does it return None when the end_date of the current state is not the same as the start_date of the current state? If the goal is to not trigger a switch to the next state because there is an inconsistency between the dates then the function name is misleading. Should the consistency of the date be enforced somewhere (workflow validation?) else instead of silently doing nothing here?

2e2a: I added a print() in a later commit, so its not completely silently.
Agreed that validating the WorkflowAssignment consistancy would be better, but not sure if there is a nice way to do this.

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
2e2a: Will fix it.

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
```diff
@@ -0,0 +1,109 @@
...
+from adhocracy_core.workflows import transition_to_states
+from adhocracy_core.utils import now
+
+
+logger = logging.getLogger(__name__)
+
+
+def auto_transition_process_workflow():  # pragma: no cover
+    """Automatically transition workflow states.
+
```
joka: a hint that the start_date is used for automatically transition would be nice

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
```diff
@@ -0,0 +1,109 @@
...
+                                  )
+    processes = catalogs.search(query).elements
+    for process in processes:
+        workflow_assignment = registry.content.get_sheet(process,
+                                                         IWorkflowAssignment)
+        if _workflow_auto_transition_enabled(registry, workflow_assignment) \
+            and _workflow_auto_transition_needed(process, registry,
+                                                 workflow_assignment):
+            _do_workflow_auto_transition(process, registry)
+
```
joka: Common styleguide hints::

For me the names are nicer to read/shorter without the "workflow" prefix, the context  is obivious. Also passing around the assingment sheet looks like needless performance optimization.

If statements should fit in one line and should check variables or very simple statements (easier to read and debug)

variable names  for booleans should be named postive: `has_xyz` or `is_xyz` (easier to read)

Functions should be short  and follow one abstraction level (at least for the first levels) (easier to read)

example refactoring first abstaction level (search/resources):
 
    ...
    has_auto_transition = auto_transition_enabled(process, registry)
    is_outdated = state_is_outdated(process, registry)
    if has_auto_transition and is_outdated:
    ...    do_auto_transition(process, registry)

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
```diff
@@ -0,0 +1,109 @@
...
+    env = bootstrap(args.ini_file)
+    _auto_transition_process_workflow(env['root'], env['registry'])
+    transaction.commit()
+    env['closer']()
+
+
+def _auto_transition_process_workflow(context: IResource, registry: Registry):
+    catalogs = find_service(context, 'catalogs')
+    query = search_query._replace(interfaces=IProcess,
+                                  resolve=True
```
joka: The resolve option return a list instead of a generator,  to decrease memory consumption we should always prefere to work with generators.

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
```diff
@@ -0,0 +1,109 @@
...
+        if _workflow_auto_transition_enabled(registry, workflow_assignment) \
+            and _workflow_auto_transition_needed(process, registry,
+                                                 workflow_assignment):
+            _do_workflow_auto_transition(process, registry)
+
+
+def _workflow_auto_transition_enabled(
+        registry: Registry, workflow_assignment: IWorkflowAssignment):
+    workflow_name = workflow_assignment.get()['workflow']
+    workflow_meta = registry.content.workflows_meta[workflow_name]
```
joka: Return value type missing in signature.
`workflow_assingment` has the type IResourceSheet  ( I know its a bit tricky) . The process resource has the marker interface IWorkflowAssingment to indicate that the WorkflowAssignemnt sheet is available.

src/adhocracy_core/adhocracy_core/scripts/test_auto_transition_workflow.py
```diff
@@ -0,0 +1,206 @@
...
+        registry.content.workflows_meta = {
+                'standard': {'auto_transition': False}}
+        mock_sheet.get.return_value = {
+            'workflow': 'standard',
+            'workflow_state': 'participate',
+            'state_data': [
+                {'name': 'participate',
+                 'description': '',
+                 'start_date': datetime(2016, 1, 1),
+                 'end_date': datetime(2016, 1, 2)}
```
joka: You only need to set the data that is needed for the test

src/adhocracy_core/adhocracy_core/scripts/test_auto_transition_workflow.py
```diff
@@ -0,0 +1,206 @@
...
+        assert not mock_transition_to_states.called
+
+    def test_ignore_processes_currently_active(self, context, registry,
+            mock_catalogs, search_result, mock_transition_to_states,
+            mock_sheet):
+        process = testing.DummyResource()
+        context['process'] = process
+        mock_catalogs.search.side_effect = [
+            search_result._replace(elements=[process]),
+            search_result]
```
joka: Why does the code  need two search results?

src/adhocracy_core/adhocracy_core/scripts/test_auto_transition_workflow.py
```diff
@@ -0,0 +1,206 @@
...
+        assert not mock_transition_to_states.called
+
+    def test_ignore_processes_without_auto_transition(self, context, registry,
+            mock_catalogs, search_result, mock_transition_to_states,
+            mock_sheet):
+        process = testing.DummyResource()
+        context['process'] = process
+        mock_catalogs.search.side_effect = [
+            search_result._replace(elements=[process]),
+            search_result]
```
joka: Why does the code need two search results?
 Add a line note

src/adhocracy_core/adhocracy_core/scripts/test_auto_transition_workflow.py
joka: some statements like creating the process, setup the search mock and setup the workflow meta is repeated in every test. You could move this to a test fixture.

joka:
Common git styleguide hints:

- use git rebase interactive  to cleanup you commit history (for larger pull requests very helpfull)

- use common keywords like "Fixup" 

http://adhocracy3.readthedocs.org/en/stable/coding_guides/git_guidelines.html

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
joka: good point

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
joka: or for an even nicer first level abstaction::

     processes = _get_processed_with_autotransition(root, registry)
     for process in procecces:
           is_outdated = state_is_outdated(process, registry)
           if is_outdated:
                    do_autotransition(process, registry)

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
```diff
@@ -0,0 +1,109 @@
...
+        if state_data['name'] == current_state:
+            current_state_data = state_data
+        if state_data['name'] == next_state:
+            next_state_data = state_data
+    if current_state_data:
+        if next_state_data and \
+                current_state_data['end_date'] != \
+                next_state_data['start_date']:
+            print('Conflicting workflow assignment: {}.'
+                  .format(workflow_assignment))
```
joka: we should use logger.info instad of print in scripts

src/adhocracy_core/adhocracy_core/scripts/auto_transition_process_workflow.py
joka: sorry for being picky, I just want to emphasise the sections about readability, variables and function in the refactoring guideline: http://adhocracy3.readthedocs.org/en/latest/coding_guides/refactor_guidelines.html#clean-code-short-summary

pallix:
why does this PR have the  'frontend' tag?

2e2a:
> why does this PR have the 'frontend' tag?
it affects the meta_api.   Not sure if this needs to be reflected in the frontend.

2e2a:
Comments are reflected in the last commit. Also some fixed were needed after local testing.

However pull request #2258 needs to be considered before merging.

2e2a:
Adjusted to removal of state_data end_date

PR #2251, opened 2016-04-21
Ensure websockets won't leak user information

slomo:


src/adhocracy_core/adhocracy_core/websockets/server.py
```diff
@@ -51,6 +51,13 @@ def subscribe(self, client: Hashable, resource: IResource) -> bool:
         if self.is_subscribed(client, resource):
             return False
         path = resource_path(resource)
+
+        # ensure that no private info leaks to the outside
+        if path.startswith('/principals/') or path.startswith('/catalogs/'):
+            logger.info('Unallowed subscribtion')
+            raise WebSocketError('forbidden', 'You are not allowed to '
+                                 'subscribe on {}'.format(path))
+
```
joka: we use colander vor client data validation,  so we should add a validator to 
adhocracy_core/websockets/schemas.ClientRequestSchema field 'resource' 

Apart from validating subscriptions I would go a bit deeper. 
websockets.client is responsible for notifing the websocket server. That means it is also responsible to filter what changelog event are send to the websocket server, and event regarding /principals/.. should not be sent at all to the outside world.


Style note: the comment is needless if you use good variables names, for example::

     forbidden_paths =  ('/principals', '/catalogs')
     if  path.starts_with(forbidden_paths):
     ...

test are missing

pallix: can we merge it nonethless as a quickfix and you implement the proper solution soon?

joka: > can we merge it nonethless as a quickfix and you implement the proper solution soon?

We already have a quick workaround (disable websockets), so we don't have to bypass our definition of done for this fix.

Do you have time to work on this or  should  I do it?

pallix: Yves told me there are some usability problem when websocket are disabled.

I would appreciate you work on it since you have a deeper understanding of the websocket server. Or we work together on it.

joka: > Yves told me there are some usability problem when websocket are disabled.

ok, but then please add a TODO refactoring  comment and exclude the lines from test coverage

> I would appreciate you work on it since you have a deeper understanding of the websocket server. Or we work together on it.

Ok i do it,  for my proposed workaround you don't need deeper understanding of the websocket server.  

As proper solution I propose we ditch the websocket server and add a ServerSendEvents endpoint  to the rest api (that does not bypass the our security mechanisms)

xi:
The frontend currently ignores error messages on websockets completely. This means that failing subscriptions will go undetected.

The only effect of this that I can think of is that frontend cache invalidation for principals is broken. But they do not change anyway, so that should not be a problem.

Still, I opened #2257 to track this issue.

pallix:
I'm rewriting this PR with a colander validator

2e2a:
Looks good. Will do further tests on staging.

PR #2252, opened 2016-04-21
Fix report comment overlay in S1

nidico:
The first commit fixes the comment overlay, the second commit adds missing routes to individual comments.


dwt:
Checked and works.

PR #2249, opened 2016-04-20
rm phaseHeaders, phase and tabs

der-john:
This fixes #2248 .

PR #2254, opened 2016-04-22
rm unnecessary context templates

xi:
building on #2249, this removes some now unnecessary context templates.


der-john:
looks good & works.

PR #2255, opened 2016-04-22
rm complicated layout

xi:
This complicated layout structure was introduced in #1077 to enable headers of flexible height (i.e. phase headers). Now that this is no longer a requirement we can go back to the older, simpler approach.

Note that I did not update the templates that are removed in #2254, so that should be merged first.


der-john:
looks good & works.

Issue #2256, opened 2016-04-25
Duplicate information in workflow-assignment state-data


2e2a:
We currently have start_date and end_date for each state in the WorflowAssignment state_data.
This is duplicate and we would need to check constancy, i.e. end of a state is equal to start of the next.

I would suggest to remove end_date.

This should be considered in the workflow auto-transition script.

joka:
thats right, adding the end_date was neeed by the frontend, but it doent make sense for the workflow state machine. 
@xi can we remove the end_date and make the frontend use the next state start_date instead?

xi:
Interestingly, I could not find any place where `end_date` is really used in the frontend. So yes, you can remove it. Still some places need to be adapted that you can find with a simple `git grep end_date`.

2e2a:
Fixed by #2258

PR #2258, opened 2016-04-25
Remove end_date from WorkflowAssignment state_data

2e2a:
Fixes issue #2256 

We have also to reflect this change in the bplan/imperia API documentation.
Why does bplan have an evaluate phase?


pallix:
@2e2a either it was simpler to reuse standard workflow or they have one (ask the PMs)

PR #2259, opened 2016-04-25
add basic admin interface (SDI)

joka:
#Backend Extensions:
--------------------------

- /manage shows basic SDI (contents and services tab) when loged in as god

Backend internal Changes:  
----------------------------------

- add request property 'user' ,  rm utils.get_user,

Backend API Changes:   Token Header and Cookie based authententication 
-----------------------------------------------------------------------------------------------

We cannot use the existing Token header authentication for the sdi, because its based on  ServerSide Rendering. Possible solution are:

1. addional server to run the sdi:

      ++ easy
      -- even more complicated deployment
      -- no SSO frontend and sdi

2. Cookie based authentication for frontend and sdi

     ++ better security possible , see https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#Local_Storage and http://www.stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage/
http://blog.ionic.io/angularjs-authentication/
     ++ SSO frontend and sdi
     -- difficult CORS setup
     -- frontend changes needed

3. Token Header and Cookie based authentication

    ++ easy
    ++ SSO frontend and sdi
    -- less security (more ways to attack) 

4. No SSO, TokenHeader for the REST API and cookie authentication for the SDI

CHANGED: Instead of 3. I choosed 4. in this pull request.  
SSO will be another pull request depending on  #2318 

TODO:

 - [x] cookies are marked as secure when login was over HTTPS (deployment allows only HTTPS access)
 - [x] cookies are http-only
 - [x] cookies are same-site (lets hope other browser will support it)
 - [x] SDI provides an additional login
 - [x] SDI login sets the cookies on `/manage` path
 - [X] CORS headers are absent from `/manage`
 - [x] SDI prevents click-jacking by using `X-FRAME-OPTIONS`
 - [x] no changes to the frontend are made
 - [x] create low-prio SSO pull-request (may need even some other changes first)

### DIFF
--- a/src/adhocracy_core/adhocracy_core/sdi/views/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/sdi/views/__init__.py
@@ -1,1 +1,1 @@
-"""Views for the sdi admint interface."""
+"""Views for the admin interface."""
--- a/src/adhocracy_core/adhocracy_core/sdi/views/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/sdi/views/__init__.py
@@ -1,0 +1,1 @@
+"""Views for the sdi admint interface."""
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Module.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Module.ts
@@ -20,6 +20,7 @@ export var register = (angular, config, metaApi) => {
         ])
         .config(["$httpProvider", ($httpProvider) => {
             $httpProvider.interceptors.push(["adhHttpBusy", (adhHttpBusy : AdhHttp.Busy) => adhHttpBusy.createInterceptor()]);
+            $httpProvider.defaults.withCredentials = true;  // allow cookie uthentication for SSO with admin interface
         }])
         .service("adhHttpBusy", ["$q", AdhHttp.Busy])
         .directive("adhHttpBusy", ["adhConfig", "adhHttpBusy", AdhHttp.busyDirective])
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Module.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Module.ts
@@ -20,7 +20,6 @@ export var register = (angular, config, metaApi) => {
         ])
         .config(["$httpProvider", ($httpProvider) => {
             $httpProvider.interceptors.push(["adhHttpBusy", (adhHttpBusy : AdhHttp.Busy) => adhHttpBusy.createInterceptor()]);
-            $httpProvider.defaults.withCredentials = true;  // allow cookie authentication for SingleSignOn with admin interface
         }])
         .service("adhHttpBusy", ["$q", AdhHttp.Busy])
         .directive("adhHttpBusy", ["adhConfig", "adhHttpBusy", AdhHttp.busyDirective])
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Module.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Module.ts
@@ -20,7 +20,7 @@ export var register = (angular, config, metaApi) => {
         ])
         .config(["$httpProvider", ($httpProvider) => {
             $httpProvider.interceptors.push(["adhHttpBusy", (adhHttpBusy : AdhHttp.Busy) => adhHttpBusy.createInterceptor()]);
-            $httpProvider.defaults.withCredentials = true;  // allow cookie uthentication for SSO with admin interface
+            $httpProvider.defaults.withCredentials = true;  // allow cookie authentication for SingleSignOn with admin interface
         }])
         .service("adhHttpBusy", ["$q", AdhHttp.Busy])
         .directive("adhHttpBusy", ["adhConfig", "adhHttpBusy", AdhHttp.busyDirective])
--- a/etc/coveragerc
+++ b/etc/coveragerc
@@ -14,6 +14,7 @@ omit =
     src/adhocracy_core/adhocracy_core/__init__.py
     src/adhocracy_core/adhocracy_core/scripts/import_resources.py
     src/adhocracy_core/adhocracy_core/scripts/import_local_roles.py
+    src/adhocracy_core/adhocracy_core/sdi/*
     src/adhocracy/adhocracy/scaffolds/__init__
     src/adhocracy_sample/adhocracy_sample/__init__.py
     src/adhocracy_frontend/adhocracy_frontend/__init__.py
--- a/src/adhocracy_core/adhocracy_core/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/__init__.py
@@ -1,5 +1,6 @@
 """Configure, add dependency packages/modules, start application."""
 from pyramid.config import Configurator
+from pyramid.interfaces import IAuthenticationPolicy
 from pyramid_zodbconn import get_connection
 from substanced.db import RootAdded
 from logging import getLogger
@@ -7,7 +8,6 @@ from logging import getLogger
 import transaction
 
 from adhocracy_core.authentication import TokenHeaderAuthenticationPolicy
-from adhocracy_core.authorization import RoleACLAuthorizationPolicy
 from adhocracy_core.resources.root import IRootPool
 from adhocracy_core.resources.principal import groups_and_roles_finder
 from adhocracy_core.auditing import set_auditlog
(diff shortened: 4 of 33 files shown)


xi:
to many abbreviations. what is SSO? what is SDI?

pallix:
@xi RTFM

Just joking. Single Sign On. SubstanceD Management Interface.

joka:
> Just joking. Single Sign On. SubstanceD Management Interface.

I`m also confused by my self, we should use one name for the sdi, currently we have:

    sdi (substanced management  interface): package name and substanced abbrevitation
    substanced: used in talks
    Admin interface: used in documentation
    Technical Adamin interface: used in documentation and talks do difference the yet to come Admin interface

src/adhocracy_core/adhocracy_core/sdi/views/__init__.py
```diff
@@ -0,0 +1 @@
+"""Views for the sdi admint interface."""
```
pallix: typo

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Module.ts
```diff
@@ -20,6 +20,7 @@ export var register = (angular, config, metaApi) => {
         ])
         .config(["$httpProvider", ($httpProvider) => {
             $httpProvider.interceptors.push(["adhHttpBusy", (adhHttpBusy : AdhHttp.Busy) => adhHttpBusy.createInterceptor()]);
+            $httpProvider.defaults.withCredentials = true;  // allow cookie uthentication for SSO with admin interface
```
pallix: typo in the comment

joka:
> Just joking. Single Sign On. SubstanceD Management Interface.
> I`m also confused by my self, we should use one name for the sdi, currently we have:
> 
> sdi (substanced management  interface): package name and substanced abbrevitation
> substanced: used in talks
> Admin interface: used in documentation
> Technical Adamin interface: used in documentation and talks do difference the yet to come Admin interface

What name should we use?  i propose `admin interface` (sdi)

joka:
@pallix  I don't know what is the best way to add secure flag to cookies if https is used.
We could check the frontend_url setting for https, but this is a bit fishy. Or we add 
a new config option, like "set_secure_cookie_flag"

pallix:
maybe the option is not preferable as we probably do not want to have https enabled without the secure cookie flag.

joka:
> maybe the option is not preferable as we probably do not want to have https enabled without the secure cookie flag.

yes, but what about developing?

joka:
@slomo  or @2e2a can you check if its safe to set authentication cookies?

joka:
> maybe the option is not preferable as we probably do not want to have https enabled without the secure cookie flag.
> yes, but what about developing?

we agreed to set the secure flag if https is used, adhocracy server must use https only anyway

joka:
Authentication cookies are only shown in log files if python stack traces are logged, i think its ok that way.

joka:
rebased master

joka:
> @slomo or @2e2a can you check if its safe to set authentication cookies?

huhu?

slomo:
Hi,

I am pretty sure that with out current CORS policy it is not a good idea to add an auth cookie. The problem is there is a lot of code in that pull-request and I am not fit to assess in which situation a cookie and with which core headers is used. So instead some general thoughts:

- Production doesn't need any cors headers, since everything runs on the same domain
- Production currently deletes all cookies from requests, since it has caching implications otherwise
- Admin interface shouldn't need cors headers in any situation, since it is not accessed from js
- Development requires cors headers to be present
- Never should cors origin be `*` on a request that uses cookies for auth
- Auth cookies should be delivered with `http-only`, `secure` and `samesite` flag. Otherwise we shouldn't use any non-restrictive cors headers (like `*`) on any location with the same domain.

I am not sure whether our SDK needs api access, what would require that we use CORS headers also in production.  And I assume that if `example.com` embeds `liqd.net` as `<iframe>` requests from that `iframe` have origin `liqd.net`, but @xi is the expert here.

We might also think about restricting embedabillity of the admin interface to prevent clickjacking by using `X-Frame-Options`.


PS: I am not to active on my github notifications, please ping me also on an other channel if you need a quick response.

xi:
That may come a bit late, but why exectly do we need single sign on? It seems to make everything much more complicated.

xi:
> I am not sure whether our SDK needs api access, what would require that we use CORS headers also in production.

SDK does not access the API. So yes, if backend and frontend are on the same origin, we may not need CORS headers. I guess the initial motivation was to allow third party frontends, but that is really not important right now.

joka:
regarding cors header:
if we don'T need

joka:
Regarding cores header:  
 
* risky to use 
* we should not set them, especially not to '*' without need.
* we are working on making developing work without cors header (refactoring taks: add /api prefix and remove   frontend server )

To make the changes in this pull request easier to understand:

requests to the api (`/api`):

- COOKIES are now enabled (withCredential) but cookies are not sent because SDI cookies have `path=/manage`. If you change the path somehow the cookies are send but not used for authentication. 
- CORS header are set to  request.origin or '*',  before always '*'

request to login (`/api/login_username`):

- response includes the authentication cookies for the sdi (thats why we need to enable cookies)

request to the sdi (`/manage`) : 

- COOKIES are enabled and used for authentication
- CORS header are not set
- CSRF token are checked for POST requests.

COOKIE  secure flags:

    - httpOnly always set
    - secure set if https is used  (this is mandatory for adhocracy3)
    - same site  not set (I did not know about it )

joka:
Sidenote: the cores header are also causing preflight request for all URLs used by the frontend, thats pretty anoying.

joka:
> We might also think about restricting embedabillity of the admin interface to prevent clickjacking by using X-Frame-Options.

Yes we should do this.

joka:
samesite cooky flag  is a great way to block all CSRF attemps,  but is only supported by Chromium 51 so far.

slomo:
> Sidenote: the CORS header are also causing preflight requests for all URLs used by the frontend, thats pretty annoying.

Preflight request will only happen in development setup (But maybe that is what you meant by CORS enabled).

slomo:
Also to second @xi question again. Why do we need a single sign-on with the admin user interface?

joka:
> Also to second @xi question again. Why do we need a single sign-on with the admin user interface?

You and the POs  told me its a NiceToHave for this story. 
The missing part is how you the PO should find the Admin interface after loggin in with the frontend.
We can also don't do it, I don't have a strong opinion in this.

The Problem  CORS '*'  with cookies for the same domain (but different path) stays the same with or without singel sing-on. Is it save to add cookie authentication for the admin inteface or should we remove the CORS header at least for production installations?
Do we want to remove the CORS header in generall for security/preflight request reasons?

> Sidenote: the CORS header are also causing preflight requests for all URLs used by the frontend, thats pretty annoying.
>> Preflight request will only happen in development setup (But maybe that is what you meant by CORS enabled).

yes

slomo:
Open questions from my side:

 * Why do we need to send cookies on the XHRRequests?
   -  the frontend in your setup would never authenticate with the cookie
   - I am not sure, but assume that cookies set on XHRResponses will be used by the browser
   - the cookie needs only to be valid on the `/manage` path 

* Couldn't we split SSO into an independent pull-request?
  -  we really would like to see the SDI working ASAP
  - maybe the frontend should anyway be switches to cookie auth (See #2299) (security requirements would be the same)
  - I guess the SDI also offers login functionality, so that could be used for now (setting cookies only for /manage, which would not have CORS enabled)

joka:
* > Why do we need to send cookies on the XHRRequests?
     * > the frontend in your setup would never authenticate with the cookie
     * > I am not sure, but assume that cookies set on XHRResponses will be used by the browser
     * > the cookie needs only to be valid on the /manage path

Only to enable SS0, without the browser does not creates the authentication cookies after the  login request.

* > Couldn't we split SSO into an independent pull-request?
  * > we really would like to see the SDI working ASAP
   * > maybe the frontend should anyway be switches to cookie auth (See #2299) (security requirements would be the same)
   * > I guess the SDI also offers login functionality, so that could be used for now (setting cookies only for /manage, which would not have CORS enabled)

 I can add a SDI login/logout page and haveing a separted ticket regarding Cookie Auth and SSO SDI sounds reasonable. 
But the basic security concerns stays the same. Is it safe to use cookies for the path /manage only and  unrestrictive CORS header for parth /api only but  with the same domain?

slomo:
Sounds great. I agreed with @xi yesterday that SSO is a really great feature, that we would like to discuss and improve in an extra pull-request. There quite some issues in the frontend authentication that have to be solved first. I will take an hour today in order to create a list of steps to use cookies in frontend authentication. Unfortunately this change and SSO cannot be our top priory right now, because all frontend devs are bound in refactorings or feature requests.

Concerning this pull request I think we might all agree on the following procedure:

 - [x] cookies are marked as secure when login was over HTTPS (deployment allows only HTTPS access)
 - [x] cookies are http-only
 - [x] cookies are same-site (lets hope other browser will support it)
 - [ ] SDI provides an additional login
 - [ ] SDI login sets the cookies on `/manage` path
 - [X] CORS headers are absent from `/manage`
 - [X] SDI prevents click-jacking by using `X-FRAME-OPTIONS`
 - [ ] no changes to the frontend are made
 - [ ] create low-prio SSO pull-request (may need even some other changes first)

If we can check all boxes on that list I would consider this secure enough, lets hope I didn't miss anything. @xi and @joka do you agree?

joka:
ok

slomo:
As far as I can see there are no more frontend related changes in this code. As soon as a backend dev agrees with what we have here, we can deploy this to a staging server and see how it works together with varnish and production environments.

pallix:
@joka so what the status of this PR regarding security? Can it be merged?

PR #2262, opened 2016-04-26
disable deleteToken on failed loadUser

xi:
This is a quick workaround for an embedding issue: If the embedding page redirects shortly after login, the session token might get deleted (non-deterministic). I am not sure how good this workaround is, and I intend to continue working on a proper fix.


2e2a:
Tested on staging

xi:
related to #2263

Issue #2263, opened 2016-04-26
mein.berlin.de: login token gets lost


der-john:
I do the following:
- navigate to mein.berlin.de/w in Firefox
- click on "Login" i.e. get to  mein.berlin.de/w/adh/#!login
- enter valid credentials and click on "Submit"

Then, **most of the time** Firefox does the following:
- it directs me to mein.berlin.de/w/adh/#!/
- then to mein.berlin.de/w/adh/#!r/organisation/kiezkasse
- (possibly) gives me a 403 error and 404 errors
- it directs me to mein.berlin.de

When I navigate to mein.berlin.de/w immediately, I am logged out.

xi:
I heard that this was also reproduced in safari, and I was also able to reproduce it in chrome, but only if the developer tools were not active.

xi:
I thinks this is caused by the following:

- mein.berlin.de redirects automatically when it detects a session in localStorage. It thereby aborts all requests from the adhocracy iframe.
- after receiving the session token and saving it to localStorage, the frontend makes a HEAD request to `/` to check whether the token is valid. If the request fails, it is concluded that the token is no longer valid, so it is removed.
- Finally, the user service requests the user resource. If this fails it is concluded that that the session is broken and it is removed. (this has been disabled in #2262 as a quickfix)

My guess is that one of these requests is aborted and the token removed when the embedding page redirects.

A fix could be to check for specific error codes instead of simply deleting the session token on any failing status.

joka:
Regarding the head request to check if the token is valid: 

I m working on a pull request to use JsonWebToken ( https://jwt.io/ ) token https://github.com/liqd/adhocracy3/tree/2016-04-jk-jwt-cookies2.  
The allows the frontend to check if the token is not expired ( https://github.com/auth0/angular-jwt ).

PR #2264, opened 2016-04-26
handle aborted requests in http/error.ts

xi:
AJAX requests can be aborted, either explicitly (by calling `request.abort()`), by a firewall or browser extension, or because the user navigated away. In this case the request data is `null` and status is 0 (even though I also saw -1 in the wild). [spec](https://www.w3.org/TR/XMLHttpRequest/#the-status-attribute).

Our error handling code expected `data` to have a specific form, so it broke when a request was aborted and data was consequently `null`. I fixed that.

This is somehow related to #2263 because I think it is caused by aborted requests. This does not fix that issue, though.


der-john:
looks good & improves error messaging. no additional tests failing.

PR #2266, opened 2016-04-26
use jwt token for authentication

joka:


joka:
sorry migration is missing

PR #2267, opened 2016-04-27
rm adhLastVersion

der-john:
this function is used nowhere so it should be removed.


xi:
This directive is mentioned in a comment in Http.ts . Could you please adapt that?

PR #2271, opened 2016-04-27
use JSONWebToken (JWT) for authentication

joka:
While doing reasearch how to improve security and using standards for authentication
I found JSONWebToken (RFC 7519). Its a standard for digitally signed (HMAC or public/private keys) authentication data and becomming widly used for OAuth2 authentication architectures. For more information: https://jwt.io/introduction/ .
                                    
Proposal:                           
                                    
 - use JWT to generate authentication tokens
                                    
Backend internal changes:           
                                    
- code simplification: no more session management, use existing libraries to mange tokens
- remove config option "validate_user_token" (hacky and not used) 
- slightly faster   (no databse read/write needed)            
- migration script to remove token manager storage
                                    
Backend API extensions:             
                                    
- client can check for expired token without a request (see for example: https://github.com/auth0/angular-jwt#checking-if-token-is-expired )
                                    
Risks:                              
                                    
- all old authentication tokens will be invalid
                                    
Possible further steps:             
                                    
- use Oauth2 bearer scheme header (instead of `X-User-Token: <token>`): `Authorization: bearer <token>`
- use additonal refresh token to automatically updated if authentication token expires (OAuth2)


joka:
@slomo @2e2a  mind that all old authentication token will be invalid when this is merged

joka:
maybe old token raise internal error, test missing

joka:
also what happens if the token is valid but the user is deleted

src/adhocracy_core/adhocracy_core/authentication/__init__.py
```diff
@@ -1,124 +1,42 @@
...
-    :param get_tokenmanager: callable that accepts `request` and returns
-                             :class:`adhocracy_core.interfaces.ITokenManager`.
-    :param hashalg: Any hash algorithm supported by :func: `hashlib.new`.
-                    This is used to create the authentication token.
-
-    This Policy implements :class:`pyramid.interfaces.IAuthentictionPolicy`
-    except that not authenticated user get the additional principal
-    `system.Anynymous`.
-    """
+    The following methods are extendend:
```
pallix: typo

pallix:
How does it play with #2259 which use two tokens (if I understand correctly) ?

joka:
> How does it play with #2259 which use two tokens (if I understand correctly) ?

with #2259  /login_name  returns both, the token header for rest api and the cookies for substanced admin interface authentication. This pull request only changes how token are generated, so there is no conflict.

joka:
We might also consider using the  jwt JIT feature (readd server server session managment ;) )
to fight against replay attacks:

https://www.owasp.org/index.php/REST_Security_Cheat_Sheet  "Protect Session State"
http://security.stackexchange.com/questions/64541/can-i-prevent-a-replay-attack-of-my-signed-jwts
https://tools.ietf.org/html/rfc7519#page-10

joka:
I`m positive about  using JWT because we start using standard solutions instead of reinventing the wheel and it might be handy for future stories.
I had a quick talk wie @ix and he is not, mostly because we loose server side session management and there might be new security straits. We came up with some security straits i ellaborated more:

1. token used for DOS attacks
2. token stolen form the local storage (according to WASP this is most likely and the reason the sugget cookiy auth https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#Local_Storage  ) 
3. replay attacks ( https://en.wikipedia.org/wiki/Replay_attack )
4. token created with stolen private key

Counter actions:

1 . use  token blacklist ( same as old solution)
2.  cooky based authentication / use use blacklist as workaround ( same as old solution)
3 ( same as old solution) ( we could use the jti claim to prevent this https://tools.ietf.org/html/rfc7519#page-10, needs more elaboration)
4.  use jti claim (server generated identifier)  to mark "our" tokens  (new strait, I dont know how likely this is)

slomo:
I don't get why we are doing this. The linked website mentions SSO and auth sharing with third partys as motivation behind JWT. Both of those use cases are currently not in our scope. But this is just my opinion.

We should make sure that the frontend can handle the stale sessions after transitions. Otherwise we will have a ton of user complaints.

joka:
> I don't get why we are doing this. The linked website mentions SSO and auth sharing with third partys as motivation behind JWT. Both of those use cases are currently not in our scope. But this is just my opinion.

At the moment we have one authentication method used for quite different use cases we never elaborated properly: SinglePage Application (with upcomming high security requirements), SSO with django apps (pcompass, meinberlin),  authentication for external api users (meinberlin bplan). In the  near future we als need Outh2 SS0 with Buergerkonto.  So JWT fits quite well in our use cases. But you are right, we have a working solution and there is no real need to change the implementation now.  The main motiviation for this pull request is to use standard libaries instead of custom solutions/remove code complexitiy.

joka:
> We should make sure that the frontend can handle the stale sessions after transitions. Otherwise we will have a ton of user complaints.

yes

joka:
> indeed, what does the frontend do when the token is invalid, is the user asked to login again?

It just displays the backend error:  "Expired or invalid session deleted"

pallix:
So what is the status of this PR?

joka:
> So what is the status of this PR?

The frontend should handle session timouts / invalid authentication in  a nice way to prevent user from complaining if we change the authentication scheme in the backend.
Apart from this i don't see any problem.

 We might consider doing the move to JWT together with #2318

pallix:
@xi do you want to had better feedback or can we merge like this?

xi:
I do not really care how we generate our tokens. But there are some limitations:

- I do not see that we will integrate crypto in the frontend to be able to validate the token or to access encrypted information from it (for performance reasons).
- IMHO we need to be able to invalidate sessions on the server, so we still need to access the database on each request and cannot use the `exp` claim.

I also don't get the argument that this is more of a standard than what we did before. JWT is a standard about how to encode information in a token. This is not really what we want to do. Just generating a random token (what we did before) is probably part of dozens of standards and much more to the point.

So if using JWT libraries to reduce custom code in the backend is fine with me. But I do not want to use any JWT features. So IMHO we should not use JWT but instead a library/standard that actually does what we need.

About adapting the frontend to changes: We do not really have the resources to do this, as there are much more preccing issues like #2318.

joka:
> I do not really care how we generate our tokens. But there are some limitations:
> 
> I do not see that we will integrate crypto in the frontend to be able to validate the token or to access encrypted information from it (for performance reasons).

as i already explained:  there is no encrypted information und validation in the frontend makes no sense.
 
> IMHO we need to be able to invalidate sessions on the server, so we still need to access the database on each request and cannot use the exp claim.

We can use the exp claim to reduce head requests to the backend, but that can be another pull request.

> I also don't get the argument that this is more of a standard than what we did before. JWT is a standard about how to encode information in a token. This is not really what we want to do. Just generating a random token (what we did before) is probably part of dozens of standards and much more to the point.
> So if using JWT libraries to reduce custom code in the backend is fine with me. But I do not want to use any JWT features. So IMHO we should not use JWT but instead a library/standard that actually does what we need.

I don't understand this point and have the impression our communiction so far did not lead to a common understanding. We should talk face to face when we have time.

> About adapting the frontend to changes: We do not really have the resources to do this, as there are much more preccing issues like #2318.

With #2318 we have the same issue: all old authentication sessions will be invalid. 
So we have to decide if want to improve the fronten to improve user experience or not.

xi:
> validation in the frontend makes no sense

ok, I understand now that the validation is done on the server and the frontend justs trusts the data.

> We can use the exp claim to reduce head requests to the backend

#2318 clearly states that we need to provide server-side logout. So the session may have been deleted on the server before it expired. #2318 also states that in that case, also the cookie should be deleted, but I don't really trust that. I also don't see how refresh tokens help with this issue.

Anyway, there is only a single HEAD request for this when the app loads, so this is not really relevant for performance. The more important aspect is probably database access, which happens on every request. Since we want to provide server-side logout, this will not improve with JWT.

Speaking as a frontend developer, I don't mind how you generate the tokens on the backend. But speaking as a general a3 developer, I imagine it is very confusing for backend developers that we use a standard that does not fit our usecase at all and of which we only use a minimal subset. Is there realy no standard that just tells you how to generate secure session tokens?

xi:
I was thinking a bit more about refresh tokens. There may actually be a way to reduce database access with JWT, but as far as I can see that would not work with cookies. So I am still not sold on JWT.

joka:
> I was thinking a bit more about refresh tokens. There may actually be a way to reduce database access with JWT, but as far as I can see that would not work with cookies. So I am still not sold on JWT.

This also works with cookies, see linkend slides in #2318

joka:
> Speaking as a frontend developer, I don't mind how you generate the tokens on the backend. But speaking as a general a3 developer, I imagine it is very confusing for backend developers that we use a standard that does not fit our usecase at all and of which we only use a minimal subset. Is there realy no standard that just tells you how to generate secure session tokens?

I only know standards similar to JWT (passing signed identity information)  but more complex.

The main use case for JWT is passing identity information between client and Identity provider, like for Oauth2 authenticaton architectures.  You are right this is not our main use case (maybe in the future with Buergerkonto or other SSO solutions we want to offer).  

The only current advantage for the frontend is the possibility to use a standard way to pass the expiration date / user url. For the backend it allows a much simpler code base.

joka:
I will split this pull request in smaler ones

joka:
split pull requests https://github.com/liqd/adhocracy3/pull/2401 https://github.com/liqd/adhocracy3/pull/2402

joka:
all commit split  to other pull requests

PR #2277, opened 2016-04-28
activate pramid_execlog to log stacktrace of internal python errors

joka:
We don't log the full stack trace of python errors, only the short
descripton. With pyramid_execlog all stack traces are logged. This is
usefull for debugging and should be activated on production server if not done otherwise (@slomo)

Mind the stackttrace contains sensitive personal data like
authentication token or cookies.

### DIFF
--- a/etc/development.ini.in
+++ b/etc/development.ini.in
@@ -13,6 +13,7 @@ pyramid.debug_routematch = false
 pyramid.debug_templates = true
 pyramid.default_locale_name = en
 pyramid.includes =
+    pyramid_exclog
     pyramid_zodbconn
     pyramid_tm
     pyramid_mailer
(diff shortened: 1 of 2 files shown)


pallix:
are password excluded?

joka:
> are password excluded?

If  the login request causes an 500 error the password might be in the stack trace. 
If the pasword validation causes an 404 error, the password is removed from the log (see rest.exceptions)

pallix:
can't always we exclude passwords? Is it already like this for 500 or only on this PR?

joka:
> can't always we exclude passwords? 

Loggin stack traces means logging sensible data, I don't think its a good idea to filter it.
If the we realy think we should not log stack traces, we don't have to enable this on productive.

> Is it already like this for 500 errors or only on this PR?

Its already like this,  just up to  now we did not  log the full python stack traces caused by client reuqest.
This pretty annoying for debugging.

joka:
ping @slome @2e2a  you have to adapt the salt deployment.

slomo:
So should this now be enabled in production?

PR #2282, opened 2016-05-01
update buildout recipe dependencies

joka:
bowerrecipe:

	* update bower packages with every buildout run
          (should fix build problems with missing bower dependencies)

collective.recipe.template
gp.recipe.node
zc.rceipe.egg:

        * minor updates


xi:
LGTM.

PR #2283, opened 2016-05-02
Edit basic sheets and create Organisation/Processes with Substanced Admin Interface (sdi)

joka:
This pull request depends on #2259 !

SDI Extensions:
----------------------

If you go to the sid as god ( /manage ) you have to new features:

* edit basic sheets
* create processes and organisations

Backend internal Changes:
-----------------------------------

* Fix request attribute of IPermisison sheet might be none
* only store sheet appstruct data if values have changed
* allow to set workflow state to current state (does no trigger transition)
* preserve sheet order

API Changes:
------------------

* field name of IName sheets is required (post only)


src/adhocracy_core/adhocracy_core/sheets/__init__.py
```diff
@@ -192,7 +193,13 @@ def set(self,
...
             self.registry.notify(event)
         return bool(appstruct)
-        # TODO: only store struct if values have changed
+
+    def _filter_unchanged_data(self, new: dict, old: dict) -> dict:
+        changed = {}
+        for key, value in new.items():
+            if value != old[key]:
+                changed[key] = value
+        return changed
```
pallix: What about:

```
diff = set(new.items()) - set(old.items())
changed = dict((k, v) for k, v in diff)
return changed
```

?

src/adhocracy_core/adhocracy_core/sdi/views/sheets.py
```diff
@@ -0,0 +1,125 @@
...
+from substanced.util import _
+
+from adhocracy_core.interfaces import IResource
+from adhocracy_core.interfaces import IResourceSheet
+
+
+class FormView(SDFormView):
+    """View class that autogenerates forms based on :term:`schema` .
+
+    I uses :mod:`colander` schema to define the data structure and
```
pallix: Can we avoid the first person for documenting? It sounds strange.

src/adhocracy_core/adhocracy_core/rest/test_views.py
```diff
@@ -910,7 +910,7 @@ def make_one(self, request, context):
 
     def test_post_without_token_authentication_policy(self, request, context):
         inst = self.make_one(context, request)
-        with pytest.raises(IndexError):
```
pallix: Commit message is not related to the change?

pallix:
LGTM, modulo the comments above. Also it has already conflicts with master.

src/adhocracy_core/adhocracy_core/sheets/__init__.py
joka: yes thats nicer, but it does work with list values

src/adhocracy_core/adhocracy_core/sdi/views/contents.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def sdi_addable_content(self):
+        registry = self.request.registry
+        introspector = registry.introspector
+        cts = []
+        meta_addable = self.request.registry.content\
+            .get_resources_meta_addable(self.context, self.request)
+        content_types_addable = [m.iresource.__identifier__ for m
+                                 in meta_addable]
+        for data in introspector.get_category('substance d content types'):
```
pallix: what the purpose of getting all types of the system and then check if they are member of the addable set? Can't we just return the set?

pallix: Also why is the `is_sdi_addable` boolean not used to filter the addable resources?

src/adhocracy_core/adhocracy_core/sheets/__init__.py
pallix: you mean it does not work when the value is a list?

src/adhocracy_core/adhocracy_core/sdi/views/contents.py
pallix: @joka ping!

src/adhocracy_core/adhocracy_core/sheets/__init__.py
joka: > you mean it does not work when the value is a list?

yes, lists are not immuatable

src/adhocracy_core/adhocracy_core/sdi/views/contents.py
joka: > what the purpose of getting all types of the system and then check if they are member of the addable set? Can't we just return the set?

Looks like,  I have check this

joka: > Also why is the is_sdi_addable boolean not used to filter the addable resources?

I used the substanced filter mechanism, it only lists resource types that have an AddView for the current context. But you are right, we can also just filter by is_sdi_addable.

joka:
lgtm

PR #2286, opened 2016-05-03
Add a tsconfig for developer convenience

slomo:
- enables compilation via bin/tsc (without arguments)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,12 @@
+{
+  "compilerOptions": {
+    "module": "umd",
+    "removeComments": true,
+    "sourceMap": true
+  },
+  "files": [
+    "src/current/current/build/js/Adhocracy.ts",
+    "src/current/current/build/js/AdhocracySDK.ts",
+    "src/current/current/build/js/AdhocracySpec.ts"
+  ]
+}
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+  "compilerOptions": {
+    "module": "umd",
+    "removeComments": true,
+    "sourceMap": true
+  },
+  "files": [
+    "src/current/current/build/js/Adhocracy.ts",
+    "src/current/current/build/js/AdhocracySDK.ts",
+    "src/current/current/build/js/AdhocracySpec.ts",
+    "src/current/current/build/js/AdhocracySDKSpec.ts"
```
xi: This line is not required, because it is imported in AdhocracySpec.ts

xi:
Should we update any documentation?

der-john:
As far as I can see, there is no documentation of ours about this.

slomo:
I added something to the development config.

xi:
Sorry for nitpicking. I remember that `--watch` had some issues with symlinks in the past. We should check whether these still exist. I can do that later.

slomo:
I am using watch. It works even if I change stuff in core, only downside that it is horribly slow, but there are already tickets on the typescript repo for this.

xi:
Thanks for checking

PR #2287, opened 2016-05-03
Replace CrossWindow factory with a provider

slomo:
- remove injection filter if no trusted sites were present, because it
   was anyway broken
 - allow to register cross window message callbacks during configuration phase

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/CrossWindowMessaging/CrossWindowMessaging.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/CrossWindowMessaging/CrossWindowMessaging.ts
@@ -33,6 +33,8 @@
  * 10 in particular, but others may be affected.)
  */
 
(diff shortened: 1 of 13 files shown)


slomo:
This is a first step towards making the cross window messaging more extensible and removing all message handling from the service to make it more generic.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/CrossWindowMessaging/CrossWindowMessaging.ts
```diff
@@ -58,35 +58,79 @@ export interface IPostMessageService {
...
+                            adhUser,
+                            _self.callbacks);
+                    } else {
+                        return new Dummy();
+                    }
+                }];
+    }
+
+    public registerMessageHandler(name, callback) {
+        this.callbacks.push({ name : name, callback : callback });
```
xi: There should be no spaces before colons, because this is an object, not a type.

xi:
Nice!

PR #2288, opened 2016-05-03
simplify headers: a new header directive in TLS

### LINKED ISSUE #2213: mein.berlin.de: on clicking AddProposalButton, two login links are visible
See https://meinberlin-stage.liqd.net/w/prozesse/altstadt-2016/

**what I did**: Navigate to a process without logging in; click on "add proposal".
**wanted result**: I see https://meinberlin-stage.liqd.net/w/adh/#!login.
**what actually happens**: I see https://meinberlin-stage.liqd.net/w/prozesse/altstadt-2016/#!login - the "wrong login location". This way I actually see two login buttons: one in the CMS and one in a3.

We need to modify the behavior of the AddProposalButtons and/or the frontend routing. I'm not sure in which way.

der-john:
This doesn't work yet...

This fixes #2213


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
```diff
@@ -76,6 +76,7 @@ export class Provider {
     public areas : {[key : string]: any};
     public default : any;
     public $get;
+    public contextHeaders : {[key: string]: string};
```
xi: should be `key : string`

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/TopLevelState/TopLevelState.ts
```diff
@@ -95,6 +96,8 @@ export class Provider {
                     self, adhEventManagerClass, adhTracking, adhCredentials, $location, $rootScope, $q, $injector, $templateRequest);
             }
         ];
+
+        this.contextHeaders = {plain: "<adh-header></adh-header>"};
```
xi: not needed because this is the default anyway.

xi:
We refactored the code and merged in some parts from #2274

der-john:
@xi and I checked it and it works.

PR #2290, opened 2016-05-03
rm needless python module for every workflow definition

joka:


pallix:
Very nice, it makes so much sense.

Issue #2284, opened 2016-05-02
German translation for comment hint missing


2e2a:
In the comment field the text "Add comment here" is not translated.

xi:
Related to #2232. We need to update the german translation on meinberlin (special case because it uses "Sie" instead of "Du" [docs](https://adhocracy3.readthedocs.io/en/latest/frontend/translation.html#german-du-sie)).

PR #2292, opened 2016-05-04
translate comment hint in meinberlin

der-john:
this solves #2284


xi:
I also pushed to transifex

PR #2297, opened 2016-05-06
add idea collection process typ for meinberlin

joka:
"Idea collection" is the third standard process for mein besides bplan and collaborative text work.
Refactoring the backend and database migration to make full use of standard processe is yet to be done and not part of this pull request. 
You can import a test intstance with adhocray_meinberlin.workflow.test_idea_collection.resources.json

API Extensions:

   - new Process type Idea collection with content: IGeoProposal and standard workflow.

### DIFF
--- a/src/adhocracy_meinberlin/adhocracy_meinberlin/workflows/test_idea_collection.py
+++ b/src/adhocracy_meinberlin/adhocracy_meinberlin/workflows/test_idea_collection.py
@@ -13,7 +13,7 @@ def _post_document_item(app_user, path='') -> TestResponse:
 
 
 @mark.functional
-class TestCollaborativeText:
+class TestIdeaCollection:
 
     @fixture
     def process_url(self):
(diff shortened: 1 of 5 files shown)


src/adhocracy_meinberlin/adhocracy_meinberlin/workflows/test_idea_collection.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def _post_document_item(app_user, path='') -> TestResponse:
+    from adhocracy_core.resources.proposal import IGeoProposal
+    resp = app_user.post_resource(path, IGeoProposal, {})
+    return resp
+
+
+@mark.functional
+class TestCollaborativeText:
```
pallix: it should be TestIdeaCollection

joka:
> it should be TestIdeaCollection
>  Add a line note

copy and paste :)

joka:
We said we start new processes without versionable content. But I want to ask the frontend team first if its ok, they do a lot of refactoring to make one, reusable IdeaCollection Workbench.

xi:
> We said we start new processes without versionable content. 

We already share a lot of code between kiezkasse and buergerhaushalt and would like to also use that shared code for the new process. So when we switch to non-versionable, it should be at the same time for all three processes. We have already a lot of refactoring going on so I would prefer not doing this switch right now.

joka:
ok, then we add this to the refactoring list.

PR #2301, opened 2016-05-09
Introduce load_fixtures and provide them for euth

slomo:
This is a proposal to add fixtures for a3.


slomo:
I could add fixtures for other packages if this approach in general is acceptable.

joka:
I like the idea to have a central place for fixtures. 
 
For a general approach I don't like the complex buildout system as a hard dependency 
and the restriction to test fixtures only.
 
## Requirements
 
I would set the following requirements:
 
 - one place for resource fixtures (and application config)
 - support different environments: test, local, stage, productive.
 - extensionable
 - mind sensible data like email/passwords
 - part of the core packages, no hard dependency to specific deployment system like buildout
 
## Implementation
 
I propose to work with the existing pyramid features: ini files and PkgResources asset files
("my.package:config", "my.package:config.yaml", "/absulute/path/config/")
 
 
### a) Example Solution with Static Views
 
"Extensionable" we can get from overwriting pyramid static view directories (guarded by permission):
Extending instead of overwriting is also possible, but needs more research.
 
    http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html
    
File layout:

     src/adhocracy_core/fixtures/tests/import_resources_collaborative_text.test.json (static view 'fixtures'
     src/adhocracy_meinberlin/fixtures/tests/import_resources_tegel.test.json (overwrite 'fixtures')
     /absolute/path/fixtures/import_users.prod.json (overwrite 'fixtures')
 
Ini file:
 
    load_local_fixtures:'/absolute/path/fixtures/' (overwrite 'fixtures')
 
The import script bin/sd_load_fixtures can lookup all files in the 'fixture' static view directory and 
can use the file names to filter by import type. 
 
### B) Example solution with extensionable pyramid config (ini file):
 
Quick research:
 
    websauna (https://websauna.org/docs/api/websauna.utils.configincluder.html):
        - part of bigger framework
        - pyramid 1.6 support?
 
    pyramid_inheritet (https://pypi.python.org/pypi/pyramid_iniherit/0.1.8)
        - does not work with gunicorn WSGI server
        - pyramid 1.6 support?
 
    tzf.pyramid_yaml (http://tzfpyramid-yml.readthedocs.io/)
        - pyramid 1.6 support?
        + yaml files for application config
        + Default values
 
File layout with tzf.pyramid_yaml:

    src/adhocracy_core/fixtures/conf.yaml (general fixtures/config)
    src/adhocracy_core/fixtures/conf.test.yaml ('test' fixtures/config)
    src/adhocracy_meinberlin/fixtures/conf.test.yaml (extends 'test' fixtures/config)
    src/adhocracy_meinberlin/fixtures/conf.prod.yaml (extends 'prod' fixtures/config)
    /absolute/path/fixtures/conf.prod.yaml (extends 'prod' fixtures/config)
 
Ini file with tzf.pyramid_yaml:
 
    yaml_location = adhocracy_core:fixtures,adhocracy_meinberlin:fixtures,/absolute/path/fixtures
    env = prod
 
All data for the import scripts is in one file, conf.<env>.yaml. Optional we could also add all adhocracy specific config.  The import script bin/sd_load_fixtures can lookup the import data in the pyramid config.
 The import scripts need to be adapted to also work with python dicts/lists instead of file paths to json files.

slomo:
I think the requirements are odd to start with. Especially I don't get why:

- fixtures need to be available in prod or stage
- mind sensible data

The problem at hand is: It takes to long to get the development environment working. That is what I tried to fix. Not more.

I agree that is might make sense to store the fixtures into some well-defined location, in this pull-request I tried to reuse as much as possible and currency json data is all over the place.

As to integration with buildout. I am not a big fan of buildout, but we must accept that it is our buildsystem. To my knowledge it is the only place where we model package dependencies or functionality that changes depending on the package.

I can update this stuff to use fixed location for fixtures and maybe use pkg_resource loading. I think that your solutions are to complicated for the problem at hand and don't reuse the existing functionality of the import scripts, so I won't implement them.

joka:
> I think the requirements are odd to start with. Especially I don't get why:
> 
> fixtures need to be available in prod or stage
> mind sensible data

That is meant to have on place for fixtures, no matter if they are for testing or stage/productive installation.
At the moment I cannot test the production installation easily because I don't now what is  the configuration and what resources  are imported.

> The problem at hand is: It takes to long to get the development environment working. That is what I tried to fix. Not more.

Do you mean  the test environment to do manual testing?

> I agree that is might make sense to store the fixtures into some well-defined location, in this pull-request I tried to reuse as much as possible and currently the existing json data is all over the place.
> 
> As to integration with buildout. I am not a big fan of buildout, but we must accept that it is our buildsystem. To my knowledge it is the only place where we model package dependencies or functionality that changes depending on the package.
> 
> I can update this stuff to use fixed location for fixtures and maybe use pkg_resource loading. I think that your solutions are to complicated for the problem at hand and don't reuse the existing functionality of the import scripts, so I won't implement them.

ok

joka:
A simpler solution without using buildout would be using the exiting acceptance test setup scripts, like:  src/meinberlin/meinberlin/tests/acceptance/setup_tests.sh

slomo:
Rewrote it completely. @joka please check if that is okay, I will then add more fixtures, extend documentation and scaffolds

joka:
> Rewrote it completely. @joka please check if that is okay, I will then add more fixtures, extend documentation and scaffolds

I refactored your code (styleguide, tests) and made it more flexible (you can register arbitrary directories as fixture, or set a custom file system path.

still missing, check docu, code coverage, scaffold

joka:
> I refactored your code (styleguide, tests) and made it more flexible (you can register arbitrary directories as fixture, or set a custom file system path.

I did some changes with the intention to make this script also useable for productive installations.
(Don't do it, you might install test user or worse). I know we don't plan to use import scripts for deployment in the future, but there are users that like this, like me and usefull use  cases like setting app a demo application or complex initial content.

slomo:
Okay, than you should continue here because I don't won't have time any more to follow the adhocracy3 development that actively.

joka:
> Okay, than you should continue here because I don't won't have time any more to follow the adhocracy3 development that actively.

oki

Currently  if you run the script all fixtures area imported. Thats is not what i would except, also it might be dangerous for non development installations. I like to change this to:

- "default": list available fixtures
- --all: import all registered  fixture
- --custom import custom fixture or file system path

joka:
superseed by #2367

src/adhocracy_core/adhocracy_core/scripts/__init__.py
```diff
@@ -173,3 +179,68 @@ def get_sheet_field_for_partial(sheet: ISheet,
...
+        print('Import type "{}" file "{}"'.format(import_type, import_file))
+        if log_only:
+            continue
+        if import_type == 'groups':
+            _import_groups(root, registry, import_file)
+            # TODO don't use private functions
+        elif import_type == 'users':
+            _import_users(root, registry, import_file)
+        elif import_type == 'resources':
+            import_resources(root, registry, import_file)
```
2e2a: Fails with `TypeError: 'module' object is not callable` as the module `adhocracy_core.scripts.import_resources` is called and not the local function with the same name.

PR #2294, opened 2016-05-04
rename proposal in euth/pcompass

xi:
We want to join the euth, pcompass, kiezkasse and buergerhaushalt workbenches to a single, generic "idea collection" workbench in core. As a first step, this renames some things to avoid conflicts with future core directives (i.e. `adhProposal`).

I did not really test this because I do not have a ready-to-use test environment for these projects. @slomo @MagdaN could you please try?

PR #2308, opened 2016-05-10
mv meinberlin proposal

xi:
This moves the meinberlin proposal to the Meinberlin subpackage to avoid conflicts with a proposal package in core. So the itention is very similar to #2294.

Note that this will likely conflict with #2307. I am sorry for that.


der-john:
looks good & works.

PR #2320, opened 2016-05-12
Fix #2317  IGeoProposal creates IProposalVersion  as first version

joka:
*  add correct item_type IGeoProposalVersion to resource metadata of IGeoProposal

PR #2310, opened 2016-05-10
use IdeaCollection instead of Buergerhaushalt workbench / process

der-john:
this continues / contains #2307. It is still simply a reorganisation of the functionality of the Buergerhaushalt. The Kiezkasse functionality will be "merged" into the IdeaCollection soon.

PR #2322, opened 2016-05-12
implement idea collection process (aka Tegel)

der-john:
This has been made possible by #2307, #2308, #2310.

**Note**: This won't work until #2320 has been merged.

Refactorings remain to be done.

Also: Feature requests such as proposal moderation and functional comment deletion remain to be done.


der-john:
better to do another rebase.

PR #2324, opened 2016-05-12
make bower dependencies more stable 

joka:
* update bower to 1.7.9		

* Force "https://" instead of "git://" to checkout git repositories

      - some proxies allow only http/https connection scheme (like my mobile network)
      - seems to make less connection problems when running bower recipe

* use revision number to checkout  socialsherprivacy/merriweather to allow caching

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,4 +1,0 @@
-[url "https://"]
-    insteadOf = git://
-[url "https://github.com"]
-    insteadOf = git@github.com
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,4 @@
+[url "https://"]
+    insteadOf = git://
+[url "https://github.com"]
+    insteadOf = git@github.com
--- a/src/adhocracy_frontend/checkcode_and_compile.cfg
+++ b/src/adhocracy_frontend/checkcode_and_compile.cfg
@@ -147,7 +147,7 @@ inline =
         "devDependencies": {
             "typescript": "1.8.7",
             "tslint": "3.5.0",
-            "bower": "1.7.7",
+            "bower": "1.7.9",
             "jasmine": "2.4.1",
             "protractor": "2.2.0",
             "sync-exec": "0.6.2",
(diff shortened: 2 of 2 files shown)


joka:
failing acceptance tests

joka:
failing jsunit tests

slomo:
So what should we do about those failed test? Is this PR incomplete? Are those issues fixed on a new master version (looks like a backend issue to me)?

joka:
> So what should we do about those failed test? Is this PR incomplete? Are those issues fixed on a new master version (looks like a backend issue to me)?

I guess its not related to this pull request. I have  similar failing test with my local master branch. But testing locally is quite usleses, on my machine the tests are failing for months now. 
I'm working on making the acceptance test more stable: https://github.com/liqd/adhocracy3/tree/2016-05-jk-update-protractor. We can wait for this, then I can verify  that the tests dont fail.

xi:
The changes look good for me, espeically pinning commits is a great idea! I am not so sure about the git config, but I trust you if you say it is a good idea.

slomo:
It think that a file in $REPO/.gitconfig is never read by git. According to [git-scm.org](https://git-scm.com/docs/git-config) only $REPO/.git/config and ~/.gitconfig are read. So I would drop that commit.

der-john:
the protractor update has been merged. this PR works for me. @xi merge?

joka:
> It think that a file in $REPO/.gitconfig is never read by git. According to git-scm.org only $REPO/.git/config and ~/.gitconfig are read. So I would drop that commit. Instead users should add this to their global gitconfig where it would also fix this for all other github related projects.

True 
Nevertheless i would prefere a solution without manual steps. Maybe set the git config with buildout, like the pre commit hook?

xi:
I reverted 78486d5 and merged this otherwise uncontroversial pull request to get these improvements into master. @joka if you want to have a working solution for 78486d5, please open a separate pull request.

joka:
> I reverted 78486d5 and merged this otherwise uncontroversial pull request to get these improvements into master. @joka if you want to have a working solution for 78486d5, please open a separate pull request.

#2346

PR #2328, opened 2016-05-18
document noheader embed parameter

xi:
it was missing before.


der-john:
looks good & works. ;)

PR #2329, opened 2016-05-18
add report-abuse function in IdeaCollection

der-john:


der-john:
for layout reasons, this'll be part of a bigger PR.

Issue #2268, opened 2016-04-27
adhResourceAction.resourceWidgetDelete does not work


xi:
We currently have 2 places where the delete resource action is used:

-  mercator2015 proposal
-  mercator2016 proposal

Unfortunately, this action only works with resourceWidgets. We decided to deprecate resourceWidgets in #1711 and converted comments in #1743. mercator2016 was added in #1863 and has never been a resourceWidget. So basically deletion is broken with these two.

der-john:
this seems not to be up-to-date:
comment.ts uses adhHttp.hide(). I tried it out in a Kiezkasse and it worked fine.

der-john:
the directive has been renamed in #2428.

der-john:
it works fine in Mercator2015 and is not used in Mercator2016.

PR #2331, opened 2016-05-19
add functions to report + delete proposals in IdeaCollection

der-john:
this continues #2329.

- report is taken from `adhResourceAction`, as was the case for edit and print.
- the delete ResourceAction doesn't work for this (see #2268), so it's done as in `adhComment`.
- the template needed a little adjustment to the fact that edit, print and report are directives and delete isn't (see 		0ae6446).

I'm a little unsure about permission (checks):
- By default, users currently can't delete their own proposals. - I think this is fine.
- Everyone can now report proposals in Buergerhaushalt, ISEK and Kiezkasse processes. - I don't know the status of this. (Everyone can report comments, this has always been the case.)

### DIFF
--- a/src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/IdeaCollection.ts
+++ b/src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/IdeaCollection.ts
@@ -93,7 +93,11 @@ export var workbenchDirective = (
 
 export var proposalDetailColumnDirective = (
     adhConfig : AdhConfig.IService,
-    adhPermissions : AdhPermissions.Service
+    adhHttp : AdhHttp.Service<any>,
+    adhPermissions : AdhPermissions.Service,
+    adhTopLevelState : AdhTopLevelState.Service,
+    $location : angular.ILocationService,
+    $window : Window
 ) => {
     return {
         restrict: "E",
@@ -102,6 +106,22 @@ export var proposalDetailColumnDirective = (
         link: (scope, element, attrs, column : AdhMovingColumns.MovingColumnController) => {
             column.bindVariablesAndClear(scope, ["processUrl", "proposalUrl"]);
             adhPermissions.bindScope(scope, () => scope.proposalUrl && AdhUtil.parentPath(scope.proposalUrl), "proposalItemOptions");
+            adhPermissions.bindScope(scope, () => scope.proposalUrl && AdhUtil.parentPath(scope.proposalUrl), "versionOptions");
+            scope.delete = () => {
+                var proposalClass = RIGeoProposal;
+                if (scope.isKiezkasse) {
+                    proposalClass = RIKiezkasseProposal;
+                } else if (scope.isBuergerhaushalt) {
+                    proposalClass = RIBuergerhaushaltProposal;
+                }
+                // FIXME: translate
+                if ($window.confirm("Do you really want to delete this?")) {
+                    adhHttp.hide(AdhUtil.parentPath(scope.proposalUrl), proposalClass.content_type)
+                        .then(() => {
+                            adhTopLevelState.goToCameFrom("/");
+                        });
+                }
+            };
         }
     };
 };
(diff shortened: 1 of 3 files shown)


der-john:
I need to check about permissions to report comments.

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/Module.ts
```diff
@@ -42,7 +42,8 @@ export var register = (angular) => {
         .directive("adhMeinberlinIdeaCollectionWorkbench", [
             "adhTopLevelState", "adhConfig", "adhHttp", IdeaCollection.workbenchDirective])
         .directive("adhMeinberlinIdeaCollectionProposalDetailColumn", [
-            "adhConfig", "adhHttp", "adhPermissions", "adhTopLevelState", "$location", "$window", IdeaCollection.proposalDetailColumnDirective])
+            "adhConfig", "adhHttp", "adhPermissions", "adhTopLevelState", "$location", "$window",
+            IdeaCollection.proposalDetailColumnDirective])
```
xi: Either one line for all entries, or one line for each entry. Do not mix.

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/IdeaCollection.ts
```diff
@@ -102,6 +106,22 @@ export var proposalDetailColumnDirective = (
         link: (scope, element, attrs, column : AdhMovingColumns.MovingColumnController) => {
             column.bindVariablesAndClear(scope, ["processUrl", "proposalUrl"]);
             adhPermissions.bindScope(scope, () => scope.proposalUrl && AdhUtil.parentPath(scope.proposalUrl), "proposalItemOptions");
+            adhPermissions.bindScope(scope, () => scope.proposalUrl && AdhUtil.parentPath(scope.proposalUrl), "versionOptions");
```
xi: This should not be bound twice with different names.

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/IdeaCollection.ts
```diff
@@ -102,6 +106,22 @@ export var proposalDetailColumnDirective = (
...
+                    proposalClass = RIBuergerhaushaltProposal;
+                }
+                // FIXME: translate
+                if ($window.confirm("Do you really want to delete this?")) {
+                    adhHttp.hide(AdhUtil.parentPath(scope.proposalUrl), proposalClass.content_type)
+                        .then(() => {
+                            adhTopLevelState.goToCameFrom("/");
+                        });
+                }
+            };
```
xi: I am not sure whether this shouldn't be in the proposal directive. This would at least have been my initial choice. But maybe it is actually better to have it here because (1) you need to know the workbench context in order to know where to redirect after successful deletion and (2) this is not really specific to a proposal.

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/IdeaCollection.ts
```diff
@@ -102,6 +106,22 @@ export var proposalDetailColumnDirective = (
         link: (scope, element, attrs, column : AdhMovingColumns.MovingColumnController) => {
             column.bindVariablesAndClear(scope, ["processUrl", "proposalUrl"]);
             adhPermissions.bindScope(scope, () => scope.proposalUrl && AdhUtil.parentPath(scope.proposalUrl), "proposalItemOptions");
+            adhPermissions.bindScope(scope, () => scope.proposalUrl && AdhUtil.parentPath(scope.proposalUrl), "versionOptions");
+            scope.delete = () => {
```
xi: should be called `hide`

xi:
Works nicely! I added some fixups, but github does not update. Feel free to merge as soon as b72f445 has landed.

PR #2332, opened 2016-05-19
2016 05 ym fix jsunit tests

slomo:


der-john:
Nice! looks very good & works. @slomo: do you agree with my changes?

PR #2333, opened 2016-05-19
Disable travis deploy because not working or used

slomo:


2e2a:
There is duplicate after_success entry

PR #2335, opened 2016-05-19
add sort logic to IdeaCollection process

der-john:


slomo:
Works. A bit strange is that when I vote the sorting is update but the list item isn't (but this is unrelated).

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/Process/Detail.html
```diff
@@ -6,15 +6,17 @@
...
                     isBuergerhaushalt: isBuergerhaushalt,
                     isKiezkasse: isKiezkasse,
-                }">
+                }"
+                data-empty-text="{{ 'TR__PROPOSAL_EMPTY_TEXT' | translate }}"
+                data-ng-show="!isShowMap"
+                data-no-create-form="true"
+                data-path="{{path}}"
+                data-sort="sort"
+                data-sorts="sorts">
```
xi: These changes should also be applied to the map listing further down.

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/Process/Process.ts
```diff
@@ -34,11 +34,25 @@ export var detailDirective = (
         templateUrl: adhConfig.pkg_path + pkgLocation + "/Detail.html",
         scope: {
             path: "@",
+            sorts: "=?",
```
xi: You don't pass `sorts` as a parameter, so it should not be here.

xi: Fixed in #2341

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/Process/Detail.html
xi: Fixed in #2341

PR #2338, opened 2016-05-20
Adapt upstream sorting and facet refactoring.

dwt:
When the refactoring was done, it didn't include s1. Testing seems to prove, that just deleting code is all that is required to get up to the current standard. So here goes.


nidico:
LGTM

src/s1/s1/static/stylesheets/scss/components/_listing_theme.scss
```diff
@@ -1,3 +1 @@
-.listing .action-bar {
-    display: none;
-}
+
```
xi: If this file is empty it should be removed.

PR #2340, opened 2016-05-23
Simplify check code and compile

slomo:
- delete all symlinks in target dir upon start
 - fixes error with moved or delted files in any source dir

### DIFF
--- a/src/adhocracy_frontend/checkcode_and_compile.cfg
+++ b/src/adhocracy_frontend/checkcode_and_compile.cfg
@@ -240,30 +240,33 @@ static_directories =
     ${adhocracy:frontend.core.static_dir}
 inline =
     #!/usr/bin/env python
+    """ Merge static directories of multiple packages.
+
+        Assumes that this script is in full control of all symlinks under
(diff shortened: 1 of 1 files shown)


xi:
If we delete a symlink to a .ts file, is it possible to also delete the compiled .js file? I feel like the residual .js files tend to cause trouble.

slomo:
Well I can pull the atomic option and delete the entire target directory. It would also be even simpler.

slomo:
I reevaluated that, and it seems that I can't simply delete the target dir. Also it would be a bit strange to add a special case for a subset of js files (only those in /js) in this rather generic script. So how urgent is this issue?

xi:
I am not completely sure how urgent this is. In the past few weeks thetre were some issues after moving or deleting typescripot code. I suspect that this was caused by residual js files, but I did not really investigate.

slomo:
Maybe those are the same error, that are already fixed by this change: Errors with stale links, and therefore non readable typescript files.

xi:
LGMT

PR #2346, opened 2016-05-24
Force "https://" instead of "git://" to checkout git repositories

joka:
some proxies allow only http/https connection scheme
seems to make less connection problems when running bower recipe

### DIFF
--- a/src/adhocracy_frontend/checkcode_and_compile.cfg
+++ b/src/adhocracy_frontend/checkcode_and_compile.cfg
@@ -8,6 +8,7 @@ parts +=
     merge_static_directories
     do_merge_static_directories
     frontend.current.link
+    force_https
     bower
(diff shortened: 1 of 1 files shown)


slomo:
Will this also affect push urls?

joka:
> Will this also affect push urls?
no, there is a special option: url.<base>.pushInsteadOf

If you don't like the new setting you can manually modify .git/config. buildout will only change this file on the first buildout run.

xi:
> some proxies allow only http/https connection scheme

In that case the issue is not specific to this repository and you should adjust ~/.gitconfig

> seems to make less connection problems when running bower recipe

I would like to have stronger evidence before merging this because I think this code is hard to maintain (newcomers won't understand why it is there).

PR #2347, opened 2016-05-24
add SWOT labels to ideaCollection

der-john:
With this PR, filters concerning SWOT labels are enabled if `has-swot-labels` is set to true on registering the IdeaCollection.

We know this is not ideal. Within the next weeks, there'll be a PR that enables filters for each badge that exists in the backend.


src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/Process/Detail.html
```diff
@@ -47,6 +52,7 @@
                     data-sorts="sorts">
                     <adh-meinberlin-proposal-map-list-item
                         data-path="{{item}}"
+                        data-has-swot-labels="custom.hasSwotLabels"
```
xi: As far as I can see, the list item does not need to know about `hasSwotLabels`.

xi:
LGTM.

PR #2349, opened 2016-05-24
simplify map pins

xi:
This improves the map listing by simplifying the pins and adding clustering.


der-john:
Very nice! Looks good & works.

PR #2330, opened 2016-05-18
try to stablize acceptance tests

joka:
After month of failing tests they pass now on my local machine. Th

* refactor: use jasmin2 async feature for send emails acceptance test			
* remove test backend without websocket server, not used			
* fix: acceptance tests do not cleanup the database / use supervisor test group
* increase backend server workers to slightly improve accptance test speed			
* add mercator test fixtures script			
* refactor: prefere beforeAll instead of beforeEach in acceptance tests  
* update protractor from 2.2 to 3.3 (breaking changes)  
* fix regression acceptance test (workaround with wait timeouts)			
* gardening protractor conf			
* fix: use same timeouts for proctractorConfSauce and protactorConfSauce			
* add pageload timeout for SauceLab config to make sure angular is loaded (this makes tests run much slower but it might help to preventing test  failures)

The acceptance test  refactoring make senses in general but its also needed to make the test run on my local machine after upgrading to protractor.

PR #2350, opened 2016-05-25
Fix missing async handling of user page test

slomo:
Should be equivalent to https://github.com/liqd/adhocracy3/pull/2330/commits/d71db1b98fff9a9c3d7554bf1946b2817e266f50


slomo:
Should be equivalent to https://github.com/liqd/adhocracy3/pull/2330/commits/d71db1b98fff9a9c3d7554bf1946b2817e266f50

xi:
LGTM

PR #2351, opened 2016-05-25
Update protractor and refactor config

slomo:


slomo:
I would also recommend to change the sauce access key after merging this PR.

joka:
> I would also recommend to change the sauce access key after merging this PR.

why is this needed?

joka:
> why is this needed?

ah, secret

src/adhocracy_frontend/checkcode_and_compile.cfg
```diff
@@ -135,7 +135,7 @@ html_check =
 
 [nodejs]
 recipe = gp.recipe.node
-version = 4.2.4
+version = 4.4.5
 scripts = node npm
```
joka: wedriver suggests 4.2 or 5.4, why did you update to 4.4?
https://www.npmjs.com/package/selenium-webdriver

etc/protractorConfSauce.js
```diff
@@ -1,45 +1,20 @@
...
-                   .replace("%(here)s", process.cwd() + "/etc");
-        };
-
-        browser.params.mail = {
-            queue_path: getMailQueuePath()
-        }
-    },
     allScriptsTimeout: 21000,
     getPageTimeout: 20000,
     jasmineNodeOpts: {
```
joka: both timeouts seem a bit low
do we need this allScriptsTimeout at all?

joka:
all acceptance test are failing due to webdriver error

src/adhocracy_frontend/checkcode_and_compile.cfg
slomo: Because it is the current LTS version of node.

slomo:
> all acceptance test are failing due to webdriver error

fixed

etc/protractorConfSauce.js
slomo: Well those values are not determined in any way, I just copied them from the old config.

> both timeouts seem a bit low

They are high enough to get this working with travis reliably (as proven by this pull-requests travis run and others). As always with timeouts it is unclear what the reasoning should be here. If we are willing to extend them to what ever value just to get the tests somehow working, than we could directly configure them to INT.max. If we argue that in an acceptance test loading a page should at maximum take X seconds, otherwise this is no acceptable behaviour for the user than the values are probably already to high. I guess we all agree that inital page load time on almost any internet connection should not exceed 20 seconds. And also after 21 seconds the application should be in a state where everything is loaded.

Obviously those values also depend a bit on connection between the executing host and saucelabs.

> do we need this allScriptsTimeout at all?

I haven't tested it. But this is not completely unrelated to the other request. If the connection to the backend/frontend webservers is slow it will also take longer for the app to finish initial http requests (see [protractor-doc](https://github.com/angular/protractor/blob/master/docs/timeouts.md))

xi:
Looks good to me now.

src/adhocracy_frontend/checkcode_and_compile.cfg
joka: > Because it is the current LTS version of node.

i know, but to be on the safe side i would prefere  the  node version suggested by webdrive

joka: > i know, but to be on the safe side i would prefere the node version suggested by webdrive

Did you test run with this node version?

PR #2353, opened 2016-05-25
Use local firefox with xvfb instead of sauce labs on travis

xi:
*continued from #2246*

By using a local firefox instead of saucelabs the tests can run faster and more stable with regard to network issues.

On the downside, we loose the possibility to run the tests on a lot of different browsers easily. But we did not do that and it would probably require a paid plan on saucelabs, so that is only a theoretical downside.


xi:
For reference: With this, tests finish in ~15min (~25min with saucelabs)

joka:
At least two of 3 builds pass.
We have no proof if this makes acceptances tests more stable or not, but we should give it a try.

joka:
If we use firefox directl we don'T need to setup the webdrive server with `webdriver-manager update` anymore, or?  This command fails from time to time and takes quite some time.

xi:
> If we use firefox directl we don'T need to setup the webdrive server with webdriver-manager update anymore, or?

Why do you think that? We are still using webdriver, so I see no reason to remove that command.

joka:
> Why do you think that? We are still using webdriver, so I see no reason to remove that command.

as I understand it we use the webdrive api but not the wedriver server. Instead connect directly to browser drivers: http://www.protractortest.org/#/server-setup

joka:
`webdriver-manager update`  will install the selenium server and ChromeDriver.

xi:
The ChromeDriver is still required, right?

joka:
> The ChromeDriver is still required, right?

This is the ChromDriver plugin for the selenium server, so we don'T need it as long as we don't connect to the selenium server.

joka:
the move to directConnect instead of Selenium proxy server is not not part of this pull request.
So removing the Selenium server installation can be part of a different pull request.

LGTM

PR #2354, opened 2016-05-25
Update sauce labs access key

2e2a:


2e2a:
Not needed, as config is done via web interface.

joka:
Is this neede or not?

2e2a:
> Is this neede or not?

We can merge this.
However there is a simpler way to do this in the future via the travis web interface.

PR #2359, opened 2016-05-25
try to make acceptance tests faster and more stable

joka:
* refactor: simplify send message acceptance test		
* refactor: prefere beforeAll instead of beforeEach in acceptance tests 
     Page loads like for Login/Logout in tests need a lot of time and may cause timeout
    problems (angular app is not fully loaded but protractor thinks otherwise)
   With using BeforAll instead of beforeEach we can redduce the nummer of page loads .

* stablize CommentsPage acceptance test helper			
* increase backend server workers to slightly improve accptance test speed

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/CommentSpec.js
+++ b/src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/CommentSpec.js
@@ -2,10 +2,11 @@
 
 var shared = require("./shared.js");
 var EmbeddedCommentsPage = require("./EmbeddedCommentsPage.js");
-
+var UserPages = require("./UserPages.js");
 
 describe("comments", function() {
-    beforeEach(function() {
+
+    beforeAll(function() {
         shared.loginParticipant();
     });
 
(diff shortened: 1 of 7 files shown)


src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/CommentSpec.js
```diff
@@ -13,22 +13,22 @@ describe("comments", function() {
     it("can be created", function() {
         var page = new EmbeddedCommentsPage("c1").get();
         var comment = page.createComment("comment 1");
-        expect(comment.isPresent()).toBe(true);
+        expect(comment.isDisplayed()).toBe(true);
```
xi: Why is this better?

xi:
The test are still unstable, but this is definitely an improvement.

src/adhocracy_frontend/adhocracy_frontend/tests/acceptance/CommentSpec.js
joka: > Why is this better?

It was needed to make the test pass after making the backend server work asynchron (increase workers from 1 to  3). I did not have time to investigate this.

joka: > It was needed to make the test pass after making the backend server work asynchron (increase workers from 1 to 2). I did not have time to investigate this.

maybe it was just because i did not call the promise function for the wait condition

PR #2361, opened 2016-05-30
use local nodejs for tests

xi:
My local acceptance tests failed, that's how I noticed that we don't use the local nodejs when using polytester.

### DIFF
--- a/tests.yml
+++ b/tests.yml
@@ -3,6 +3,6 @@ pyunit:
 pyfunc:
     command: bin/py.test --timeout=60 -q -m"functional and not jasmine" src/adhocracy_*
 jsunit:
-    command: bin/jasmine JASMINE_CONFIG_PATH=etc/jasmine.json
+    command: bin/node bin/jasmine JASMINE_CONFIG_PATH=etc/jasmine.json
 acceptance:
-    command: bin/protractor etc/protractorConf.js
+    command: bin/node bin/protractor etc/protractorConf.js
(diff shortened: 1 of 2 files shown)


tests.yml
```diff
@@ -3,6 +3,6 @@ pyunit:
 pyfunc:
     command: bin/py.test --timeout=60 -q -m"functional and not jasmine" src/adhocracy_*
 jsunit:
-    command: bin/jasmine JASMINE_CONFIG_PATH=etc/jasmine.json
+    command: bin/node bin/jasmine JASMINE_CONFIG_PATH=etc/jasmine.json
 acceptance:
```
joka: where should the directory bin/node_bin come from?

joka:
ah thats why I have failing tests all the time.

joka:
Do we still need to execute `source source_env` after this fix?

xi:
I am not sure as I did not use source_env myself. But thanks for pointing out it existed. Now the question is: Should we use source_env or my version?

joka:
> I am not sure as I did not use source_env myself. But thanks for pointing out it existed. Now the question is: Should we use source_env or my version?

we should prefere you fix,  it does not rely on manual steps

xi:
Sre you sure that source_env is no longer required? It also sets `LD_LIBRARY_PATH`

joka:
> Sre you sure that source_env is no longer required? It also sets LD_LIBRARY_PATH

This is needed if you compile python with PIL (image processing) and libjpeg/libpng.
As long as we have this as an installation option we should keep it.

What ist about  `A3_ROOT` (see  src/adhocrac_frontend/base.cfg)

xi:
I think we can leave source_env as it is. It may be helpful in many situations, but with this change it is less required.

joka:
> I think we can leave source_env as it is. It may be helpful in many situations, but with this change it is less required.

you are right.

`A3_ROOT`  is only used here: src/mercator/mercator/tests/fixtures/fixturesMercatorProposals1.py is this still needed by the frontend ?

joka:
lgtm

PR #2365, opened 2016-05-31
physical deletion

### LINKED ISSUE #2336: Introduce real delete
Currently A3 does support two types of delete:

 * flag as deleted
 * flag as hidden

( see current [documentation](https://github.com/liqd/adhocracy3/blob/d86ca20492273028bee85f652edb52be8ef75303/src/adhocracy_core/docs/deletion.rst) on deletion)

This results in some unexpected behaviour when deleting some special resources, like badge assignments. Once User Marcella has delete an badge assignment, User Lisa can't recreate the same badge assignment. She can't assign the badge again, because it is already assigned. She also can't remove the delete flag from the old assignment because she is neither creator of this assignment nor admin.

**Proposal:**

Change the current "flag as delete" feature into a real delete, using the HTTP `delete` verb.

**Analysis:**

In the frontend this delete is only used in two places:

 - User, Moderator or Inititator deletes a badge assignments
    - mark as deleted functionality is not optimal, reason for the pull-request
 - Creator of a proposal deletes a blog entries in mercator
    - TODO: what happens here if change the semantic of delete

FYI:  There are also three occurrences of hide:
 - Initiator or Moderator clicks "delete" button on a comment
 - Intitator or Moderator click "delete" on a proposal in mercator 2015
 - Intitiator or Moderator clicks "delete" button on a document in Alexanderplatz

As we see the semantic holds so far.

**Questions:**

 - Do we want to allow users in certain conditions to recover deleted Material.
     * We don't want it in any because it might cause problems as with badge assignments
 - If so, could we also use hide in those conditions or do we need a third concept for real deletion?

xi:
* fixes parts of  #2336*

This is about replacing the current "flag as deleted" by a real delete. Hiding resources and the existing deleted flag should stay as it is.

TODO:

-  [x] update docs
-  [x] add 'delete' http method to REST API
-  [X] update frontend

TODO in upcomming pull request:

- [ ] add 'removed'/ 'removed_child' notification to the websocket api 
- [ ] remove 'delete' tag from metadata sheet / migration
- [ ] remove special 'hidden' field for metadatda sheet  in options requests 
-  [ ] backup 'deleted' resources for 2 weeks

### DIFF
--- a/src/adhocracy_core/docs/deletion.rst
+++ b/src/adhocracy_core/docs/deletion.rst
@@ -7,8 +7,8 @@ Anyone with the *delete_resource* permission can *delete* it by using
 the HTTP DELETE verb.  Deleted resources can not be recovered.
 
 Deleting an existing resource is only possible for updatable
-resources, i.e. *not* for Versions, as this would mess up the version
-DAG.
+resources like Simple, Pools, Items, i.e. *not* for Versions as this would mess up the version
+:term:`DAG` and *not* for AssetDownloads.
 
 The effect of deleting is as follows:
 
@@ -19,11 +19,12 @@ The effect of deleting is as follows:
 * If the frontend attempts to retrieve a *deleted* resource via
   GET, the backend responds with HTTP status code *404 Not Found*, just
   as if the resource had never existed.
-* *Deleted* resources may still be referenced from other
-  resources. If the frontend follows such references it must therefore
+* *Deleted* resources may not be referenced from other
+  resources. If the frontend follows an outdated references it must therefore
   be prepared to encounter *404 Not Found* responses and deal with them
   appropriately (e.g. by silently skipping them or by showing an
   explanation such as "Comment deleted").
+* *DELETE* http method is idempotent
 
 Hiding Resources
 ================
@@ -63,7 +64,7 @@ The effect of these flags is as follows:
   resources will be found regardless of the value of their *hidden*
   flag.  However, only those with *hide_resource* permission are ever
   able to view the contents of hidden resources.  It's also possible to
-  set *include=visible* to get only non-deleted and non-hidden
+  set *include=visible* to get only non-hidden
   resources, but it's not necessary since that is the default.
 * If the frontend attempts to retrieve a *hidden* resource via GET, the
   backend normally responds with HTTP status code *410 Gone*.
@@ -175,21 +176,25 @@ Both pools show up in the pool sheet::
     ...                        ['elements']))
     ['.../pool1/',.../pool2/'...
 
-Lets check whether we have the permission to delete or hide resources.
+Lets check whether we have the permission to delete resources.
 The person who has created a resource (creator role) has the right to delete
 it::
 
-    >>> resp = anonymous.get(document_item).json
-
     >>> resp = participant.options(document_item).json
-    >>> resp['PUT']['request_body']['data']['adhocracy_core.sheets.metadata.IMetadata']
-    {'deleted': [True, False]}
+    >>> 'DELETE' in resp
 
-But they cannot hide it -- that special right is reserved to managers::
+But they cannot hide it::
+
+    >>> 'adhocracy_core.sheets.metadata.IMetadata' not in pprint(resp['PUT']['request_body']['data'])
+    True
+
+-- that special right is reserved to managers::
 
     >>> resp = moderator.options(document_item).json
     >>> pprint(resp['PUT']['request_body']['data']['adhocracy_core.sheets.metadata.IMetadata'])
     {'deleted': [True, False], 'hidden': [True, False]}
+
+FIXME: remove deleted flag, not used anymore
 
 Note: normally the sheets listed in the OPTIONS response are just mapped to
 empty dictionaries, the contained fields are not listed. But IMetadata is a
@@ -198,6 +203,9 @@ Therefore, the presence of the 'deleted' and/or 'hidden' fields indicates
 that PUTting a new value for this field is allowed. Once more, the
 corresponding value is just a stub (the empty string) and doesn't have any
 meaning.
+
+FIXME: remove the special 'hidden' field for option requests, not need if
+deleted flag is removed
 
 Lets hide pool2::
 
@@ -224,7 +232,7 @@ Now we get an error message when trying to retrieve the pool2::
     >>> 'modification_date' in resp.json
     True
 
-Nested resources inherit the deleted/hidden flag from their ancestors. Hence
+Nested resources inherit the hidden flag from their ancestors. Hence
 the child of the pool2 is now hidden too::
 
     >>> resp = anonymous.get("/pool2/child")
@@ -273,3 +281,14 @@ resources::
 
     >>> 'http://localhost/principals/users/0000001/' in resp.json['updated_resources']['modified']
     True
+
+In the end we can cleanup with some real deletion::
+
+    >>> resp = participant.delete("/pool1")
+    >>> resp.status_code
+    201
+
+    >>> resp = anonymous.get("/pool1")
+    >>> resp.status_code
+    410
+
--- a/src/adhocracy_core/docs/deletion.rst
+++ b/src/adhocracy_core/docs/deletion.rst
@@ -1,27 +1,46 @@
 .. _api-deletion:
 
-Deleting and Hiding Resources
-=============================
+Deleting Resources
+==================
 
-Two boolean fields (flags) defined in the
-*adhocracy_core.sheets.metadata.IMetadata* sheet can be used to delete or
-hide resources: *deleted* and *hidden*. Both default to false. If this sheet
-is omitted when POSTing new resources, the default values are used.
+Anyone who can edit a resource (editor role) can *delete* it by using
+the HTTP DELETE verb.  Deleted resources can not be recovered.
+
+Deleting an existing resource is only possible for updatable
+resources, i.e. *not* for Versions, as this would mess up the version
+DAG.
+
+The effect of deleting is as follows:
+
+* All child/descendant resources (whose resource path includes the path
+  of the deleted ancestor as prefix) are also deleted.
+* Deleted resources are no longer listed in parent pools or search
+  queries.
+* If the frontend attempts to retrieve a *deleted* resource via
+  GET, the backend responds with HTTP status code *404 Not Found*, just
+  as if the resource had never existed.
+* *Deleted* resources may still be referenced from other
+  resources. If the frontend follows such references it must therefore
+  be prepared to encounter *404 Not Found* responses and deal with them
+  appropriately (e.g. by silently skipping them or by showing an
+  explanation such as "Comment deleted").
+
+Hiding Resources
+================
+
+Apart from physically deleting a resource, it can also be marked as
+"hidden" using a boolean field (flag) defined in the
+*adhocracy_core.sheets.metadata.IMetadata* sheet. It default to false.
+If this sheet is omitted when POSTing new resources, the default value
+is used.
 
 The usecase for deleting is that users want to withdraw some content.
 The usecase for hiding is that moderators want to hide unapproriate
-content.  In terms of implementation, they differ only by permissions.
+content.
 
-Deleting or hiding an existing resource is only possible for updatable
+Hiding an existing resource is only possible for updatable
 resources, i.e. *not* for Versions (which are immutable and hence don't
 allow PUT).
-
-Anyone who can edit a resource (editor role) can *delete* it by PUTting an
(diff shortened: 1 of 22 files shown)


xi:
We should also think about what to send over websocket on deletion.

joka:
> We should also think about what to send over websocket on deletion.

the notification is called: "removed", by the way, we have not notification for 'hidden' resources

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
-Deleting and Hiding Resources
-=============================
+Deleting Resources
+==================
+
+Anyone who can edit a resource (editor role) can *delete* it by using
+the HTTP DELETE verb.  Deleted resources can not be recovered.
+
+Deleting an existing resource is only possible for updatable
+resources, i.e. *not* for Versions, as this would mess up the version
```
joka: we use the verb "editable" in the documentation

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
+DAG.
+
```
joka: :term:`DAG` to reference the glossary

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
+* Deleted resources are no longer listed in parent pools or search
+  queries.
+* If the frontend attempts to retrieve a *deleted* resource via
+  GET, the backend responds with HTTP status code *404 Not Found*, just
+  as if the resource had never existed.
+* *Deleted* resources may still be referenced from other
+  resources. If the frontend follows such references it must therefore
+  be prepared to encounter *404 Not Found* responses and deal with them
+  appropriately (e.g. by silently skipping them or by showing an
+  explanation such as "Comment deleted").
```
joka: * Delete http verb should be idempotent

xi:
> > We should also think about what to send over websocket on deletion.
>
> the notification is called: "removed", by the way, we have not notification for 'hidden' resources

The documentation states that there are "modified" events in hide. I am not sure whether this also makes sense for deletion.

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
-      { "reason": "deleted|hidden|both",
+* A positive value of the *hidden* flags is inherited by
+  child/descendant resources (whose resource path includes the path of
+  the hidden ancestor as prefix). Hence a *hidden* resource is one that
+  has its own *hidden* flag set to true or that has an ancestor whose
+  *hidden* flag is true.
+* Normally, only resources that are not *hidden* are listed in parent
+  pools and search queries.
+* FIXME Not implemented yet, since the frontend doesn't yet need it: The
+  parameter *include=hidden* can be used to include hidden resources in
```
joka: this paragraph also mentions the old "delete" . 
We can remove this paragraph, we wont fix this fixme anyway not soon.

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
+  to change that and include backreferences to hidden resources. The
+  same restrictions apply, i.e. normal users can use this parameter to
+  find out whether hidden backreferences exist, but they won't be able
+  to see their contents. In any case the frontend should be prepared to
+  deal with *410 Gone* when following backreferences in the same way as
+  when following forward reference -- even if it didn't explicitly ask
+  to include them, they might show up due to caching.
 
 FIXME We should extend the Meta API to expose the distinction between
 references and backreferences to the frontend, currently only the backend
```
joka: cleanup not related to this pull request: can be removed to cleanup fixmes we won't fix or are part of the refactoring list

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
+  hidden (as that status is inherited). In that case, the person who
+  last modified the child resource likely has nothing to do with the
+  person who hid the ancestor resource.
+* *Hidden* resources may still be referenced from other resources. If
+  the frontend follows such references it must therefore be prepared to
+  encounter *410 Gone* responses and deal with them appropriately (e.g.
+  by silently skipping them or by showing an explanation such as
+  "Comment deleted").
 * FIXME Not implemented yet, since the frontend doesn't yet need it.
-  *Deleted* and *hidden* resources will normally not be shown in
```
joka: cleanup not related to this pull request: can be removed to cleanup fixmes we won't fix or are part of the refactoring list

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
-  pool. FIXME At same point in the future, we might want to change that and
+* This document is about deletion of resources (JSON documents).
+  Deletion of uploaded assets (images, PDFs etc.) is outside its current
+  scope.
+* Currently, the hidden status of resources isn't treated as special by
+  the Websocket server. So, if an resource is flagged as hidden, a
+  "modified" event is sent to subscribers of that resource and a
+  "modified_child" event is sent to subscribers of the parent pool.
+  FIXME At same point in the future, we might want to change that and
   send "removed"/"removed_item" messages instead.
```
joka: cleanup not related to this pull request: can be removed to cleanup fixmes we won't fix or are part of the refactoring list

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
-* This document is about deletion of resources (JSON documents). Deletion
-  of uploaded assets (images, PDFs etc.) is outside its current scope.
-* Currently, the deleted/hidden status of resources isn't treated as special
-  by the Websocket server. So, if an resource is flagged as deleted/hidden
-  (or undeleted etc.), a "modified" event is sent to subscribers of that
-  resource and a "modified_child" event is sent to subscribers of the parent
-  pool. FIXME At same point in the future, we might want to change that and
+* This document is about deletion of resources (JSON documents).
+  Deletion of uploaded assets (images, PDFs etc.) is outside its current
+  scope.
```
joka: We should elaborated more what can be deleted:

Simples,  Pools, Items (including all versions), but not single versions
Assets but not AssetDownloads

Deletion of RateItems is disabled in the default permissions setting 
Deletion of ICommentable may be disabled if the activitiy for this resource is highg (multiple comments,...)

joka:
We also need to update websockets.rst

joka:
> The documentation states that there are "modified" events in hide. I am not sure whether this also makes sense for deletion.

Where?  websockets.rst  points out that hidden resources are like unsubscribed resources for the client

src/adhocracy_core/docs/deletion.rst
xi: > this paragraph also mentions the old "delete" . 

where?

> We can remove this paragraph, we wont fix this fixme anyway not soon.

Can we do that later? Let's concentrate on physical deletion here.

xi:
> > The documentation states that there are "modified" events in hide. I am not sure whether this also makes sense for deletion.
>
>Where? websockets.rst points out that hidden resources are like unsubscribed resources for the client

In deletion.rst:

> Currently, the hidden status of resources isn't treated as special by
>  the Websocket server. So, if an resource is flagged as hidden, a
>  "modified" event is sent to subscribers of that resource and a
>  "modified_child" event is sent to subscribers of the parent pool.

joka:
> In deletion.rst:
> 
> Currently, the hidden status of resources isn't treated as special by
> the Websocket server. So, if an resource is flagged as hidden, a
> "modified" event is sent to subscribers of that resource and a
> "modified_child" event is sent to subscribers of the parent pool.

We should delete this paragraph, websockets.rst is the right place for this documentation

joka:
The documentaion in 

> In deletion.rst:
> 
> Currently, the hidden status of resources isn't treated as special by
> the Websocket server. So, if an resource is flagged as hidden, a
> "modified" event is sent to subscribers of that resource and a
> "modified_child" event is sent to subscribers of the parent pool.

In opposit to websockets.rst this is what the websocket server is actually doing.
Should we change this behaviour? Then hidden/deletion would be the same regarding the websocket notifications.

src/adhocracy_core/docs/deletion.rst
joka: > this paragraph also mentions the old "delete" .
> where?

here line 66:  set *include=visible* to get only non-deleted and non-hidden

joka:
> In deletion.rst:
> 
> Currently, the hidden status of resources isn't treated as special by
> the Websocket server. So, if an resource is flagged as hidden, a
> "modified" event is sent to subscribers of that resource and a
> "modified_child" event is sent to subscribers of the parent pool.

Instead of the modfied event the websockte server sends "removed"
To continue the current behaviour we should sent in addition a "removed_child" event to the subsribers of the parent pool

src/adhocracy_core/docs/deletion.rst
```diff
@@ -1,123 +1,123 @@
...
+
+* All child/descendant resources (whose resource path includes the path
+  of the deleted ancestor as prefix) are also deleted.
+* Deleted resources are no longer listed in parent pools or search
+  queries.
+* If the frontend attempts to retrieve a *deleted* resource via
+  GET, the backend responds with HTTP status code *404 Not Found*, just
+  as if the resource had never existed.
+* *Deleted* resources may still be referenced from other
+  resources. If the frontend follows such references it must therefore
```
joka: This does not work with real deletion. Tthe frontend might have an outdated reference to an deleted resource, but the backend itself will delete all references/backreferences to deleted resources.

joka:
Documention looks fine besides:

- some remaining notions of the old "delete" flag 
- functional test for the new delete http method

commit is comming soon

joka:
We only implement the deletion http method here. Furthe backend and API refactorings should go into a separated pull request: 

TODO in upcomming pull request:

- [ ] add 'removed'/ 'removed_child' notification to the websocket api 
- [ ] remove 'delete' tag from metadata sheet / migration
- [ ] remove special 'hidden' field for metadatda sheet  in options requests 
-  [ ] backup 'deleted' resources for 2 weeks

src/adhocracy_core/docs/websockets.rst
```diff
@@ -116,8 +116,8 @@ that has been subscribed.
 
         { "event": "removed", "resource": "RESOURCE_PATH" }
 
-    In practice this usually means that the resource has been marked as deleted
-    or hidden (see :doc:`deletion`).
+    In practice this usually means that the resource has been delted or marked
```
xi: typo: deleted

src/adhocracy_core/docs/deletion.rst
```diff
@@ -19,11 +19,12 @@ The effect of deleting is as follows:
 * If the frontend attempts to retrieve a *deleted* resource via
   GET, the backend responds with HTTP status code *404 Not Found*, just
   as if the resource had never existed.
-* *Deleted* resources may still be referenced from other
-  resources. If the frontend follows such references it must therefore
+* *Deleted* resources may not be referenced from other
+  resources. If the frontend follows an outdated references it must therefore
   be prepared to encounter *404 Not Found* responses and deal with them
   appropriately (e.g. by silently skipping them or by showing an
   explanation such as "Comment deleted").
```
xi: How does this work? Are all resources referencing a deleted resource also deleted? Or are the references removed?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Http.ts
```diff
@@ -267,15 +267,16 @@ export class Service<R extends ResourcesBase.IResource> {
...
     public delete(path : string, contentType : string, config : IHttpConfig = {}) : angular.IPromise<any> {
-        var obj = {
-            content_type: contentType,
-            data: {}
-        };
-        obj.data[SIMetadata.nick] = {
-            deleted: true
-        };
+        if (this.adhPreliminaryNames.isPreliminary(path)) {
+            throw "attempt to http-put preliminary path: " + path;
```
der-john: why does this constitute a an attempt to http-*put*, and not an attempt to http-*delete*?

der-john:
I tried deleting badge assignments. The backend successfully deleted the badges. But the frontend (namely, `transaction.commit`) still assumed that any http request would yield a resource.

@slomo and I impemented a quick fix that I'll push right away.

On the long run, a good fix is needed.

@joka: Can you add documentation on `delete` to the docs at REST-API #batch-requests?

der-john:
looks good & works.

der-john:
@xi Could you have a look at 0a60b9b ?

Issue #2343, opened 2016-05-23
Update/improve glossary


xi:
Our [glossary](http://adhocracy3.readthedocs.io/en/latest/glossary.html) is not up-to-date and not easy to understand. I think we could really benefit from improving it.

joka:
+1

xi:
In #2371 (after a long discussion) I added a separate "concepts" section that is targeted at a wider audience. The glossary is still referenced in backend docstrings.

I am still not convinced that the  backend glossary is very useful in its current form. We may want to remove (including all references in docstrings)

der-john:
Won't be extended.

PR #2371, opened 2016-06-01
retarget glossary at anyone instead of only developers

xi:
*related to #2343*

This is a first step to make the glossary in our documentation more useful to people other than developers. Until now it was used to define few very technical terms, such as "ACM" or "groupid".

My proposal is to instead use it to establish a common terminology that can be used to talk about adhocracy and online participation in general. The terminology should be useful for very different stakeholders, most importantly software developers,  project managers, and initiators, but to a lesser degree also users.

The definitions are not final. However, I would like to get this merged soon to give a framework for future inmprovements, also by non-technical contributors.

Here is what I did in detail:

-   remove some very technical terms
-   add some more general definitions
-   rephrase existing definitions


docs/glossary.rst
```diff
@@ -7,58 +7,73 @@ Glossary
...
-      If `resource sheet` field with backreferences sets a
-      :class:`adhocracy_core.schema.PostPool` field, the
-      referencing resources can only be postet at the :term:`post_pool`.
-      This assumes that a post_pool exists in the :term:`lineage` of the referenced
-      resources.
-      If a `resource sheet` field with references sets this, the
-      referenced resource type can only be posted to :term:`post_pool`.
+   resource
+      Anything that exists in adhocracy is a resource: A proposal, a
+      comment, but also users or individual rates.
```
der-john: Nice!

docs/glossary.rst
```diff
@@ -7,58 +7,73 @@ Glossary
...
-       or :term:`roleid`. It is provided by an :term:`authentication policy`.
-       For more information about the permission system read
-       :doc:`api/authentication_api`.
+   sheet
+      Sheets are the features of :term:`resources <resource>`. A
+      proposal may for example have the sheet
+      ``adhocracy_core.sheets.title.ITitle`` that allows it to have a
+      title and the sheet ``adhocracy_core.sheets.comment.ICommentable``
+      that allows it to be commented on. A resource is really not much
+      more than the sum of its sheets.
```
der-john: What is it more at all?

docs/glossary.rst
```diff
@@ -7,58 +7,73 @@ Glossary
...
-       Unique id of one permission :term:`role`: "role:<name>".
-
    local role
-       A :term:`role` mapped to a :term:`principal` within a local
-       context and all his children.
+   permission
+      What a user is permitted to do depends on their role.  The roles a
+      user has often depend on the context. Example: Amelia (user) may
+      be the creator (role) of one proposal (context) and therefore
+      permitted to edit it (permission).
```
der-john: Nice!

docs/glossary.rst
xi: path, content_type, maybe some other general fields.

slomo:
Either change file ending to `.txt` or use valid reStructured text.

slomo:
Is DAG really needed here?

joka:
Adding conceptual documentation for POs and Desginers is very needed. 
I tried to push this issue but did not succed, also we never succed in finding a common place for this kind of documentation. Adding this to the read the docs documentation makes sense.

Regarding the glossary: 

You must not  remove the technical terms, this is aimed for developeres and used all over the code in various docstirngs.

joka:
Mind we also have a kind of technical glossary here ("sheets", "resources" are explained: http://adhocracy3.readthedocs.io/en/latest/backend/overview.html 
http://adhocracy3.readthedocs.io/en/latest/frontend/overview.html#backend-api

May a better solution is to have a techical glossary and  an overview of concepts to understand and model Resources/Workflows

Also we should restructure the documentation to allow it to grow and to have a good approach for each target audience (for example administrators, core developer, extension package developers, Integrators/POs/Designers/Customers)

joka:
> May a better solution is to have a techical glossary and an overview of concepts to understand and model Resources/Workflows

We have a bit outdated plone page with this topic and Nadin (Designer) started a document for Customers (somewhere at an external web hoster). Both should be merged into the read the doc documentation.

docs/glossary.rst
joka: A resource is the sum of its sheets and the path.
The only general field "Content_type" would fit better in the  metadata sheet.
Apart from this we have different python APIs for Pools and Simples, but this is a drawback due to technical reasons.

docs/glossary.rst
```diff
@@ -7,58 +7,75 @@ Glossary
...
 
-   userid
-      The unique id for one userique id of one :term:`group`: "group:<name>".
+   backend
+   frontend
+      The backend is the part of the software that stores the data.  The
+      frontend on the other hand is in charge of showing the data to
+      users.  Having a clear separation between these makes development
+      simpler and theoretically allows to have more than one frontend,
+      e.g. a website and a mobile app.
```
joka: In fact we already have two frontends, mind the Substanced Admin interface

docs/glossary.rst
```diff
@@ -7,58 +7,75 @@ Glossary
...
+   workflow
+   phase
+      Participation :term:`processes <process>` typically have multiple
+      phases: For example, you may want to first publish an
+      announcement, then have the actual participation for some time,
+      and display the results once that is over. This is possible by
+      using workflows.  Workflows can be used with processes, but also
+      with any other kind of :term:`resource`. An important feature of
+      workflows is that you can change the :term:`permissions
+      <permission>` for each role based on the phase.
```
joka: The technical term is `state` . We can use `phase` instead because its more common  to use in our team. 
But then we cannot difference anymore between The visible `phase` in the frontend (phase widget) and the real worfklow state in the backend.

xi: The phase widget has been removed, so that issue is resolved.

joka:
LGTM if we readd the deleted terms

xi:
> Either change file ending to .txt or use valid reStructured text.

This is valid, but the `glossary` directive is a (valid) sphinx extensions. Github is right in not displaying it. But I agree that it makes editing unnecessarily complicated and we should stick to core rst syntax.

xi:
No one was really happy with putting this into the glossary, so I created a separate document instead. The following criticism was raised:

-   We need a technical glossary for API documentation
-   Using the `glossary` directive results in an [empty preview on github](<https://github.com/liqd/adhocracy3/blob/1356b31/docs/glossary.rst>)

Apart from this, there was also an argument that some of the concepts are not specific to adhocracy and should be documented separately. Still I think there are enough concepts that are specific to a3 to justify this PR even if we create additional external documentation.

docs/concepts.rst
```diff
@@ -0,0 +1,60 @@
...
+
+workflow and phases
+    Participation processes typically have multiple
+    phases: For example, you may want to first publish an
+    announcement, then have the actual participation for some time,
+    and display the results once that is over. This is possible by
+    using workflows.  Workflows can be used with processes, but also
+    with any other kind of resource. An important feature of
+    workflows is that you can change the permissions
+    for each role based on the phase.
```
joka: To ease referencing in other documents  we should:

- use headings
- separate multi explanations  (backend/frontend, core/customize)

joka: workflow are not specific to processes, we its intended and used for every resource type.
In https://cms.liqd.net/intranet/copy_of_projekte/adhocracy-3/grundkonzept-und-terminologie) we difference

    open workflow:  phases transitions allow circle
    closed workflow: phases transitions don't allow circles

Then we also use:
   
     process workflow: workflow for all process content
     content workflow: workflow for specific process content, phases may differ

joka: We should add  a note about tranistions: "When entering a new phase (transition) arbitrary action may happen (like send email, assign badges,..)"

xi: That is not essential right now. When this gets relevant, we can add it.

joka:
> That is not essential right now. When this gets relevant, we can add it.
>  Add a line note

ok, 
what about ease referencing with headings?

xi:
> what about ease referencing with headings?

Interestingly, there does not seem to be a simple way to reference definitions from a definition list in rst. Can you give a brief example how to reference a term based on heading?

joka:
Actually i was thinking about  using http://www.sphinx-doc.org/en/stable/markup/inline.html#role-doc
to reference sections, but this does not work. 

Sphinx suggest using `:ref :` (http://www.sphinx-doc.org/en/stable/markup/inline.html#role-ref) references:

   :ref:`glossary-resource-with-heading`

   :ref:`resource <glossary-resource-without-heading>`

and to add explict anchors:

    .. _glossary-resource-with-heading:

    resource
    --------
   
    bla blup blu

    .. _glossary-resource-without-heading:

    resource
             bla blup blu



Alternatively you can use external links `resource <concepts.html#resource>`__

xi:
I would like to avoid putting labels into this document to ease editing. Under this assumption, changing to headers does not seem to have any benefits.

joka:
The glossary would be more helpfull if we can link the terms. We could do this without anchors because sphinx automatically creates anchors for headings, but this brakes easily if you modify the heading.
So adding Headings or better headings and anchors would ease writing documentation.
Can you ask the glossary editors if this is realy such an big issue?

joka:
I added the question to the pad https://pad.liqd.net/p/GitHub-glossary-markdown_Workshop

xi:
@joka changed as discussed.

2e2a:
lgtm

Issue #2179, opened 2016-03-23
Refactor backend code issues


joka:
- [x] remove needless resources from database (performance)  #1896 and other middle
- [x] search implementation (performance) #1686 #1905  #1924 hard
- [x] working with workflow/permission/roles (development, DomainSpecificLanguage) partly #2132  #2175 hard
- [x] resource sheet implementation (development)  #2147 hard
- [x] colander schema only for validation (development) #2172 middle
- [x] view_config decorator for rest api views (development) #2176 middle
- [ ] config methods to add resource/sheet types (development) easy
- [ ] improve caching hits (performance) easy (make current concept work right) / middle (other caching concept)
- [x] update pyramid to 1.6 easy
- [x] update substanced 1.0 (rm _is_service_ flag, fix migrations) easy
- [ ] update colander to 1.2 (fix schemas with missing binding,  maybe remove needless workarounds)  easy
- [ ] prefix '/API' for all rest api resources (deployment) middle
- [ ] remove frontend_server (backend can do it the same way) (development, deployment) middle
- [ ] workflow per object instead per resource type (maintenance, development) middle
- [ ] remove find_service/add_service from pools, not used (development) easy
- [ ] remove sheet interfaces (development, maintenance (less migrations)) hard
- [ ] script to remove not used images (maintenance) easy
- [ ] cleanup TODO comments (development) middle
- [ ] use content_name instead of IResource to create resourcess (prepare more decoupling frontend/backend and backend internal) middle
- [ ] use sheet name instead of ISheet interface to create sheets (prepare more decoupling frontend/backend and backend internal) hard
 
- [ ] pyvenv instead of virtualenv (embrace standard way) (docs) easy
- [ ] build docs without warnings, add check to keep it that way (docs) middle
- [ ] nice python api documentation (docs) middle
- [ ] fix type annotations for functions / use python 3.5 typing module (docs) middle
- [ ] document all scripts / add help text output to documentation (maintenance, docs) easy
- [ ] narrative docs for backend development (docs) middle

joka:
initial refactor issue list for domain `backend`. Other possible domains: `deployment`, `frontend`, `api`, `docs`

PR #2372, opened 2016-06-01
fix sphinx issues

xi:
While working on #2371 I noticed that building sphinx produces an extreme amount of warnings and errors. So I tried to fix at least some of them.

To reproduce, run the follwing commands:

```
$ rm -r docs/build
$ bin/ad_build_doc
```

Many issues I fixed were related to the fact that reStructuredText requires whitespace around inline directives. So ``:term:`resource`s`` is invalid. Instead you can do ``:term:`resources <resource>` ``.

Another issue I found was broken indentation. Please stick to the 4-space indentation rule and note that lists must not be indented (indentation signifies a blockquote).

I did not fix any warnings or errors related to broken python references (and there is a lot of them!). There were also some issues I could not fix because I did not know how to do that without violating against the maximum line length (e.g. in the docstring of `adhocracy_core.interfaces.SearchQuery`).

### DIFF
--- a/docs/api/embed-api.rst
+++ b/docs/api/embed-api.rst
@@ -17,7 +17,7 @@ Host
 Widget
     A piece of adhocracy that can be embedded on its own.  Some
     functionality (e.g. registration) will require the user to switch
-    to the *platform*.  See also :doc:`../development/coding_guide/css_guidelines`.
+    to the *platform*.  See also :doc:`../development/coding_guides/css_guidelines`.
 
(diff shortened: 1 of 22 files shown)


der-john:
looks good & works.

joka:
great.

This  issue is also in the refactoring list: https://github.com/liqd/adhocracy3/issues/2179
Once sphinx pass without error, we can add a check to travis, to keep it that way

Issue #2374, opened 2016-06-02
Follow-up: Introduce DELETE


der-john:
DELETE has been implemented in #2365.

Some problems discussed in #2336 remain.

 *  [x] backup 'deleted' resources for 2 weeks
    i suggest extending the audit log and add a script to make it usable by sysadmins (@joka)
 *  [x] remove 'deleted' flag form metadata sheet (only used in Debate Workbench / `Document.ts`)
   as far as I can see. I suggest it be replaced by hide  ( @der-john)
 *  [x] check if delete permission is given
 *  [x] remove 'deleted field in IMetadata sheet in backend 
 *  ([ ] enable 'removed'/ 'removed_child' notification for the websocket api see #2700 )
 *  [x] remove special 'hidden' field for metadatda sheet in options requests (not needed anymore)
 *  [x] Fix delete in pools for versions (e.g. blog entries)

joka:
I merged the remaining todos from https://github.com/liqd/adhocracy3/pull/2365

xi:
I elevated this to high prio because of the backup feature which is critical

der-john:
I was wrong about Debate Workbench / Document.ts: there is no delete action relating to the backend there at all (only a frontend attribute called `deleted` that makes a paragraph not be posted in the next version of the document).

joka:
Removed high prio cause backup feature  is implemented

PR #2375, opened 2016-06-02
add DELETE permission in the frontend

der-john:
solves part of #2374.


xi:
Faling test is unrelated (#2379)

PR #2376, opened 2016-06-02
2016 05 mh fix top bar is missing

dwt:
Was too small and integrated in the wrong way. Should now be good enough to be (barely) usable again. Still needs to be made nicely responsive for mobile use.


der-john:
Hi Martin! I'm sorry for the confusing names that we used recently: CustomHeader is not actually a header - as ec9a9bb seems to assume -, but a slot with in a header where the "add proposal" button goes (I suggested renaming it in #2377, coincidentally).

dwt:
Hi,

> Hi Martin! I'm sorry for the confusing names that we used recently: CustomHeader is not actually a header - as ec9a9bb <https://github.com/liqd/adhocracy3/commit/ec9a9bb2aa84624b72d375b805214897963f55d4> seems to assume -, but a slot with in a header where the "add proposal" button goes (I suggested renaming it in #2377 <https://github.com/liqd/adhocracy3/pull/2377>, coincidentally).


:/ Well, ok I`ll change it on Thursday.

Incidentally: might this have something to do with the fact that the user list currently does`t render anymore? (In S1)

Best Regards,
Martin

dwt:
@der-john Can you please take another look?

nidico:
Well, it's not nice, but definitely an improvement.

PR #2377, opened 2016-06-02
renew 'customHeaders'

der-john:
This renews the refactoring achieved by #2113: The core of the issue is that @xi and I found out it was illogical to have `adhProcessProvider.templateFactories` and `adhResourceArea.customHeader`, which both customised something in a similar way depending only on the process type, but function differently and are at different places.

So here I suggest four things:

- to replace the customHeader logic with logic that's completely analogous to `adhProcessProvider.templateFactories`,
- call this logic `processButtonSlot`, because that's really the UI component that it regulates,
- (129bb56) do away with the `AlexanderplatzAddProposalButton` - it's identical to `IdeaCollectionAddProposalButton`,
- (0b7b101) move the `IdeaCollectionAddProposalButton` to Meinberlin/Proposal and simply call it `MeinberlinAddProposalButton`, since it's now the only one in meinberlin.


src/meinberlin/meinberlin/static/js/Packages/Meinberlin/Buergerhaushalt/Module.ts
```diff
@@ -32,9 +32,6 @@ export var register = (angular) => {
             var registerRoutes = AdhMeinberlinIdeaCollection.registerRoutesFactory(processType);
             registerRoutes(processType)(adhResourceAreaProvider);
             registerRoutes(processType, "buergerhaushalt")(adhResourceAreaProvider);
-
-            var customHeader = adhConfig.pkg_path + AdhMeinberlinIdeaCollection.pkgLocation + "/CustomHeader.html";
-            adhResourceAreaProvider.customHeader(processType, customHeader);
```
xi: adhConfig is no longer required

xi:
Nice! This sure is an improvement.

I have some comments:

-  The "process" in `adhProcessProvider.processButtonSlot` is redundant.
-  If the one is called "templateFactories", the other should be called `buttonSlotFactories`. (Maybe also remove the "slot"?)
-  `customHeader` had an extremely simple approach, while `templateFactories` has an extremely complicated approach. I had hoped we could meet somewhere in the middle. But that simplification can also be done in a separate pull request.

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/Buergerhaushalt/Module.ts
```diff
@@ -41,5 +41,8 @@ export var register = (angular) => {
                 return $q.when("<adh-meinberlin-idea-collection-workbench data-is-buergerhaushalt=\"true\">" +
                     "</adh-meinberlin-idea-collection-workbench>");
             }];
+            adhProcessProvider.processButtonSlots[processType] = ["$q", ($q: angular.IQService) => {
```
xi: Space before colon missing.

src/mercator/mercator/static/js/Packages/Mercator/2016/Proposal/Module.ts
```diff
@@ -65,6 +70,11 @@ export var register = (angular) => {
             adhEmbedProvider.registerDirective("mercator-2016-proposal-detail");
             adhEmbedProvider.registerDirective("mercator-2016-proposal-listitem");
         }])
+        .config(["adhProcessProvider", (adhProcessProvider: AdhProcess.Provider) => {
+            adhProcessProvider.processButtonSlots[processType] = ["$q", ($q: angular.IQService) => {
+                return $q.when("<adh-mercator-2015-add-proposal-button></adh-mercator-2015-add-proposal-button>");
+            }];
+        }])
```
xi: Why is this not in `Workbench/Module.ts`

der-john: Very good question.

der-john: I was wrong. I'll move it.

der-john:
this has been transformed into another PR.

PR #2380, opened 2016-06-06
fix comment navigation in mercator

### LINKED ISSUE #2205: Comment listing not updating on navigation
See https://advocate-europe.eu/en/idea-space#!/r/advocate-europe/2016/proposal_0000669/@comments:pitch

- **what I did**: Navigate to one proposal; Navigate to the comments of one section; Without closing the comment column, navigate to the comments of another section.
- **expected result**: I see the comments of the second section.
- **what actually happens**: I still see the comments of the first section.

I guess this happens bacause the comment listing does not watch its `path` parameter.

der-john:
This fixes #2205 in a surprisingly indirect way.

`AdhListing` didn't use, nor watch, the `refersTo` value of comments. This value is crucial in Mercator, since all comments on (any parts of) a proposal are stored in one `pool`, but their `refersTo` values indicate which subresource they belong to.

(When you first loaded a given subresource comment view, this wan't a problem:
In `src/mercator/mercator/static/js/Packages/Mercator/2016/Workbench/Workbench.ts`, in `registerRoutes.specific` assigned this same subresource as `commentableUrl`. But `registerRoutes` is not executed in the case of #2205.)

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
@@ -1,7 +1,4 @@
-import * as _ from "lodash";
-
 import * as AdhConfig from "../Config/Config";
-import * as AdhCredentials from "../User/Credentials";
 import * as AdhHttp from "../Http/Http";
 import * as AdhMovingColumns from "../MovingColumns/MovingColumns";
 import * as AdhPermissions from "../Permissions/Permissions";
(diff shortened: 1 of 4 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
```diff
@@ -357,75 +353,6 @@ export var adhCommentListing = (
...
-                        });
-                    }
-                },
-                (msg) => {
-                    console.log("Could not query given postPool");
-                }
-            ).then(adhDone);
-        }
-    };
-};
```
xi: This directive is actually used for acceptance tests and it is embedded on policycompass. So it is *not* dead code.

xi:
> `AdhListing` didn't use, nor watch, the `refersTo` value of comments

That is not completely right. `refersTo` is passed to the listing via the `params` parameter in [Comment.ts:103](https://github.com/liqd/adhocracy3/blob/a9f5f17/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts#L103). We should really not add comment-specific code to the listing. I guess this can be fixed by also watching `params`.

PR #2384, opened 2016-06-06
Use Markdown editor for Text field in SDI

joka:
The code for the markdown editor has a separted repository: https://github.com/liqd/deform_markdown
and is already released on pypi.


xi:
The editor looks nice! I particulary like that you have formatting in the input field. That's IMHO much better than a split input/preview.

2e2a:
lgtm

PR #2386, opened 2016-06-07
Fix resource path in auditlog

2e2a:


joka:
true, makes more sense
9 python tests are failing, is this related?

2e2a:
| 9 python tests are failing, is this related?

Tests for auditlog are passing. It looks like s1 tests are failing.

joka:
> Tests for auditlog are passing. It looks like s1 tests are failing.

do you haven an idea why we did not notice this?

2e2a:
| do you haven an idea why we did not notice this?

No, I noticed this the first time running the tests for this pull request.

PR #2389, opened 2016-06-08
renew 'customHeaders' (version 2)

der-john:
(Continues #2377.)

This renews the refactoring achieved by #2113: The core of the issue is that @xi and I found it was illogical to have `adhProcessProvider.templateFactories` and `adhResourceArea.customHeader`, which both customised something in a similar way depending only on the process type, but function differently and are at different places.

So here I suggest five things:

- to replace the `customHeader` logic with logic that's completely analogous to `adhProcessProvider.templateFactories`,
- *new*: simplify this logic,
- call this logic (*new*) `buttonSlot`, because that's the UI component that it regulates,
- (129bb56) do away with the `AlexanderplatzAddProposalButton` - it's identical to `IdeaCollectionAddProposalButton`,
- (0b7b101) move the `IdeaCollectionAddProposalButton` to Meinberlin/Proposal and simply call it `MeinberlinAddProposalButton`, since it's now the only one in meinberlin.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Process/Module.ts
```diff
@@ -13,6 +13,7 @@ export var register = (angular) => {
         .provider("adhProcess", AdhProcess.Provider)
         .directive("adhWorkflowSwitch", ["adhConfig", "adhHttp", "adhPermissions", "$window", AdhProcess.workflowSwitchDirective])
         .directive("adhProcessView", ["adhTopLevelState", "adhProcess", "$compile", AdhProcess.processViewDirective])
+        .directive("adhButtonSlot", ["adhTopLevelState", "adhProcess", "$compile", AdhProcess.processButtonSlot])
```
xi: Should be something like `adhProcessButton`.

xi:
Looks good, but I still have to properly test this.

Just one note: Could you apply the simplifications also to `templateFactories`?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Process/Module.ts
der-john: I was actually putting your suggestion from #2377 into place:
> The "process" in adhProcessProvider.processButtonSlot is redundant.

xi: this does not apply here: `processButtonSlot` was always used as a property of `adhProcessProvider`, so the "process" is redundant. But `adhButtonSlot` is used on its own.

xi:
There is a cirucular dependency between `AdhTopLevelState` and `AdhProcess` because `adh-process-button-slot` is used in `adh-default-header`. This problem is not introduced in this PR, though.

PR #2390, opened 2016-06-08
refactor hide / delete actions

der-john:
Strongly related to #2374 (follow-up: introduce `DELETE`), and thus to #2336 (physical deletion).

The original aim of this was:
- translate the popup saying 'Do you really want to delete this?' when you hit 'delete' on an IdeaCollection proposal.

While doing this, I noticed that hide/delete functions were wrapped in extremely similar ways (including the aforementioned question) in the following five places:
- IdeaCollection
- Alexanderplatz
- Blog
- Comment
- Mercator2015

@xi and I discussed the possibility to factor out this functionality. I suggested adding a 'resourceAction' - eventually this action should replace `adhDeleteAction`, since `HIDE` should be used where 'mark-as-deleted' used to be used.

In this PR, 
- a `resourceAction` called `adhHideAction` is implemented;

the following use cases are replaced with `adhHideAction`:
- IdeaCollection
- Alexanderplatz
- Blog (#2365 replaced 'mark-as-deleted' with physical deletion. The
use case for this was deleting badge assignments, not blog contents.
Therefore @2e2a and I suggest to replace the delete action in `Blog.ts` with hide, even though I couldn't test it right now.)

- Mercator2015 - because the functionality is part of a `ResourceWidget`.

TODO:
- [x] check for remaining use cases of 'mark-as-deleted' (-> none),
- [x] check for remaining use cases of `DELETE` (-> none),
- [x] check for remaining use cases of  `adhDeleteAction` (-> only Mercator2015).
- [x] check whether Mercator2015 can be changed affordably.


der-john:
rebased.

PR #2394, opened 2016-06-09
replace 'mark-as-deleted', hide and adhDeleteAction with adhHideAction

### LINKED ISSUE #2268: adhResourceAction.resourceWidgetDelete does not work

der-john:
(continues #2390.)

Strongly related to #2374 (follow-up: introduce `DELETE`), and thus to #2336 (physical deletion).

The original aim of this was:
- to translate the popup saying 'Do you really want to delete this?' when you hit 'delete' on an IdeaCollection proposal.

While doing this, I noticed that hide/delete functions were wrapped in extremely similar ways (including the aforementioned question) in the following five places:
- IdeaCollection
- Alexanderplatz
- Blog
- Comment
- Mercator2015

@xi and I discussed the possibility to factor out this functionality. I suggested adding a `resourceAction` to replace `adhDeleteAction`, since HIDE should be used where 'mark-as-deleted' used to be used.

In this PR,
- a resourceAction called `adhHideAction` is implemented;
- the resourceAction called `adhDeleteAction` is removed - this fixes #2268.

The following use cases are replaced with adhHideAction:
- IdeaCollection
- Alexanderplatz
- Mercator2015
- Blog (#2365 replaced 'mark-as-deleted' with physical deletion. The use case for this was deleting badge assignments, not blog contents. Therefore @2e2a and I suggest to replace the delete action in Blog.ts with hide, even though I couldn't test it right now.)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
```diff
@@ -62,24 +63,6 @@ export var shareActionDirective = () => {
...
-            class: "@"
-        },
-        link: (scope, element, attrs, column : AdhMovingColumns.MovingColumnController) => {
-            scope.delete = () => {
-                column.$broadcast("triggerDelete", scope.resourcePath);
-            };
-        }
-    };
-};
-
```
xi: Why did you remove this?

src/meinberlin/meinberlin/static/js/Packages/Meinberlin/IdeaCollection/IdeaCollection.ts
```diff
@@ -100,7 +100,8 @@ export var proposalDetailColumnDirective = (
     adhPermissions : AdhPermissions.Service,
     adhTopLevelState : AdhTopLevelState.Service,
     $location : angular.ILocationService,
-    $window : Window
+    $window : Window,
+    $translate
```
xi: Not needed here

src/mercator/mercator/static/js/Packages/Blog/Blog.ts
```diff
@@ -72,15 +72,16 @@ export var bindPath = (
...
     adhHttp : AdhHttp.Service<any>,
     adhPermissions : AdhPermissions.Service,
     adhPreliminaryNames : AdhPreliminaryNames.Service,
     adhShowError,
     adhSubmitIfValid,
-    adhUploadImage
+    adhUploadImage,
+    $q: angular.IQService,
+    $window: angular.IWindowService,
+    $translate
```
xi: angular services should be injected first

Also: `$translate` is not used.

xi:
You should not remove the delete functionality from Blog. adding hide is fine, but delete is required.

xi:
It is a bit hard to review this for me. Could you please rebase this into 2 pull requests: One adding `hideActionDirective` and another one implementing translation on top of that?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
der-john: Because it was only used for Mercator2015, where I replaced it with `adhHideAction`.

der-john:
No problem.

der-john:
I've split it up.

PR #2395, opened 2016-06-09
Remove user list button from header

dwt:
as that feature got deleted from adhocracy core in https://github.com/liqd/adhocracy3/pull/1967


nidico:
Ok, we can live without the user list at the moment.

PR #2398, opened 2016-06-09
add description sheet to user

joka:
This allows to use the description field as a free text field to
describe the user. 

This is needed by the s1 process but also useful in general (discussed with moritz and @nidico  https://pad.liqd.net/p/paas-2016-06-09 )


2e2a:
lgtm

PR #2399, opened 2016-06-09
make field sheets.image.IImageMetadata.mime_type readonly

joka:
API Change:

- make field sheets.image.IIMageMetadata.mime_type readonly

Backend Internal change:

- simplify mime type validation for uploaded assets

Editing this field is not needed (The mime type is given implicit by
uploaded file) and confusing when using auto generated form like for the
admin interface.


xi:
frontend looks fine

2e2a:
lgtm

PR #2403, opened 2016-06-10
gardening: remove empty file

dwt:


nidico:
This breaks the css build, as the removed file is still included in `src/s1/s1/static/stylesheets/scss/_theme.scss`.

nidico:
Fixed in 0e6193edcdaa232141615e3424fd358c92928229.

PR #2404, opened 2016-06-10
use "test_fixture" fixture  to create test  users 

joka:
Test users for each user role are now create with the "test_fixture" fixture.
To do it manually just run:  bin/ad_fixtures etc/development -a

* more obvious
* less code and better SRP

depends on pull request #2401


2e2a:
lgtm

xi:
aa477ff broke many user-related acceptance tests. @joka how can this be fixed?

PR #2405, opened 2016-06-11
Add CSS concept links

xi:
- pattern labs was discussed some time ago, but never added to the documentation
- SUIT is extremely similar to our current concept. I was thinking about switching over to that completely sometime in the future.


der-john:
nice.

Issue #2396, opened 2016-06-09
Can't upload PNG to (S1) user space


der-john:
@dwt and I found this issue when testing s1:

Logged in to the functioning s1 site, we tried to upload an ordinary PNG file.

The Ui didn't react and the browser threw the following error message:
> "Claimed MIME type is image/png but file content seems to be application/octet-stream"

dwt:
to reproduce:
* log in to s1
* go to your own users page
* click user image
* select image (png)
* try to submit

joka:
guess its similar to #2273, the system dependencies to install python magic are not fullfilled: https://github.com/ahupp/python-magic , is libmagic-dev installed?

dwt:
installing libmagic did fix this locally. However, uploading an image (jpg/png) still does`t seem to work.

To be more precise: uploading the image seems to work, but then downloading it again gives this error:

```
{
"errors": [
{
"location": "internal",
"name": "",
"description": "OSError: decoder jpeg not available; time: 2016-06-10 11:06:13,784"
}
],
"status": "error"
}
```

Maybe another dependency issue?

dwt:
My googling seems to indicate that my Pillow installation lacks jpeg / zlib support. I do have all the necessary libraries installed though. Could you perhaps advise how I can check pillow is really lacking links to those libraries?

dwt:
I just checked again by reinstalling pillow, and it supposedly has found the required dependencies:

```    --------------------------------------------------------------------
    PIL SETUP SUMMARY
    --------------------------------------------------------------------
    version      Pillow 3.2.0
    platform     darwin 3.5.1 (default, May  3 2016, 15:11:20)
                 [GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.29)]
    --------------------------------------------------------------------
    --- TKINTER support available
    --- JPEG support available
    *** OPENJPEG (JPEG2000) support not available
    --- ZLIB (PNG/ZIP) support available
    --- LIBTIFF support available
    --- FREETYPE2 support available
    *** LITTLECMS2 support not available
    *** WEBP support not available
    *** WEBPMUX support not available
    --------------------------------------------------------------------```´

dwt:
forget that, I was checking the wrong installation of pillow. :/

dwt:
Works for me now after reinstalling the missing libraries (lzlib, libjpqeg) and then actually reinstalling adhocracy.

joka:
[...]

Issue #2273, opened 2016-04-27
Image upload in Euth is broken


MagdaN:
BE Error 400: "Claimed MIME type is image/jpeg but file content seems to be application/octet-stream"

MagdaN:
Reset button also does not work (no error in chrome debugging console)

joka:
Can you add the image that is causing this error, or does this error always occure?

MagdaN:
The error always occurs

joka:
I cannot test it locally, build error: src/euth/euth/build/js/AdhocracySpec.ts(19,22): error TS2307: Cannot find module './Packages/LocalSocket/LocalSocketSpec'.
The acceptance test on master were passing yesterday,   maybe this is a deployment issue, is https://github.com/ahupp/python-magic isntalled and working properly (@slomo) ?

MagdaN:
The problem occurs on my local machine

joka:
> The problem occurs on my local machine
Your are using MacOs or? please check that the dependencies for python-magic are installed https://github.com/ahupp/python-magic and  that `import magic` works. 

We should  add installation instructions for MacOs to the docs.

joka:
Do the acceptance test pass on your machine?

MagdaN:
Installing libmagic fixed the problem

joka:
Can you check that the installation instructions cover this?

joka:
Also if libmagic is a hard dependency the backend should raise an exception

slomo:
I added a pull-request for please review it.

PR #2407, opened 2016-06-11
raise import error if python-magic is not properly installed

joka:
python-magic is hard dependency to make image upload work.
Checking this sould make error issues like #2396, #2273 less likely.


2e2a:
lgtm

PR #2408, opened 2016-06-11
update Pillow from 2.7 to 3.2

joka:
Among other improvement zlib and libjpeg are now required (https://github.com/python-pillow/Pillow/issues/1412 ) ([changelog](https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst)).
This should fix issues like https://github.com/liqd/adhocracy3/issues/2396 .


xi:
Seems to work fine. There are 9 failing, but they are also failing in master.

PR #2409, opened 2016-06-13
add select widgets for reference fields in admin interface (SDI)

joka:
To use the support for select widget the field node has to provide the function `choices_getter`
accepting `context` and `request` and returning a list with select options.


2e2a:
lgtm

Issue #2393, opened 2016-06-08
Can't build M2015 in `organisation/mercator` and then create proposal


der-john:
I've removed the build folder, built adhocracy-mercator afresh, restarted the server, imported the process resource, set the workflow state to 'participate', browsed to my local a3, filled in the form and clicked on 'submit'.

The proposal isn't posted. In the UI it says:

```
"<InterfaceClass adhocracy_mercator.resources.mercator.IMercatorProposal>" is not one of <InterfaceClass adhocracy_core.resources.document_process.IDocumentProcess>, <InterfaceClass adhocracy_core.resources.organisation.IOrganisation>, <InterfaceClass adhocracy_core.resources.process.IProcess>, <InterfaceClass adhocracy_mercator.resources.mercator.IProcess>
```

der-john:
@slomo just gave me the workaround: I imported the Mercator2015 process into `/organisation/`. THat did the trick.

joka:
It would be usefull to have a test import fixture for mercator to prevent this kind of errors

slomo:
Isn't this still a frontend code issue? Or are we saying won't fix here.

der-john:
You're right.

xi:
This process will probably not be used again, so it is not really relevant to fix it.

PR #2411, opened 2016-06-13
add mercator resources.json + fixtures

der-john:
Prevents #2393.

And adds fixtures for Mercator 2015 + 2016.

### DIFF
--- a/src/adhocracy_mercator/adhocracy_mercator/__init__.py
+++ b/src/adhocracy_mercator/adhocracy_mercator/__init__.py
@@ -20,7 +20,7 @@ def includeme(config):
     config.include('.evolution')
     config.add_translation_dirs('adhocracy_core:locale/')
     config.registry.registerUtility('', IFixtureAsset,
-                                    name='adhocracy_euth:fixture')
+                                    name='adhocracy_mercator:fixture')
 
 
 def main(global_config, **settings):
--- a/src/adhocracy_mercator/adhocracy_mercator/__init__.py
+++ b/src/adhocracy_mercator/adhocracy_mercator/__init__.py
@@ -2,6 +2,7 @@
 from pyramid.config import Configurator
 
 from adhocracy_core import root_factory
+from adhocracy_core.interfaces import IFixtureAsset
 
 
 def includeme(config):
@@ -18,6 +19,8 @@ def includeme(config):
     config.include('.resources')
     config.include('.evolution')
     config.add_translation_dirs('adhocracy_core:locale/')
+    config.registry.registerUtility('', IFixtureAsset,
+                                    name='adhocracy_euth:fixture')
 
 
 def main(global_config, **settings):
--- a/src/adhocracy_mercator/adhocracy_mercator/fixture/resources/mercator.json
+++ b/src/adhocracy_mercator/adhocracy_mercator/fixture/resources/mercator.json
@@ -1,0 +1,38 @@
+[ {"path": "/",
+   "content_type": "adhocracy_mercator.resources.mercator.IProcess",
+   "data": {"adhocracy_core.sheets.name.IName":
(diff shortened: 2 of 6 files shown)


joka:
we recently added the bin/sd_fixture script to fix this kind of issues.
Can you add test_fixture für mercator? (for example see  adhocracy_core.test_fixture/ and adhocracy_core.__init__.includeme) Then we can also remove src/mercator/mercator/tests/acceptance/setup_tests.sh .

der-john:
When I run
> bin/ad_fixtures etc/develeopment.ini
my new fixtures aren't shown as registered. How do I register them?

2e2a:
> my new fixtures aren't shown as registered. How do I register them?

The folder for fixtures is `fixture/`

der-john:
Thanks. I changed the name, but they're still not registered.

2e2a:
> Thanks. I changed the name, but they're still not registered.

Should be fixed with last commit

joka:
We should update the documentation regarding how to register a fixture.
You should add to the includeme function of the wantend backend customization package:
from adhocracy_core.interfaces import IFixtureAsset
 `config.registry.registerUtility('' ", IFixtureAsset, name='adhocracy_mercator:fixture')`

The important part ist the `name` keyword. This is a `pkg_resource`definition based of <python_package>:<relative path> or a just <absolute_file_system_path>

joka:
works now :)

joka:
The name of the fixture directory can be arbitrary, but should descript the fixtures, like 'test_fixture'

joka:
LGTM

joka:
Note: We could also  simplify registration with a config method like:

`config.register_fixture_directory(''my_fixtures_directory")

src/adhocracy_mercator/adhocracy_mercator/__init__.py
```diff
@@ -18,6 +19,8 @@ def includeme(config):
     config.include('.resources')
     config.include('.evolution')
     config.add_translation_dirs('adhocracy_core:locale/')
+    config.registry.registerUtility('', IFixtureAsset,
+                                    name='adhocracy_euth:fixture')
```
joka: you register the fixtures from the euth package here.

xi:
I dont't see the value of this pull request: It simply duplicates `src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator/resources.json` and `src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator2/resources.json`.

IMHO this PR makes the situation even more confusing than it already is. Instead, we should switch to the new fixture system completely (for all projects and removing the old files).

2e2a:
The resource files in `/workflows/test_some_process` are there for testing of the process workflows and we just misused them for import.

I agree that adding `src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator/resources.json` does not make sense, as it is not used in the tests.

But in general duplication can occur if the same data is used to create local processes and for the workflow tests.

joka:
> The resource files in /workflows/test_some_process are there for testing of the process workflows and we just misused them for import.
> 
> I agree that adding src/adhocracy_mercator/adhocracy_mercator/workflows/test_mercator/resources.json does not make sense, as it is not used in the tests.

yes this file should be removed

> But in general duplication can occur if the same data is used to create local processes and for the workflow tests.

We could also use the test_fixture for functional tests, that was one intended use case.
But this does not need to be part of this pull request.

der-john:
> we should switch to the new fixture system completely (for all projects and removing the old files).

I'll do that some time soon. (It's low prio, though.)

xi:
I cherry-picked 278a153 as c3d2e6fb88ef01f4bc8bb09ba1352a98cb488eea because it already adds value. Please open a new pull request when you convert the fixtures to the new system.

PR #2413, opened 2016-06-13
change wording + fontsize in UI (for ISEK)

der-john:
I'm totally unsure whether this change in font-size is sensible - I think we should ask Nico.

**Edit:** Talked to Nico and @JanaGa and decided to change `font-size-small`.


xi:
LGTM

PR #2415, opened 2016-06-13
2016 06 js process button slot+4

xi:
I tried to take over the spirit of #2389 while avoiding circular dependencies.


der-john:
has been rebased.

PR #2417, opened 2016-06-14
print idea collection process

xi:
Note that I changed the default print styling to be less aggressive.


der-john:
@JanaGa says we should try make the pin not use a background color.

der-john:
Very good improvement of the printed version.

PR #2418, opened 2016-06-14
2016 06 js process button slot+5

der-john:
Replaces #2415 & #2389 & #2377.

Continues #2113.


xi:
I really like the word "slot". But to me it seems strange that the thing that is put into a slot is also called slot. So not quite there yet…

der-john:
TODO:
- [ ] move `Process` to `ResourceArea`
- [ ] make `template` more similar to `buttonSlot`

PR #2419, opened 2016-06-14
replace 'mark-as-deleted' with adhHideAction

### LINKED ISSUE #2268: adhResourceAction.resourceWidgetDelete does not work

der-john:
(continues #2390 & #2394.) Needs #2420 for i18n.

I noticed that hide/delete functions were wrapped in extremely similar ways (including the browser window asking 'Do you really want to delete this?') in the following five places:

- IdeaCollection
- Alexanderplatz
- Blog
- Comment
- Mercator2015

@xi and I discussed the possibility to factor out this functionality. I suggested adding a resourceAction to replace `adhDeleteAction`, since HIDE should be used where 'mark-as-deleted' used to be used.

In this PR,

- a resourceAction called `adhHideAction` is implemented;
- the resourceAction called `adhDeleteAction` is removed - this fixes #2268.

The following use cases are replaced with `adhHideAction`:

- IdeaCollection
- Alexanderplatz
- Mercator2015

- Blog - because it's a similar use case to deleting badge assignments (see #2365).

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
@@ -25,7 +25,7 @@ var pkgLocation = "/Comment";
 export interface ICommentResourceScope extends angular.IScope {
     path : string;
     submit : () => any;
-    delete : () => angular.IPromise<void>;
+    hide : () => angular.IPromise<void>;
     refersTo : string;
     poolPath : string;
     hideCancel? : boolean;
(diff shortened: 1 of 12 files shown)


der-john:
Needs #2420 for i18n.

der-john:
everything that's sensible for now has been implemented in #2428.

PR #2420, opened 2016-06-14
translate delete confirmation

der-john:
Makes the string 'Do you really want to delete this?' translatable and translates it.

Continues #2390 & #2394.


xi:
I would have liked to merge #2419 first, but it is also fine this way.

PR #2421, opened 2016-06-14
make printers print markers on map by default

### LINKED ISSUE #2382: Printing bug for map pins
When we tried to print out a proposal in map view in the idea-collection, the pin of the proposal was white (instead of blue). The bug already appears in the preview.

lhttps://meinberlin-stage.liqd.net/w/prozesse/tegel/#!r/isek/tegel/

bildschirmfoto 2016-06-06 um 17 52 33

der-john:
This continues #2417. It *partially* fixes #2382.

**First attempt (outdated):**
See: [MDN: -webkit-print-color-adjust](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-print-color-adjust)
According to MDN, this only works for Chrome + Safari and is not a part of any standard. Still, I think it's an improvement.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_map.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_map.scss
@@ -83,14 +83,27 @@ rather than in response to user action.
     @include border-radius(50%);
     @include box-shadow(0 0 1px 1px $color-text-inverted);
     background: $color-brand-two-normal;
+    border: solid 10px $color-brand-two-normal;
     color: $color-text-inverted;
     text-align: center;
 
     &.is-active, &:hover, &:focus {
         background: $color-brand-one-introvert;
+        border: solid 10px $color-brand-one-introvert;
     }
 
-    &.marker-cluster {
+@media print {
+    .marker-cluster {
+        border: solid 20px $color-brand-two-normal;
+
+        &.is-active, &:hover, &:focus {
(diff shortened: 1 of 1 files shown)


xi:
I also considered using this but decided against it because this hides the issue instead of fixing it. Also `print-color-adjust` is not on any standards track, so it will lose support eventually.

der-john:
**Second attempt**: "Gigantic border solution". (Ugly.)

Unfortunately, centering the number on clusters (see #2349) *so far* was incompatible with
enlarging the border. So, for clusters, there's now a screen version and a printed version, and in the printed version, the numbers aren't centered.

der-john:
@xi @localgrr Maybe we can improve this and implement it?

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_map.scss
```diff
@@ -83,14 +83,27 @@ rather than in response to user action.
...
 
     &.is-active, &:hover, &:focus {
         background: $color-brand-one-introvert;
+        border: solid 10px $color-brand-one-introvert;
     }
 
-    &.marker-cluster {
+@media print {
+    .marker-cluster {
+        border: solid 20px $color-brand-two-normal;
```
xi: the 20px value is dynamically generated by Leaflet.markercluster. We should not hard-code it here. As far as I can see there is no way to determine `border-width` from `width` in CSS (e.g. `border-width: 100%` is invalid). So this approach is doomed to fail.

PR #2425, opened 2016-06-15
Add permission to delete proposal

JanaGa:
Add permission to delete proposal  to all user roles in participate phase in meinBerlin B-Plan workflow


JanaGa:
test: please don't merge

PR #2427, opened 2016-06-15
fixup #2404

xi:
#2404 Broke some acceptance tests. I fixed most of them,  but "user password reset email is sent to user" still seems to fail consistently.


der-john:
I get one failure:
> 1) user password reset error displayed if the email is not associated to an user

der-john:
looks good & works.

PR #2428, opened 2016-06-15
2016 06 js adh hide action+3

xi:
I rebased #2419 and did quite some changes. Unfortunately I am not completely done, but maybe you can already merge this or at least some commits.

Note that I removed 09caadb because I would like to look into it separately, considering #1711.


der-john:
I removed `contentType` from IdeaCollection and Alexanderplatz.

looks good & works.

PR #2429, opened 2016-06-16
move overlays from MovingColumns to Process

der-john:
this prepares the redesign: it decouples overlays/ alerts from the MovingColumns module, which is still essential to the layout & rendering of the frontend, though.


der-john:
It seems to me that `moving-column-menu`, `moving-column-tab` and similar layout elements should probably move to process.Column, too.

xi:
I don't really get what you did here. As far as I can see you renamed "moving column" to "column" and moved to the process module. But why?

der-john:
We talked about it yesterday. This PR makes sense.

But today Nico Di Marco and @xi decided to replace overlays and alerts with collapsible textboxes. So modifying this PR would be *verlorene Liebesmüh*.

PR #2435, opened 2016-06-17
introduce one-column design depending on context

der-john:
Needs + continues #2429.

Another step towards one-column design.

In the `mein.berlin.de` context, `IdeaCollection` routing hereby always sets `movingColumns` to `is-show-hide-hide`, thus displaying only one column.

The logic for displaying this column is moved to `Process` and simplified.

Todo:
- [ ] try to further simplify `showColumn`.
- [ ] decide whether making the design depend on the context suffices.
- [ ] implement this configuration in other workbenches.
- [ ] do away with `focus`.
- [ ] simplify layout.


der-john:
this has been an interesting experiment. we most liekly won't use this setup, though.

PR #2436, opened 2016-06-18
add support for requires.io

joka:
To list requirements go to:
 https://requires.io/github/liqd/adhocracy3/requirements/?branch=2016-06-jk-requirements

To update the requirement listing run: 
`bin/pip install  requires.io`
`bin/requires.io update-site -t a54831113b039e9edbb2d26c2d2f9a9c99887437 -r adhocracy3`


xi:
- How can I get permissions to change settings on requires.io
- Is it possible to add frontend dependencies?

2e2a:
We probably should also add the usage example to documentation.

joka:
> We probably should also add the usage example to documentation.

yes

joka:
> How can I get permissions to change settings on requires.io

You have to login with you github account, but the only option is to revoke the authentication token.

> Is it possible to add frontend dependencies?

No, requires only reads buildout and pip requirements files.

joka:
> We probably should also add the usage example to documentation.
> 
> yes

Do we want to add the authentication token to docs?

2e2a:
| Do we want to add the authentication token to docs?

Now its on github, so I think its ok to have it also in the docs. Not sure where we could put it otherwise.

2e2a:
lgtm

PR #2437, opened 2016-06-20
move embedded directives' registration

der-john:
`AdhEmbedProvider.registerDirective` gets used all over the place. So it's illogical to have seven directives be included in `AdhEmbed.Provider` by default for no reason. This PR moves them to the appropriate modules, so that only `empty` is left.


xi:
Nice!

PR #2373, opened 2016-06-01
add filters generically to meinberlin IdeaCollection

der-john:
this continues #2370:

as announced in #2347, IdeaCollection now gets the badges to be filtered from the backend; the `has-swot-labels` logic is replaced.

PR #2438, opened 2016-06-20
tidy up badge interfaces and functions

der-john:
Does a little cleaning up after #2373.

In this PR, I strongly suggest we rename `IBadge` to `IBadgeAssignment`. It becomes clear in `Badge.ts` and all its other use cases, that the values of this object are values connected to the backend resource called `BadgeAssignment`. Furthermore, in this PR, an interface for real badges is added.


xi:
LGTM. I will just wait for the tests to finish

PR #2439, opened 2016-06-20
fix s1 proposal detail recompile

xi:
In S1, the user name stays the old one after clicking to another user's proposal.


der-john:
looks good.

2e2a:
lgtm

PR #2440, opened 2016-06-20
fix redirect after hide

### LINKED ISSUE #2387: on deleting IdeaCollection proposal, user is redirected to Kiezkasse
When Petra deleted a proposal, she was redirected from ISEK Tegel to Kiezkassen (standard). It should stay on ISEK Tegel.

der-john:
fixes #2387.


xi:
I don't think it is a good idea to add a click handler to `adh-hide-action` for several reasons:

-  It does not work in the default case where `adh-hide-action` is used as part of `adh-resource-actions`.
-  For developers it is easy to miss that there are two click handlers.
-  For developers it is easy to miss that you have to set `cameFrom`.

PR #2245, opened 2016-04-20
Ensure blob dir for test database is created

slomo:
fixes #2242

PR #2441, opened 2016-06-20
make leaflet configurable

der-john:
We want to display a different map in a demo case. So the options for using leaflet should be configurable per installation.

### DIFF
--- a/buildout-meinberlin.cfg
+++ b/buildout-meinberlin.cfg
@@ -12,6 +12,7 @@ frontend.static_dir = src/meinberlin/meinberlin/build
 frontend_package_name = meinberlin
 backend_package_name = adhocracy_meinberlin
 redirect_url = /r/organisation/kiezkasse/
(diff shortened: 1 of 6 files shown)


buildout-meinberlin.cfg
```diff
@@ -12,6 +12,7 @@ frontend.static_dir = src/meinberlin/meinberlin/build
 frontend_package_name = meinberlin
 backend_package_name = adhocracy_meinberlin
 redirect_url = /r/organisation/kiezkasse/
+leaflet_options = "'https://maps.berlinonline.de/tile/bright/{z}/{x}/{y}.png', {maxZoom: 18}, {attribution: '© berlinonline'}"
```
xi: This is not valid JSON

xi:
This does not work because the settings from buildout are not automatically included in `/config.json`. See #2245 for a fixed version.

PR #2442, opened 2016-06-20
Fixup 4820d126: Add rtd dependency

xi:
readthedocs broke yet again: https://readthedocs.org/projects/adhocracy3/builds/4118724/

This has happened several times already and is becoming an annoyance. Does someone know how to check for rtd build errors on every pull request (like travis)?

### DIFF
--- a/rtd.txt
+++ b/rtd.txt
@@ -100,6 +100,7 @@ snowballstemmer==1.2.0
 sphinx-rtd-theme==0.1.9
 sphinxcontrib-blockdiag==1.5
 sphinxcontrib-actdiag==0.8.5
+sphinxcontrib-programoutput
 statsd==3.0.1
 testfixtures==4.1.2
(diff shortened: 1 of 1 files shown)


joka:
the problem is rtd needs a pip requirements file and we always forget to maintain it along with versions.cfg (buildout) and setup.py (distribute) dependencies.
We could use our make_wheels script to create a requirements file that can be added to the repro, or we could  try to use the requirements files instead of versions.cfg to pin versions (  https://pypi.python.org/pypi/collective.recipe.pip/0.3.4 )

joka:
> This has happened several times already and is becoming an annoyance. Does someone know how to check for rtd build errors on every pull request (like travis)?

no

joka:
@2e2a what do you thing about this?

joka:
I merge this now, the discussion is not part of the pull request.

PR #2443, opened 2016-06-20
change the colors of overlay and badge

der-john:
Our meinberlin POs alerted us to the fact that the colors of badges and overlays are not easily readible. That's why we're changing it for adhocracy3.


xi:
-  We should definitely add contrast checks to this pull request to make sure that these changes are actually sufficient.
-  You adapted the function colors in core, but they are typically overwritten in many projects. Please also adapt the overwrites.

xi:
I talked to @nicoladimarco and he will look into the colors again tomorrow.

der-john:
> You replaced color-function-neutral by color-function-informative in one place.

I can't see where.

der-john:
has been rebased.

PR #2445, opened 2016-06-20
allow to configure leaflet tile layer per installation

xi:
*supersedes #2441*

Before this, we used `"https://maps.berlinonline.de/tile/bright/{z}/{x}/{y}.png"` as a tile layer for all maps. Now this is configurable per installation and the default is `"http://{s}.tile.osm.org/{z}/{x}/{y}.png"`.

The new default was selected because it covers the whole planet instead of only germany.

For development, we may want to stick with berlinonline for some installations.

For others, we may want to switch to mapbox (e.g. `"https://api.mapbox.com/styles/v1/liqd/cipo49m9w002ldfm1omjmi5nw/tiles/256/{z}/{x}/{y}"`) (ask me or @MagdaN for the access token). Note however that with a free plan this is only allowed for [non-comercial use](https://www.mapbox.com/help/commercial/). In any case, we need to add proper [attribution](https://www.mapbox.com/help/attribution/#other-mapping-frameworks)

@joka I did not add tests for the python code. Is this ok? (Pending review.)

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/__init__.py
+++ b/src/adhocracy_frontend/adhocracy_frontend/__init__.py
@@ -1,4 +1,5 @@
 """Frontend view and simple pyramid app configurations."""
+import json
 import pkg_resources
 
 from pyramid.renderers import render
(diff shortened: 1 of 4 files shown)


der-john:
a3 has successfully arrived in Poznań.

PR #2446, opened 2016-06-21
Add deform_markdown to wheels configuration

2e2a:
There seems to be still a problem with the wheel creation in deform_markdown.


joka:
This package is released on pypi, we don't need a local checkout.
What is the problem with the wheel creation?

2e2a:
> This package is released on pypi, we don't need a local checkout.
> What is the problem with the wheel creation?

Started this pull request as we had problems with the deform_markdown wheel on staging.
The resulting wheel was almost empty.
I'll close the pull pull request.

PR #2450, opened 2016-06-22
more resource actions (wip)

der-john:
This works, but it's an eyesore. We should talk about it.


der-john:
**Update**: We don't need to talk about it anymore, because it sucks.

PR #2451, opened 2016-06-22
integrate overlays into resource actions

xi:
This implements overlays directly in resource actions.

Should work while keeping compatibility with legacy system.

TODO: static layout for alerts.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.html
@@ -11,9 +11,11 @@
         data-resource-path="{{resourcePath}}"></adh-moderate-action>
     <adh-report-action
         data-ng-if="report"
+        data-modals="modals"
         data-class="action-bar-item"></adh-report-action>
     <adh-share-action
         data-ng-if="share"
+        data-modals="modals"
         data-class="action-bar-item"></adh-share-action>
     <adh-hide-action
         data-ng-if="hide && resourcePath && options.hide"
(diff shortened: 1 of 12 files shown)


der-john:
very nice! looks great & works.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.html
```diff
@@ -34,4 +36,24 @@
...
+        {{ alert.message | translate }}
+    </li>
+</ul>
+
+<div class="action-bar-modal" data-ng-if="modals.overlay">
+    <adh-report-abuse
+        data-ng-if="modals.overlay === 'abuse'"
+        data-modals="modals"
+        class="report-abuse"
+        data-url="{{ proposalUrl | adhResourceUrl | adhCanonicalUrl }}">
```
der-john: Why did you leave `proposalUrl` and didn't change it to `resourcePath`? (It seems to work fine, but I thought `resourceActions` are not just about proposals.)

xi: Indeed this is an error. I will fix it.

der-john:
Is it possible that we forgot `adh-share-action` here as well?

xi:
> Is it possible that we forgot `adh-share-action` in bcd399e?

As far as I can see it is not used anywhere outside of AdhResourceActions

der-john:
> Is it possible that we forgot adh-share-action in bcd399e?
> > As far as I can see it is not used anywhere outside of AdhResourceActions

You're right.

PR #2452, opened 2016-06-23
use resource actions + modals in comments

der-john:
Needs #2451.

The create and edit functions on comment were too specific as to quickly make them into `resourceActions`. So this PR just converts report and hide - these are the ones currently using modals, anyways - we want to use the new moodal functionality (see #2451) and display it where the comment is. Unfortunately, right now the "overlay"/modal appears above the rate buttons, even though these buttons were originally right next to the hide and delete buttons. This should be changed.

I'm sure the design needs polishing - e.g. in the report-abuse helptext, there's unnecessary whitespace in front of the two links. Plus we should consider using something like `AdhMovingColumnsController.bindVariablesAndClear` so that the modal disappears when you click outside of it (or so).

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
@@ -235,20 +235,6 @@ export var commentDetailDirective = (
             });
         };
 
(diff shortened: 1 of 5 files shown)


der-john:
after mulling about this, I've realized `resourceActions` in comments introduce more problems than benefits. the new modals/ pop-ins will be implemented differently.

PR #2453, opened 2016-06-23
reimplement adhDeleteAction for Blog

der-john:
This is made possible by #2365.

Unfortunately, I haven't tested it. There seems to be no UI for assigning badges in Mercator2015, so I can't see a journey.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
@@ -103,6 +103,43 @@ export var hideActionDirective = (
     };
 };
 
+export var deleteActionDirective = (
+    adhHttp : AdhHttp.Service<any>,
+    adhTopLevelState : AdhTopLevelState.Service,
+    adhResourceUrlFilter,
+    $translate,
+    $window : Window
+) => {
+    return {
+        restrict: "E",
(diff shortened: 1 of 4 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
```diff
@@ -103,6 +103,43 @@ export var hideActionDirective = (
...
+                return $translate("TR__ASK_TO_CONFIRM_DELETE_ACTION").then((question) => {
+                    var path = scope.parentPath ? AdhUtil.parentPath(scope.resourcePath) : scope.resourcePath;
+                    if ($window.confirm(question)) {
+                        return adhHttp.delete(path).then(() => {
+                            var url = scope.redirectUrl;
+                            if (!url) {
+                                var processUrl = adhTopLevelState.get("processUrl");
+                                url = processUrl ? adhResourceUrlFilter(processUrl) : "/";
+                            }
+                            adhTopLevelState.goToCameFrom(url);
```
xi: The blog entries do not have their own pages, so it does not make sense to redirect here. Note that in the original code, `scope.onChange()` was called.

der-john:
we need this feature for AE2016.

der-john:
the concept of this PR was not very good and its relevancy is diminishing all the time because the code is only used in `mercator`.

PR #2454, opened 2016-06-27
pop-ins for comments

der-john:
Needs #2451. Replaces #2452. (I seriously considered implementing `adhResourceActions` in Comment. But a) the create and edit functions on comment were too specific as to quickly make them into `resourceActions`, and b) on triggering a pop-in, the rate interface was displayed below it.)

There was a design decision with Nico Di Marco last week: we will henceforth display overlays and alerts wherever they have been triggered - as pop-ins.

The reason is this: Up to now, when you saw an embedded adhocracy with a long list of comments and clicked on 'report' concerning the very last comment, an overlay was triggered at the very top of the iframe. But this might be momentarily invisible, the user would have to scroll to the top in order to either report an abuse or close the overlay and then scroll back down to the comment.

Todo (maybe outside of this PR):
- [ ] Design refinements are probably necessary.
- [ ] We need to decide whether a 'user action' outside of the pop-in should cause it to disappear.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
@@ -185,8 +185,9 @@ export var commentDetailDirective = (
     var _postEdit = postEdit(adhHttp, adhPreliminaryNames);
 
     var link = (scope : ICommentResourceScope, element) => {
+        scope.modals = new AdhResourceActions.Modals($timeout);
+
         scope.report = () => {
-            scope.modals = new AdhResourceActions.Modals($timeout);
             scope.modals.toggleOverlay("abuse");
         };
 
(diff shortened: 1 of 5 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Comment.ts
```diff
@@ -176,18 +178,17 @@ export var commentDetailDirective = (
...
-    var link = (scope : ICommentResourceScope, element, attrs, column? : AdhMovingColumns.MovingColumnController) => {
-        if (column) {
-            scope.report = () => {
-                column.$scope.shared.abuseUrl = scope.data.path;
-                column.toggleOverlay("abuse");
-            };
-        }
+    var link = (scope : ICommentResourceScope, element) => {
+        scope.report = () => {
+            scope.modals = new AdhResourceActions.Modals($timeout);
```
xi: why do you re-create `scope.modals` on every report? Wouldn't it be better to create it once (directly in `scope.report()`)?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Comment/Detail.html
```diff
@@ -51,6 +51,21 @@
             </div>
         </div>
 
+        <ul class="moving-column-alerts">
+            <li data-ng-repeat="(id, alert) in modals.alerts" class="moving-column-alert m-{{alert.mode}}" data-ng-click="modals.removeAlert(id)">
+                {{ alert.message | translate }}
+            </li>
+        </ul>
+
+        <div class="action-bar-modal" data-ng-if="modals.overlay">
```
xi: It is not a good idea to reuse `action-bar-modal` here. But we can do CSS refactoring later.

xi:
Nice! Except for my comment this looks really good.

PR #2455, opened 2016-06-27
cleanup #2451

xi:
These are some amendments to #2451. I am sorry for this arbitrary collection, it was just stuff that I stumbled upon while trying to figure out what I have to do next.

I am not absolutely sure about 3bc252f: It seems like a useful replacement for `adhMovingColumnController.bindVariablesAndClear()`, but it is less flexible (only clears on changes of `resourceUrl`). It is also not necessary for the cases that I saw because `<adh-resource-actions>` is used inside of an `<adh-recompile-on-change>`.


der-john:
looks good & works.

PR #2456, opened 2016-06-27
use resourceActions in alexanderplatz

xi:


der-john:
I found out a wrong variable name (`path` instead of `poolPath`) made creating proposals impossible and fixed it.

der-john:
looks very good & works.

PR #2457, opened 2016-06-27
rename default group & set local roles for default group with workflow 

### LINKED ISSUE #2430: Simplify setup private processes
Motivation:

We need to setup mixed private and public processes for S1 and the setup like for euth is quite complicated to understand.

Goal:

- Extending standard  workflow should be enough to setup private workflows in  the following use cases:  register users dis/enabled,  mixed/only private/public processes

Proposal:

- rename authenticated group in default_group (default user group for new created users)
- remove global participator role from default_group   (the need for global roles is platform specific and should not be the default)
- extend workflow definition with local rules : `add_local_role_participant_to_default_group: True`

Risks: 

   low, not complicated and no productive installation is modified, except the inital setup scripts should set the 'participator' role for the default group if wanted.

joka:
fixes #2430 

Internal API changes:

-   add optionn `'add_local_role_participant_to_default_group'` to workflow definition 
    `True` for 'standard' workflow, `False` for private workflows. 
    DEPLOYMENT:  Using deplyoment scripts to set local role:participant for default group is not needed anymore

- set local roles with workflow state

- rename default group from `authenticated` to `default_group`
  DEPLOYMENT:  mind to invalidate varnish cache


src/adhocracy_meinberlin/adhocracy_meinberlin/__init__.py
```diff
@@ -18,6 +19,8 @@ def includeme(config):
     config.include('adhocracy_meinberlin.evolution')
     config.add_translation_dirs('adhocracy_core:locale/',
                                 'adhocracy_meinberlin:locale/')
+    config.registry.registerUtility('', IFixtureAsset,
+                                    name='adhocracy_meinberlin:test_fixture')
```
2e2a: Fixture does not exit

PR #2458, opened 2016-06-27
use modals for user messages

xi:
Made possible by #2451. Part of decoupling functions from MovingColumns.

Note that using resource actions was not easily possible here because the modal can also be triggered from `<adh-user-profile>`.


der-john:
Nice! looks good & works.

xi:
This broke an acceptance test: "user page is possible to send a message". We should fix it as soon as possible.

PR #2461, opened 2016-06-28
Use remove in pools instead of delete to be substanced compliant

2e2a:


joka:
Fine, can you squash all commits to clean up the commit history?

joka:
and rebase master

2e2a:
> and rebase master
done

PR #2462, opened 2016-06-28
add jquery-based animation

xi:
See https://docs.angularjs.org/api/ngAnimate#id=%22javascript-based-animations%22 and https://api.jquery.com/slideDown/


der-john:
looks good & works.

PR #2463, opened 2016-06-28
Fix badge reindexing

2e2a:
- Add graph function to send SheetBackReferenceRemoved events for references
- Trigger this events after deletion
- Recompute badge index on this event


joka:
LGTM

PR #2465, opened 2016-06-28
 refactor token authentication, MIGRATION issue

joka:
Internal changes:

- simplify backend code to mange the authentication token. Implementation is now  based on https://jwt.io (token signed by external servers are not allowed, so the basic security architecture stays the same)

Migration:

- all  old authentication tokens will be invalid. Logged in frontend users will see the error message: "Expired or invalid authentication token" and have to relogin.


2e2a:
lgtm

PR #2468, opened 2016-06-29
more modals

der-john:
this makes euth & pcompass use modals instead of moving-column overlays. needs #2451.


xi:
LGTM.

I guess it is possible to re-use some individual resource actions here, e.g. edit and assigne badges. But that can be done separately.

PR #2472, opened 2016-06-30
Add content to auditlog

2e2a:


src/adhocracy_core/adhocracy_core/auditing/__init__.py
```diff
@@ -92,8 +99,17 @@ def audit_resources_changes_callback(request: Request,
...
+        if _is_audit_change(change):
+            action_name = _get_entry_name(change),
+            sheets = _get_content_sheets(change, registry)
+            sheet_data = _get_sheet_data(sheets, request)
+            log_auditevent(change.resource,
+                           action_name,
+                           user_name=user_name,
+                           user_path=user_path,
+                           sheet_data=sheet_data)
+            transaction.commit()
```
joka: Not part of this pull request, but why do we need a commit for every single auditlog?

src/adhocracy_core/adhocracy_core/sheets/__init__.py
```diff
@@ -228,7 +228,7 @@ def _store_references(self, appstruct, registry, send_event=True):
                                               registry,
                                               send_event=send_event)
 
-    def serialize(self, params: dict=None):
+    def serialize(self, params: dict=None, add_back_references: bool=True):
         """Get sheet appstruct data and serialize.
```
joka: test is missing

src/adhocracy_core/adhocracy_core/auditing/__init__.py
2e2a: | Not part of this pull request, but why do we need a commit for every single auditlog?

It was already there, I don't see why. I think we can remove it.

PR #2478, opened 2016-07-04
abstract resource actions

xi:
I propose three changes to resource actions:

- convert some similar actions into parameterized/generic actions
- make the `action-bar-item` class implicit
- remove `parentPath` parameter

With neither of them I am completely sure, so this is a proposal rather than a request.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.html
@@ -19,14 +19,18 @@
         data-class="action-bar-item"
         data-resource-with-badges-url="{{resourceWithBadgesUrl}}"
         data-resource-path="{{resourcePath}}"></adh-assign-badges-action>
-    <adh-report-action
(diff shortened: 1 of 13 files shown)


der-john:
589ffb2 was reverted because in the forseeable future, i.e. after a redesign, we will use resource actions outside of action bars.

der-john:
looks good & works.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.html
```diff
@@ -3,13 +3,11 @@
...
         data-label="TR__EDIT"
-        data-parent-path="parentPath"
         data-resource-path="{{resourcePath}}"></adh-view-action>
     <adh-view-action
         data-ng-if="moderate && resourcePath"
         data-view="moderate"
         data-label="TR__MODERATE"
-        data-parent-path="parentPath"
         data-resource-path="{{resourcePath}}"></adh-view-action>
     <adh-assign-badges-action
```
der-john: this possibly caused #2572.

PR #2479, opened 2016-07-05
mercator 2016 blog/logbook/journey

xi:
This is a simple port of the logbook feature from mercator 2015 to mercator 2016. This PR only covers frontend and is on hold until backend is done. We should check that everything works as expected once both parts are merged.


der-john:
Todo: see #2374 
- [ ] Fix delete in pools for versions (e.g. blog entries)

der-john:
I have two problems:
1. The old proposals created before this PR cause an error because they have no `ILogbook`. Therefore I couldn't run AE.
2. I wanted to create a new proposal in a new AE process. I couldn't because of the following error:
> FileNotFoundError: [Errno 2] No such file or directory: 'var/db/adhocracy_mercator/blobs/tmp/BUC0__4smxy'

2e2a:
> I have two problems:
> 1. The old proposals created before this PR cause an error because they have no ILogbook. Therefore I couldn't run AE.

There is a migration fixing this, however there is another migration problem that still open and is causing problems.

> 2. I wanted to create a new proposal in a new AE process. I couldn't because of the following error:

Both problems could be fixed with a removing the database and rebuilding the project.

der-john:
> Both problems could be fixed by removing the database and rebuilding the project.

You were right.

der-john:
looks good & works.

PR #2480, opened 2016-07-05
resource header CSS

xi:
This adds a first version of the resource header CSS. It is not yet used anywhere. Also note that there is not yet a dropdown widget for resource actions.


der-john:
looks good & works.

PR #2482, opened 2016-07-05
Switch from compass to node-sass

xi:
This pull request removes the dependency on compass, sass, and susy and instead adds dependecy on grunt-sass, node-sass, and sass-planifolia. The resons for that are:

-  node-sass is faster
-  compass is no longer developed
-  we may be able to remove the ruby dependency completely (hologram and and scss-lint still remaining)

Note that [sass-planifolia](https://github.com/xi/sass-planifolia) is not a popular library. It is in fact maintained by myself. I think that is fine though as a large part of it is based on the contrast code from a3. So you could think of planifolia as a library that originated from a3 and is now maintained separately.


joka:
yeah, less ruby more javascript

der-john:
See c498d79: @include inline-block has previously set vertical align in many places. This was not always
required. Some other places could also be affected, but this commit covers most of it.

PR #2488, opened 2016-07-06
switch to js based sass linter

xi:
Following #2486, this replaces the ruby based scss-lint by node based sass-lint.

sass-lint does not yet support inline comments to disable specific rules (there is a pull request though: https://github.com/sasstools/sass-lint/pull/677). I switched the affected rules to warning mode, meaning they will provide warning messages, but the process will still exit with exitcode 0.


.sass-lint.yml
```diff
@@ -0,0 +1,41 @@
+files:
+  include: '**/*.scss'
+  ignore: '**/third_party/**/*.scss'
+options:
+  formatter: stylish
+  merge-default-rules: true
```
der-john: I get `merge-default-rules: false` from pasting our scss lint config on https://sasstools.github.io/make-sass-lint-config/ .

.sass-lint.yml
```diff
@@ -0,0 +1,41 @@
+files:
```
der-john: > For all rules, setting their severity to 0 turns it off, setting to 1 sets it as a warning (something that should not be committed in), and setting to 2 sets it to an error (something that should not be written).
-- https://github.com/sasstools/sass-lint

der-john:
looks good & works.

.sass-lint.yml
xi: Yes, I adapted the output manually.

PR #2489, opened 2016-07-07
move meinberlin ideaCollection to core

der-john:
In order to simplify the workbench template, I removed the process-edit functionality once contained in Kiezkasse. It hasn't been used and will be managed by SDI in the future.

In another PR, euth & pcompass ideaCollections will be replaced by this one. This will reduce maintenance in the future.

Process-type-specific options of the ideaCollection are stored in a new object: `AdhProcess.IProcessOptions`.

@xi and I considered using adapters instead. The problem is: say we register a directive X with a base adapter. If we want to register directive X with another adapter, we need to give it another name in its registration (it can't be registered twice with the same name with different adapters). But meanwhile, directives that use directive X need to use the name X in their templates, or else they don't access the second adapter's code.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/IdeaCollection/Workbench/Workbench.ts
```diff
@@ -148,7 +148,6 @@ export var addProposalButtonDirective = (
 export var registerRoutesFactory = (
     ideaCollection : string
 ) => (
-    processType : string = "",
```
xi: I don't think this is a good idea because it reduces consistency.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/IdeaCollection/Proposal/Proposal.ts
```diff
@@ -249,7 +251,8 @@ export var detailDirective = (
         scope: {
             path: "@",
             isKiezkasse: "=?",
-            isBuergerhaushalt: "=?"
+            isBuergerhaushalt: "=?",
+            processOptions : "="
```
xi: No space here

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/IdeaCollection/Proposal/Proposal.ts
```diff
@@ -75,17 +67,19 @@ var bindPath = (
...
+                return adhHttp.get(processUrl).then((process) => {
+                    var locationUrl = process.data[SILocationReference.nick]["location"];
+                    return adhHttp.get(locationUrl).then((location) => {
+                        return location.data[SIMultiPolygon.nick]["coordinates"][0][0];
+                    });
+                });
+            } else {
+                return $q.when();
+            }
+        };
```
xi: Indentation is off

xi:
Nice one! It is so good to get this finally done.

xi:
For reference:

There are two options:

1. Have all features in core and selectively activate them in the different usecases
    - Issue: Resource classes are not available in core.
2. Have only the basic features in core and extend them for each usecase.
    - Issue: Templates can not easily be extended

Solution: Have a hybrid! The template is completely in core with all features. But mapping resources to the scope can happen in customizations (adapter).

`IProcessOptions` is a very simple adapter that only contains information about which resource classes to use (+ one sheet). This works because the available proposal classes are very similar. It may not scale though.

der-john:
I've forgotten to test `maxBudget = Infinity' .

xi:
> The max attribute, if specified, must have a value that is a valid floating-point number.
>
> -- https://www.w3.org/TR/html5/forms.html#number-state-(type=number)

> The Infinity and Not-a-Number (NaN) values are not valid floating-point numbers.
>
> -- https://www.w3.org/TR/html5/infrastructure.html#valid-floating-point-number

PR #2491, opened 2016-07-07
Switch to sass-planifolia 0.3.0

xi:
I just released a new version, so we can switch to a stable release now.

Note: I did not test this as I do not currently have a adhocracy development setup


der-john:
works.

der-john:
I've gist-reviewed the diff of sass-planifolia 0.3.0 > 0.2.1.

PR #2494, opened 2016-07-08
Fix landing and login alignment (centered!) in IE

dwt:
Rework how the landing page is centered and make it work in IE. Also take of custom styling of the .l-center class as it is reused for login (which also didn't center correctly in IE 10+).


xi:
LGTM.

PR #2495, opened 2016-07-08
Fix weird spacing in header.

dwt:
Was caused by the way spaces where suppressed between the buttons (with a float). Fix is to just remove the spaces.


xi:
Sorry for the regression. I did some CSS refactoring in #2482 and it seems like I caused some regressions. See c498d79 for this one. The canonical fix is to add `vertical-align: middle`. Feel free to merge.

dwt:
LGTM - Thanks

PR #2502, opened 2016-07-12
rm dead css

xi:
I used the following script to find unused classes:

```
classes=$(grep '\.[a-z]' `git ls-files | grep scss$` | grep -v 'url(' | grep -v '<' | grep -o '\.[a-z][^ ,.#:>;()`/]*' | sed 's/^\.//' | grep -v '^color-' | sort | uniq)
for cls in $classes; do
    matches=$(git grep -l $cls | grep -v 'scss$' | wc -l)
    if [ $matches -eq 0 ] || [ $1 = '-a' ]; then
        echo $cls $matches
    fi
done
```


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_forms.scss
```diff
@@ -91,7 +91,7 @@ Textinput that represents a title.
 ```
 */
 
-.error-state {
+%error-state {
```
der-john: nifty.

der-john:
nice.

PR #2503, opened 2016-07-12
use default proposal css in idea collection

xi:
*Depends on #2502*

The styles are mostly the same anyways.


der-john:
looks good & works.

PR #2504, opened 2016-07-12
fix idea collection list item

xi:
*Depends on #2503*

The position of the list item contents was broken:

2016-07-12-144631_1440x900_scrot

This is the fixed version:

2016-07-12-144426_1440x900_scrot


der-john:
I noticed that in the meinberlin proposal list, fallback image thumbnails are shown, which is not desired.

xi:
Please do not rebase public branches!

der-john:
I missed several tasks. Thanks for getting them for me!

PR #2508, opened 2016-07-12
only show image upload option when existent

### LINKED ISSUE #2507: don't show image upload in meinberlin ideaCollection
in the proposal detail view in meinberlin ideaCollection, the action-bar-item for uploading an image is shown, even though it doesn't work.

der-john:
Fixes #2507. Cleanup of #2492.


xi:
LGTM.

PR #2511, opened 2016-07-12
idea collection resource header

xi:
I applied the resource header introduced in #2480 in the core idea collection (see #2489). It was a bit more complicated that I had hoped because rate and background image.

Note that the resource header design is far from final and may change significantly in the future. Still I think it is good to get this normalized.


der-john:
As written by @xi, 		8b7480c temporarily breaks the pcompass design.

der-john:
looks good & works.

PR #2481, opened 2016-07-05
cleanup css

xi:
Just some low hanging fruits for now.

PR #2516, opened 2016-07-13
css cleanup

xi:
Some more CSS  cleanup. this one is mostly about moving files.

This continues #2481 and #2502. I plan to continue creating pull requests about CSS refactoring in small steps.


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/a3.scss
```diff
@@ -19,7 +19,7 @@
 @import "general/layout";
 @import "general/print";
 
-@import "components/type";
+@import "components/base";
```
der-john: nice.

src/meinberlin/meinberlin/static/stylesheets/scss/_theme.scss
```diff
@@ -1,4 +1,5 @@
 @import "general/layout_theme";
+@import "components/base_theme";
```
der-john: coincidentally, this is ambiguous. but I like "base" nevertheless.

src/s1/s1/static/stylesheets/scss/components/_base_theme.scss
```diff
@@ -1,4 +1,3 @@
 html {
-    @include rem-baseline;
```
der-john: I'm not up-to-date on s1 - but this isn't part of renaming.

der-john:
looks good & works.

src/s1/s1/static/stylesheets/scss/components/_base_theme.scss
xi: This property was a duplicate (also in core) so I removed it. I agree that it does not really belong into this commit.

PR #2523, opened 2016-07-14
Allow multiple S1 processes

nidico:
* Don't use hardcoded `s1_process_url` for S1 header buttons
* Show meaningful process information in header and landing page

Note that this makes use of the header slot concept, so that other processes can also inject their `ProcessHeaderSlot` into the S1 platform header.

### DIFF
--- a/src/s1/s1/static/js/Packages/S1/Proposal/Proposal.ts
+++ b/src/s1/s1/static/js/Packages/S1/Proposal/Proposal.ts
@@ -306,7 +306,7 @@ export var listingDirective = (
         },
         link: (scope) => {
             scope.contentType = RIProposalVersion.content_type;
-            scope.$on("$destroy", adhTopLevelState.bind("processUrl", scope));
+            scope.$on("$destroy", adhTopLevelState.bind("processUrl", scope)); // REFACT is this even neccessary?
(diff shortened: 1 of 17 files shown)


src/s1/s1/static/js/Packages/S1/Proposal/Proposal.ts
```diff
@@ -306,7 +306,7 @@ export var listingDirective = (
         },
         link: (scope) => {
             scope.contentType = RIProposalVersion.content_type;
-            scope.$on("$destroy", adhTopLevelState.bind("processUrl", scope));
+            scope.$on("$destroy", adhTopLevelState.bind("processUrl", scope)); // REFACT is this even neccessary?
```
der-john: yes. It's used in `Listing` to get list elements (=proposals).

dwt: Feel free to remove the comment then. :)

dwt:
When testing, please note that the s1-process should have a title, short and long description (editable via the SDI) which should be displayed on the landing page and the header.

der-john:
looks good & works.

dwt:
thanks

PR #2525, opened 2016-07-18
resource dropdown

der-john:
This is part of the redesign. Like the resource header (see #2511 ), it's now first implemented in the idea collection, and may be implemented elsewhere easily some time later.

It turns out the new dropdown - I tentatively named `resource dropdown` - shouldn't be part of the resource header itself: When a command in the dropdown opens a modal, this modal shouldn't appear within the area of the background image. So, in this PR, it's installed **above** the resource header.

Icons seem to be working at the moment, but I'm not entirely sure how I accomplished it.

PR #2530, opened 2016-07-20
dropdown above the resource header

der-john:
This (replacing #2525) is part of the redesign. Like the resource header (see #2511 ), it's now first implemented in the idea collection, and may be implemented elsewhere easily some time later.

It turns out the new dropdown - I tentatively named 'resource dropdown' - shouldn't be part of the resource header itself: When a command in the dropdown opens a modal, this modal shouldn't appear within the area of the background image. So, in this PR, following Nico Di Marco's suggestion, the dropdown is placed above the resource header.

Icons seem to be working at the moment, but I'm not entirely sure how I accomplished it.


der-john:
test on a narrow screen passed.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/IdeaCollection/Proposal/Proposal.ts
```diff
@@ -235,7 +235,9 @@ export var detailDirective = (
             path: "@",
             processOptions: "="
         },
-        link: (scope : IScope) => {
+        link: (scope) => {
+            scope.$on("$destroy", adhTopLevelState.bind("processUrl", scope));
+            scope.$on("$destroy", adhTopLevelState.bind("proposalUrl", scope));
```
xi: Not necessary, this is `path` in here.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceDropdown.html
```diff
@@ -0,0 +1,74 @@
...
+            data-redirect-url="{{data.deleteRedirectUrl}}"
+            data-resource-path="{{data.resourcePath}}"></adh-hide-action>
+        <adh-resource-widget-delete-action
+            data-ng-if="data.resourceWidgetDelete && data.resourcePath && options.hide"
+            data-resource-path="{{data.resourcePath}}"></adh-resource-widget-delete-action>
+        <adh-print-action
+            data-ng-if="data.print"></adh-print-action>
+        <adh-cancel-action
+            data-ng-if="data.cancel && data.resourcePath"
+            data-resource-path="{{data.resourcePath}}"></adh-cancel-action>
```
xi: This is invalid. Each action must be wrapped in a `<li>`

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_dropdown.scss
```diff
@@ -73,21 +73,21 @@ A basic dropdown that is currently used for resource actions.
...
 .dropdown-menu {
-    border-radius: 3px;
+    border-radius: 2px;
     float: left;
     min-width: 120px;
     padding: 0;
     margin: 5px 0 0;
     background-color: $color-text-inverted;
-    border: 1px solid $color-text-introvert;
+    border: 0.5px solid $color-structure-normal;
```
xi: `0.5px`?? How is that supposed to work?

https://stackoverflow.com/questions/20116132/half-pixel-in-border-width-size-it-is-not-showing

xi:
I did not go into all depth yet. But I think I can already see that we have to be careful with accessibility on this one. I propose to skip the icons for now and concentrate on the dropdown itself first.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_dropdown.scss
der-john: `0.5px` per `li` element inside the menu plus `0.5px` for the `dropdown-menu` itself makes the border exactly `1px` everywhere (I hope).

der-john:
This last commit fixed issues (1) and (4).

der-john:
I'll rebase and republish soon.

PR #2533, opened 2016-07-21
Delete follow-up: metadata and options

2e2a:
See #2374
-  remove 'deleted field in IMetadata sheet in backend
- remove special 'hidden' field for metadatda sheet in options requests (not needed anymore)


2e2a:
@xi, @der-john:
As the hidden/deleted differentiation is obsolete, the flags are removed from the IMetadata sheet in the options request (See deletion.rst). This requires  a frontend adaptation.

der-john:
I'm a little out of my depth here: I can see no occurrences of "deleted" in the frontend anymore. And deleting comments works perfectly fine. But the options request for `hide` seems to return `false` all the time, even for god. Some part of the `Http` module probably needs adaption, but I don't which right now.

src/adhocracy_core/docs/deletion.html
```diff
@@ -0,0 +1,640 @@
...
+
+&gt;&gt;&gt; resp = admin.get(&quot;/pool1&quot;)
+&gt;&gt;&gt; resp.status_code
+404
+</pre>
+</div>
+</div>
+</div>
+</body>
+</html>
```
joka: this html file does not belong here

joka:
yeah, less code, backend looks good

2e2a:
tested frontend adaptation

PR #2537, opened 2016-07-25
cleanup #2489

der-john:
As @nidico pointed out recently, the name 'meinberlin' still occurs in the code of the idea collection.

This PR removes those legacy occurrences for the sake of transparency.


xi:
LGTM

PR #2541, opened 2016-07-25
remove irrelevant theming

xi:
-  The login styles have next to no effect due to changes in core
-  The comment styles do have an effect, but they do not justify deviating from core styles.


der-john:
looks good & works.

PR #2542, opened 2016-07-25
rm chapter

xi:
This removes the chapter widget. It did not do much in the first place, still it was surprisingly hard to remove it without braking usability (especially in mercator).

**Note**: The `.chapter` selector is still used in some mercator acceptance tests. These tests are disabled by default and fail anyway (probably because of a wrong workflow state). I did not want to spend time on this now, but it should be easy to fix when the time comes.


src/mercator/mercator/static/stylesheets/scss/components/_base_theme.scss
```diff
@@ -4,3 +4,13 @@ h1, h2, h3, h4, h5, h6 {
...
+
+body,
+moving-column-body {
+    counter-reset: h2;
+}
+
+h2.m-counted:before {
+    counter-increment: h2;
+    content: counter(h2) ". ";
+}
```
der-john: cool. I didn't know we numbered chapters just with CSS.

der-john:
looks good & works.

PR #2545, opened 2016-07-26
refactor list item

xi:
Until now there was no general concept on how list items look. The layout (especially badge position) was widely different across workbenches.

I tried to fix this. However, there were two cases that I skipped: users (very different and not used) and s1 proposals (very different and up for refactoring soon).


der-john:
looks good & works. tests pass locally.

PR #2548, opened 2016-07-26
resource header actions

der-john:
this replaces #2530. Like the resource header (see #2511 ), it's now first implemented in the idea collection, and may be implemented elsewhere easily some time later.

Icons seem to be working at the moment, but I'm not entirely sure how I accomplished it.


der-john:
oops I missed something.

PR #2552, opened 2016-07-27
refactor jump navigation

xi:
This pull request merges `section-jump`,  `jump-navigation` and `mercator-proposal-form-nav` as they had overlapping functionality. I hope that the new version is clearer.


der-john:
I don't think it makes mercator 2016 prettier. But it makes the CSS more maintainable.

der-john:
looks good & works. test pass locally.

PR #2556, opened 2016-07-29
Editable user description

dwt:
This implements the PAAs story https://tree.taiga.io/project/liqdinga-paas-participation-as-a-service/us/28?milestone=65707

The idea is that a) the short description and description of a user is displayed in the profile and b) a user can himself modify this description on the users page.


src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_markdown.scss
```diff
@@ -9,3 +9,17 @@
         margin-bottom: 5px;
     }
 }
+
+.inline-editable-markdown {
+
+    .inline-editable-markdown-edit-button-container {
+        text-align: right;
+        margin-bottom: rem(10px);
```
xi: should be `@include rem(margin-bottom, 10px)`

xi:
Code looks good, but I am not completely convinced by the UI. I talked to @nicoladimarco and we had the following ideas/questions:

-   Why are there two descriptions?
-   The description should be limited in length (150-200 chars)
-   It should be plain text (no markdown)
-   The buttons feel a bit out of place

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/general/_assistive.scss
```diff
@@ -0,0 +1,29 @@
+/*doc
+---
+title: Screen reader only modifier
```
xi: IMHO this name is misleading. This is visually hidden, i.e. it is not hidden on a wide range of user agents apart from screen readers. This is a minor issue though.

nidico:
> Why are there two descriptions?

Because they're in the model like that...

> The description should be limited in length (150-200 chars)

This sounds very short to me. Also, the restriction should be applied in the backend / SDI, no?

> It should be plain text (no markdown)

Why?

> The buttons feel a bit out of place
> / ... not completely convinced by the UI

We came up with this rather quickly, knowing that we're no designers. If @nicoladimarco has other ideas, feel free to implement them.

xi:
> > Why are there two descriptions?
>
> Because they're in the model like that...
>
> > The description should be limited in length (150-200 chars)
>
> This sounds very short to me. Also, the restriction should be applied in the backend / SDI, no?

The IDescription sheet is not specific to users, so it is up to us to decide how to use it in this case.

nidico:
> The IDescription sheet is not specific to users, so it is up to us to decide how to use it in this case.

True. Thinking about it again, I'm also fine with a short description.

dwt:
In the last design meeting, we discussed to use both sheets, so one of them can be pre-initialised to the company position, and the other one can then be used by users to describe themselves (short: company title, long: self description).

By making both fields editable, users should then be able to self-correct / update company titles as well.

dwt:
I'd love to get input on how to better place the edit buttons. Do you have any suggestions?

One Idea we had was to put them as overlays on top of the text when hovering, like the profile image edit button. Still inline editing is a rather new concept in adhocracy afaik.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/general/_assistive.scss
dwt: I agree, I'm willing to follow the example of twitter bootstrap here though, until I find a better way to achieve this. Do you have an Idea of how to better do this?

dwt:
I've had a talk the local designer and will follow his suggestions for how to make this look. Expect a revised pull request soon.

dwt:
Please take care when pulling, I have rebased this branch onto the current master to be able to respect the design changes that have happened there.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Markdown/Markdown.ts
```diff
@@ -43,8 +43,21 @@ export var inlineEditableMarkdownDirective = (
...
         link: (scope) => {
             scope.isEditing = false;
+            // Sorry for the hubhub with the watches here, but the template has multiple levels of child 
+            // scopes (due to nested ng-if's), so we need a way to ensure that we can get a notification back
+            // here when the value is changed deep inside. data = {} is that way.
+            // REFACT would be nice to enhance scope with $bind()...
+            scope.data = { parsetext: scope.parsetext };
             scope.$watch("parsetext", (parsetext) => {
-                scope.isEmpty = parsetext.trim().length === 0;
+                scope.isEmpty = 0 === parsetext.trim().length;
```
der-john: what is this change's purpose?

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_user_profile.scss
```diff
@@ -134,5 +134,14 @@
 
 
 .user-description {
-    margin-top: -4.5rem;
+    // counteract margin bottom from the username
+    @include rem(margin-top, -72px);
```
der-john: personally, I think you overdid this margin.

src/adhocracy_frontend/adhocracy_frontend/static/i18n/core_de.json
```diff
@@ -12,6 +12,10 @@
     "TR__BADGE_ASSIGNMENT_UPDATED": "Etiketten wurden aktualisiert.",
     "TR__BY": "von",
     "TR__CANCEL": "abbrechen",
+	"TR__EDIT_MARKDOWN": "{{ title }} bearbeiten",
+	"TR__SAVE_MARKDOWN": "{{ title }} sichern",
```
der-john: "save" is "speichern" elsewhere.

der-john:
great work! looks good to me & works. feel free to merge.

nidico:
Very neat!

xi:
The inline editing works nicely! I still have some issues though:

-  centered text in a textarea looks weird
-  markdown should not be allowed in the short description
- there should be a length limitation
- allowing to set two different descriptions may make sense in paas, but not anywhere else.
- I can turn the text into a textarea by clicking on it even without the required permissions.

der-john:
"Kurzbeschreibung" is not translated to English yet.

xi:
> "Kurzbeschreibung" is not translated to English yet.

probably a duplicate of https://github.com/liqd/adhocracy3/issues/2614

PR #2562, opened 2016-08-01
fix resourc header image modifier

xi:
current master:

2016-08-01-152155_1440x900_scrot

fixed:

2016-08-01-152129_1440x900_scrot

I did not find a better solution than to add an extra wrapper div. Maybe we can refactor this later.


der-john:
looks alright & works.

PR #2563, opened 2016-08-01
Allow everyone to view image files

### LINKED ISSUE #1973: 403 response on assets in protected process
If a process is in draft state and the creator wants to create some content that contains an image, the image can not be displayed (403 forbidden). The reason is that the request is not triggered by JavaScript but by a `` element, so the authentication headers are missing.

2e2a:
Fixes #1973 

To make it more difficult to view all images through the API, image resource names are now randomly generated.


2e2a:
@xi, @der-john 
Can you confirm that changing the image auto-naming to random does not affect the frontend.

xi:
I checked out this branch, ran migrations and everything still worked as expected (testing with a public process though)

PR #2565, opened 2016-08-01
refactor css

xi:
Sorry for this messy pull request. The commits are mostly unrelated, but I did not want to create 5 separate pull requests.

The first few commits deal with removing irrelevant styling. c4ea980 is interesting because it applies a consistent padding in many places. The rest is some general cleanups.


der-john:
looks good & works.

PR #2571, opened 2016-08-02
replace cancel actions with breadcrumbs

der-john:
This is part of the redesign. Breadcrumbs get introduced - in this PR, (solitary) cancel actions everywhere outside of Mercator get replaced by such 'cancel' breadcrumbs.

Besides this: breadcrumbs plus dropdown now constitute `resource-navigation` - they get renamed in the first commit. The navigation will move outside of the `resource-header` asap.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/DebateWorkbench/DebateWorkbench.ts
```diff
@@ -53,7 +53,8 @@ export var documentDetailColumnDirective = (
 
 export var documentCreateColumnDirective = (
     adhConfig : AdhConfig.IService,
-    adhTopLevelState : AdhTopLevelState.Service
+    adhTopLevelState : AdhTopLevelState.Service,
+    adhResourceUrl
```
xi: This is not required, or am I missing something?

der-john: I introduced it in the template.

xi: Then it is required in as a module dependency in Module.ts, but it is not necessary to inject it here.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Image/Upload.html
```diff
@@ -1,3 +1,8 @@
+<div class="resource-navigation">
+    <div class="resource-navigation-breadcrumbs">
+        <a data-ng-href="{{cancelUrl}}">{{"TR__CANCEL" | translate}}</a>
+    </div>
+</div>
```
xi: This directive is also used on the user profile and was not meant to be used as a page on its own (the solution for euth is a hack really).

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_resource_header.scss
```diff
@@ -45,6 +45,7 @@ and a primary action (e.g. create proposal or like).
     color: $color-text-introvert;
     font-weight: bold;
     list-style: none;
+    padding: 0;
```
der-john: (unrelated.)

xi:
The cancel action uses `goToCameFrom`. This behavior should be kept.

der-john:
this is being rebased a little.

PR #2575, opened 2016-08-03
paas colors

xi:


der-john:
works & looks alright - in my opinion, the "human-perceived" difference between buttons in focus and out of focus is too small. but #2576 changes something.

PR #2579, opened 2016-08-03
Consider local roles when getting assignable badges

### LINKED ISSUE #2535: locally, badge assignment modal has no badges
**Weird**: This is only an issue on local installations and on the staging server, not on the production server.
**Weird**: This is only an issue for non-god users.

I built meinberlin (or euth) idea collection with badges, logged in as "user" and created a proposal.

The option "Badges" appears in the UI. After I click on it, the regular modal dialog appears. But instead of displaying the four available badges, it displays none.

2e2a:
Fixes #2535


joka:
lgtm

joka:
lgtm

PR #2580, opened 2016-08-03
paas style fixups

xi:
Some random bits that popped up while talking to @nicoladimarco

### DIFF
--- a/src/s1/s1/static/js/Packages/S1/Workbench/ProposalEditColumn.html
+++ b/src/s1/s1/static/js/Packages/S1/Workbench/ProposalEditColumn.html
@@ -1,4 +1,11 @@
 <div data-ng-switch="transclusionId">
+    <div data-ng-switch-when="menu">
+        <div class="moving-column-menu-nav">
+            <a
+                class="moving-column-menu-nav-button"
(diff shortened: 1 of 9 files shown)


der-john:
looks great & works.

src/s1/s1/static/js/Packages/S1/Workbench/ProposalEditColumn.html
```diff
@@ -3,7 +3,7 @@
         <div class="moving-column-menu-nav">
             <a
                 class="moving-column-menu-nav-button"
-                data-ng-href="{{ processUrl | adhResourceUrl:meeting }}">{{ 'TR__CLOSE' | translate }} X</a>
+                data-ng-href="{{ proposalUrl | adhResourceUrl }}">{{ 'TR__CLOSE' | translate }} X</a>
```
xi: I don't understand this change. The close button is supposed to close the column.

der-john: I thought we wanted the same behavior of "cancel" in other proposal-edit-view, e.g. of IdeaColl. But the wording and style are "close", so I should revert this.

PR #2590, opened 2016-08-08
Fix decision date in S1

### LINKED ISSUE #2587: S1 differentiation between archive and current session does not work
In the transition to propose the result start_date should not be cleared as it leads to a query with decision_date=null and an error.

After this all proposals are show under the current session and none under archive.

This error seems to occur not often, as the date is cached in the frontend and not removed, when it is deleted in the backend.

2e2a:
Do not remove the start date of the result phase when transiting to participate.

Fixes #2587


joka:
lgtm

PR #2594, opened 2016-08-15
Some changes to user description

xi:
These are some improvements to #2556. I am not sure if all agree that these are actual improvements though.


dwt:
lgtm

PR #2598, opened 2016-08-16
fix core build

xi:
`bin/buildout -c buildout-core.cfg` failed with the following error:

```
While:
  Installing.
  Getting section frontend_development.ini.
  Initializing section frontend_development.ini.
Error: Referenced option does not exist: adhocracy redirect_url
```


2e2a:
lgtm

PR #2600, opened 2016-08-16
new rate design

xi:
This implements a new rate design across all projects. All theming is removed.

A classic issue with the rate widget is that you want to show a lot of different information:

1.  How popular is this issue (total number of rates)?
2.  What is the result (percentage of pro rates)?
3.  Did I rate, and if so, how?

Especially looking at the first two items, we have tried a lot:

1.  Display pro minus contra. This way you get a quick idea about both things (e.g. +100 -> popular and a lot of pro) but can also be ambiguous (e.g. +2 -> not popular or many pro and many contra rates?). [this is what we did before]
2.  Display pro and contra. This way you have all the information, but it is up to you to interpret them. [this is what we are doing now]
3.  Display the ratio. This way only the second aspect is covered. [this is what we do in the opinion component]

Additionally, you may display additional information in a less prominent way, e.g. on hover or just in smaller print. [this is what we were doing before with different rate modifieres and also with the opinion component]

By choosing the pure second option and making it consistent across themes we get a more straight forward implementation and also create a simpler and more consistent user experience. Note that this design is not only used in adhocracy, but also in https://github.com/liqd/euth_wagtail/pull/170.


der-john:
looks great in meinberlin as well as s1.

der-john:
works.

xi:
> looks great in meinberlin as well as s1.

image

There is definitely a difference in height but as the comments may change soon we will just leave it as it is.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_rate.scss
```diff
@@ -172,6 +172,7 @@ The whole widget may have the `is-disabled` state.
     &.is-active {
         color: $color-text-inverted;
         text-decoration: none;
+        font-weight: $font-weight-extrovert;
```
xi: This made the button's size change on hover, which was really annoying. After talking to our designer I reverted it in 7c511ba5976642c26b8c0fc31c25a28f3a70ce34.

PR #2602, opened 2016-08-16
fix paas getMeeting

xi:
steps to reproduce:

- create a proposal
- step through one full cycle

expected result:

if you go to the archive and click on the proposal, you are still in the
archive

actual result:

you are back in current meeting


der-john:
looks good & works.

PR #2604, opened 2016-08-16
anonymity fe for comments + ideaColl proposals

xi:
This is a proposal on how to do the anonymity frontend.

Requirements:

- There are actions that can not be done anonymously
- There are actions that can only be done anonymously
- (users should be able to set a default in their profile settings) NTH
- users should be able to overwrite their default setting at least on some actions (e.g. not anonymous by default, but this comment should be posted anonymously)
- with each action it should be clear whether it is performed anonymously or not
- the complete feature (anonymity) can be enabled via settings

Urgent requirements:

- users should be able to overwrite their default setting at least on comments and S1 proposals
- the complete feature (anonymity) can be enabled via settings

My approach is to display an additional checkbox in form footers. This acts as both a control and an indicator. The user default setting is not yet implemented, but it should be easy to add it later on. There are also no checks yet on whether it is possible to perform an action anonymously, as the API for that is not fixed yet.

An important issue is that this approach only works for forms. Any action that does not have a form (e.g. rate, like) is not covered. It may use the user default, but there is no indication yet.

**How to try out this branch:**
Merge master (resolve the conflicts conform to this branch).
Add the following line to `etc/frontend_development.ini.in`.
```adhocracy.anonymize.enabled = true```
Build an idea_collection and add either a proposal or a comment.


joka:
Why do you rely on a default setting in the user profile? this is not part of any user story.

xi:
> Why do you rely on a default setting in the user profile? this is not part of any user story.

You are right, this is no strict requirement right now. However, I think this would be nice and it was also discussed in the design process.

der-john:
We're having the following backend-related problem. @2e2a will look into it.
> XMLHttpRequest cannot load http://localhost:6541/batch. Request header field X-Anonymize is not allowed by Access-Control-Allow-Headers in preflight response.

der-john:
works such of  good (= so was von gut).

Issue #2606, opened 2016-08-16
User cannot assign badges to idea collection


2e2a:

2e2a:
Fixed by #2607

PR #2607, opened 2016-08-16
consistent assignable badges

xi:
related to #2606 

There are two places where we check whether a user can assign badges:

-  in the badge assignment UI (which badges can be assigned?)
-  in the resource action/dropdown (should we link to the badge assignment UI at all?)

While the badge assignment UI checks the OPTIONS response for assignable badges, the resource actions/dropdown did check for (1) the existence of badges at most 4 levels below the process resource and (2) PUT permissions on the assignment pool.

This PR changes that by using the first approach (OPTIONS) also with the resource action/dropdown, potentially fixing related bugs.


2e2a:
Tested successfully

PR #2617, opened 2016-08-18
S1: Show correct meeting in left column for older proposals

dwt:
This is a followup to https://github.com/liqd/adhocracy3/pull/2590. After that fix old proposals rendered in the wrong column and deep links to archived proposals would load the current column on the left.

### DIFF
--- a/src/s1/s1/static/js/Packages/S1/Workbench/Workbench.ts
+++ b/src/s1/s1/static/js/Packages/S1/Workbench/Workbench.ts
@@ -294,7 +294,7 @@ var getMeeting = (proposal : RIProposal, process : RIS1Process) => {
         var proposalDecisionDate = AdhUtil.deepPluck(
             AdhProcess.getStateData(proposal.data[SIWorkflowAssignment.nick], proposalState), ["start_date"]);
 
-        return (processDecisionDate === proposalDecisionDate) ? "current" : "archive";
+        return (proposalDecisionDate <= processDecisionDate) ? "archive" : "current";
     }
 };
 


xi:
This is very similar to #2602. I will try to find out which one is better.

xi:
Just from manual testing, I find that both this and #2602 fix the issue. However, this one introduces a regression where proposals from the current meeting will be displayed in the archive in result phase.

My proposal is to merge #2602 as it does not break anything. If it is not sufficient we can continue working on this one.

dwt:
> Am 22.08.2016 um 14:59 schrieb Tobias Bengfort <notifications@github.com>:
> 
> Just from manual testing, I find that both this and #2602 <https://github.com/liqd/adhocracy3/pull/2602> fix the issue. However, this one introduces a regression where proposals from the current meeting will be displayed in the archive in result phase.
> 
> My proposal is to merge #2602 <https://github.com/liqd/adhocracy3/pull/2602> as it does not break anything. If it is not sufficient we can continue working on this one.
> 
> 
Be my guest in using the other fix. I thought I did test all the phases - but quite probably overlooked something.

Thanks!

Martin

src/s1/s1/static/js/Packages/S1/Workbench/Workbench.ts
```diff
@@ -294,7 +294,7 @@ var getMeeting = (proposal : RIProposal, process : RIS1Process) => {
         var proposalDecisionDate = AdhUtil.deepPluck(
             AdhProcess.getStateData(proposal.data[SIWorkflowAssignment.nick], proposalState), ["start_date"]);
 
-        return (processDecisionDate === proposalDecisionDate) ? "current" : "archive";
+        return (proposalDecisionDate <= processDecisionDate) ? "archive" : "current";
```
xi: Just so we do not forget it. IIRC, these values are strings, so comparison might not work as expected.

dwt: > Am 24.08.2016 um 10:51 schrieb Tobias Bengfort <notifications@github.com>:
> 
> In src/s1/s1/static/js/Packages/S1/Workbench/Workbench.ts <https://github.com/liqd/adhocracy3/pull/2617#discussion_r76017832>:
> 
> > @@ -294,7 +294,7 @@ var getMeeting = (proposal : RIProposal, process : RIS1Process) => {
> >          var proposalDecisionDate = AdhUtil.deepPluck(
> >              AdhProcess.getStateData(proposal.data[SIWorkflowAssignment.nick], proposalState), ["start_date"]);
> >  
> > -        return (processDecisionDate === proposalDecisionDate) ? "current" : "archive";
> > +        return (proposalDecisionDate <= processDecisionDate) ? "archive" : "current";
> Just so we do not forget it. IIRC, these values are strings, so comparison might not work as expected
> 
Actually on rfc formatted date strings it does. Thats precisely the reason they are formatted the way they are. Or am I missing something?

dwt:
I checked the other fix - it seems to work nicely. Closing this now.

PR #2618, opened 2016-08-18
Better wording for the show comments functionality.

dwt:
User testing for S1 has shown that people miss the fact that they can show the comments column because they think that this is just an indicator how many comments are available. (Reasoning: It looks the same as other comment indicators, but it behaves different).

Thus the resolution: Make it look different too, to give people a reason to suspect that it does something. :)

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/i18n/core_de.json
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/i18n/core_de.json
@@ -17,6 +17,7 @@
     "TR__CHOOSE_BADGES": "Wähle ein oder mehrere Etiketten",
     "TR__CLOSE": "schließen",
     "TR__COMMENTS": "Kommentare",
+    "TR__COMMENTS_SHOW": "{{ number }} Kommentar zeigen",
     "TR__COMMENTS_TOTAL": "Kommentare (gesamt)",
     "TR__COMMENT_EMPTY_TEXT": "Noch keine Kommentare.",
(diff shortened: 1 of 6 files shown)


dwt:
When considering to merge, you guys might want to consider to extend this change to the other adh projects. I didn't feel empowered to decide this on my own.

xi:
> When considering to merge, you guys might want to consider to extend this change to the other adh projects. I didn't feel empowered to decide this on my own.

You already implemented this for almost everything, right??

xi:
image

I talked to our designer and we agreed that this is not the best approach. The link you changed is primarily information. It is only a link because -- why not? The primary way to get to the comments should be the icon in the bottom right. So we should change that instead.

dwt:
Well, I don't mind the solution - but right now the comments functionality is not self explaining and discoverable at all.

I'd say go with the text based change for now, till a better solution is built and then reverse this (smallish) change.

der-john:
I'm sorry, @dwt, but I don't think we all approve. I just talked to Petra and she agreed with Nico and @xi, because the link isn't just for showing comments, but for writing them and reporting them, too.

But I positively agree with you that we should take your user testing into consideration. In his most recent clickdummy on meinberlin's design, Nico once again suggested to list comment rights underneath the proposal itself, cf. https://marvelapp.com/1gc2d4i#14210116. (Furthermore, I believe in @stadine's design there was a call-to-action button saying "Comment on this proposal" inside the resource header.)

I would take your case as an argument to relocate the comments asap.

PR #2586, opened 2016-08-04
fix paas buttons

der-john:
These changes make it easier for the user to know whether he/she's currently looking at an 'archive' proposal list, a 'current meeting' proposal list, or a 'next meeting' proposal list. (Partially reverts #2576.)

PR #2619, opened 2016-08-18
Restore highlighting behavior of the meeting selector.

dwt:
This was lost sometimes in the recent past, which led to the problem that users have no way to determine which meeting is currently active. It’s not so nice that the ‚create proposal button‘ shows the same way - but at least users know again what meeting is displayed.

I'm open to proposals on how to differentiate the 'create proposal' button better.


xi:
Related to #2586

xi:
Thanks for fixing this!

Some background: I had removed the CSS code in #2576, but IIRC it was already broken before that.

PR #2623, opened 2016-08-22
s/votable/voteable/g

xi:
I think "votable" is actually the more common spelling, but in our code it was only used in 4 places, 3 of them documentation.


2e2a:
lgtm

PR #2626, opened 2016-08-22
fix edit action permissions

xi:
I noticed (in s1) that sometimes the edit action shows up if I am not allowed to edit. The reason is most likely that `options.PUT` is only a valid check for simples. For versions, we have to check POST on the item.

Note: While thinking about this I had a second idea that might be worth investigating: Checking `options.PUT` might not actually be sufficient for simples because this may also just be `options.hide`, i.e. you can only put `SIMetadata.hidden = true`.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
@@ -96,10 +96,19 @@ export var resourceDropdownDirective = (
 
             scope.modals = new Modals($timeout);
             adhPermissions.bindScope(scope, () => scope.resourcePath, "options");
+            adhPermissions.bindScope(scope, () => scope.itemPath, "itemOptions");
 
             scope.$watch("resourcePath", () => {
                 scope.modals.clear();
             });
+
+            scope.canEdit = () => {
+                if (scope.resourcePath === scope.itemPath) {
+                    return scope.options.PUT;
+                } else {
+                    return scope.itemOptions.POST;
+                }
+            };
 
             scope.toggleDropdown = () => {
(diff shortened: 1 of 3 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
```diff
@@ -95,7 +95,7 @@ export var resourceDropdownDirective = (
             };
 
             scope.modals = new Modals($timeout);
-            adhPermissions.bindScope(scope, scope.resourcePath, "options");
+            adhPermissions.bindScope(scope, () => scope.resourcePath, "options");
```
der-john: thanks. this is unrelated, though, right?

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/ResourceActions/ResourceActions.ts
```diff
@@ -163,6 +126,19 @@ export var resourceDropdownDirective = (
...
+export var resourceActionsDirective = (
+    $timeout : angular.ITimeoutService,
+    adhConfig : AdhConfig.IService,
+    adhPermissions : AdhPermissions.Service
+) => {
+    var directive = resourceDropdownDirective($timeout, adhConfig, adhPermissions);
+    directive.scope["createDocument"] = "=?";
+    directive.templateUrl = adhConfig.pkg_path + pkgLocation + "/ResourceActions.html";
+    return directive;
+};
```
der-john: nice.

der-john:
looks good & works.

> Note: While thinking about this I had a second idea that might be worth investigating: Checking options.PUT might not actually be sufficient for simples because this may also just be options.hide, i.e. you can only put SIMetadata.hidden = true.

I'm opening an issue on that...

PR #2627, opened 2016-08-23
allow to anonymise rates

joka:
add IAllowAddAnonymized sheet to rates service

add helper function get_system_user_anoymous

modify rate sheet validators to ensure "uniqueness" with anonymized  rates:

- validate rate subject is always the authenticated user or the anonymous
system user
- replace subject with anonymous user if anonymize request
- check anonymized_creator when validating "uniqueness" of rates


joka:
A functional test is missing, but manual testing with the rest_api.rst use case worked fine

2e2a:
lgtm

PR #2610, opened 2016-08-17
user settings

xi:
This implements a basic user settings UI. You can currently edit username and password.

The edit page is not yet linked from the profile page as this requires #2588.

It might be a good idea to improve the UI, e.g. a second password field to avoid typos (like to one in the registration form).

PR #2631, opened 2016-08-23
anonymity default in user settings

xi:
This adds the anonymize default setting from #2615 to the user settings page from #2610. It already uses the `anonymize_enabled` config from (and thus actually needs) #2604.

Note that this default value is not yet applied to the use cases in #2604.


der-john:
looks good & works.

PR #2636, opened 2016-08-24
anonymize frontend abstraction

xi:
As an update to #2604, this adds some abstractions and some missing features, i.e. permission check and user default.


der-john:
looks fantastic & works.

PR #2641, opened 2016-08-25
more anonymity fe

der-john:
- link to anonymous' user profile is eliminated
- S1 proposals can be posted anonymously
- rates can be posted anonymously\*
- on editing a resource, preserve anonymity setting (i.e. "by anonymous" or "by user") of this resource per default - this setting is called `localDefault`
- use `localDefault` in ideaColl proposals + S1 proposals + comments

\*Note that, except for S1, there is no place in the UI to show which user has rated in which way, anyways. In S1, there is the `user-activity-overview`, that currently informs on *how many* rates a user has casted, not where.


der-john:
the password repeat from #2649 doesn't work properly. I'll fix it soon.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Anonymize/Anonymize.ts
```diff
@@ -46,6 +47,7 @@ export var anonymizeDirective = (
                 ]).then((args) => {
                     scope.isOptional = args[1];
                     scope.data.model = scope.isOptional ? adhUser.data.anonymize : false;
+                    scope.data.model = scope.localDefault || scope.data.model;
```
xi: IMHO this is not exactly what we want. I see two issues:

-  **`isOptional` has no effect on `localDefault`** (If `isOptional` is false and `localDefault` is true, the result will be true; I would expect false)
-  **`localDefault=false` is ignored** (If `localDefault` is false and `adhUser.data.anonymize` is true, the result will be true; I would expect false)

I propose something like this:

```
var default = typeof scope.localDefault === "undefined" ? adhUser.data.anonymize : scope.localDefault;
scope.data.model = scope.isOptional ? default : false;

File: src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Http.ts [hunk 2]
```diff
@@ -418,7 +418,8 @@ export class Service {
     public postNewVersionNoFork(
         oldVersionPath : string,
         obj : any,
-        rootVersions? : string[]
+        rootVersions? : string[],
+        config? : IHttpOptionsConfig
```
xi: Should be `IHttpConfig`

xi:
I did not look into 8de064c, because I will do a review of anonymity in rates in relation to #2645 separatly.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Anonymize/Anonymize.ts
der-john: I totally agree.

der-john: funny. trying out your code literally produced strange errors because `default` is a keyword.

der-john: fixed in #2703.

PR #2642, opened 2016-08-25
Gardening

dwt:
The use of this file was removed when the functionality was inlined into the user detail pane.


der-john:
thanks! looks good & works.

PR #2647, opened 2016-08-29
adjust user activity overview

### LINKED ISSUE #2645: anonymization doesn't harmonize with user activity overview
I build S1.
I log in as participant.
I set my user anonymity default to `true`.
On the user activity overview, I see that participant has rated `n` times.
I rate a fresh proposal (a proposal I haven't rated before).

Then I revisit the user activity overview.
**Expected result:** I see that participant has rated `n` times.
**Actual result:** I see that participant has rated `n+1` times.

The cause seems to be the following: 
- by rating anonymously, an anonymous new version is created - the rate item itself lists participant as the creator. (That's what I see in meinberlin.)
- the user activity (in `src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts:890`) queries the backend for `RIRate`s with participant as their creator. Thus, the rate count increases. (As far as I know, the frontend cannot easily find out which rate version is the newest, as rate versions do not get the "LAST" tag.)

der-john:
*Fixes #2645*.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Views.ts
@@ -16,9 +16,9 @@ import * as AdhUtil from "../Util/Util";
 import * as AdhCredentials from "./Credentials";
 import * as AdhUser from "./User";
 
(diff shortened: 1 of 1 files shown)


joka:
works :)  i merge this to make it part of the new release tomorrow

PR #2648, opened 2016-08-29
check commentability of resources

### LINKED ISSUE #2644: Comment UI doesn't check permissions
If commenting is not allowed, the UI is shown nevertheless, e.g. on a rejected S1 proposal.

der-john:
*Fixes #2644*.


xi:
There is in fact a permission check: in the listing. We should investigate why it did not work.

PR #2650, opened 2016-08-29
Ensure anonymity options for items and item version are the same fix 2634

joka:
…

### DIFF
--- a/src/adhocracy_core/adhocracy_core/content/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/content/__init__.py
@@ -13,6 +13,7 @@ from zope.interface.interfaces import IInterface
 
 from adhocracy_core.authentication import is_created_anonymized
 from adhocracy_core.exceptions import RuntimeConfigurationError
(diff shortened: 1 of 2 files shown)


src/adhocracy_core/adhocracy_core/content/__init__.py
```diff
@@ -307,7 +308,11 @@ def get_workflow(self, context: object) -> IWorkflow:
 
     def can_add_anonymized(self, context: object, request: Request) -> bool:
         """Check if children can be created/added anonymized to `context`."""
-        has_sheet = IAllowAddAnonymized.providedBy(context)
+        if IItem.providedBy(context):
+            # if item was anonymized you can also add versions anonymized
+            has_sheet = _is_anonymized_and_has_permission(context, request)
```
2e2a: Shouldn't the check be for ItemVersion and the call for the parent item?

has_sheet is misleading at that point and the permissions check is also done two times.

2e2a:
Created #2686 to implement comments.

PR #2651, opened 2016-08-30
iconlos durch die nacht

der-john:
A special solution for browsers that don't download icons or logos.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_action_section.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_action_section.scss
@@ -68,6 +68,7 @@ States:
 
     .action-section-button {
         @include action-section-button;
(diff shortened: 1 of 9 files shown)


dwt:
This is a good start, but I would recommend to remap the current font icon code points to the ones used here.

The idea being that if the icon font cannot be loaded, then at least the browser can fall back to standard unicode icons.

To explain: AFAIK we are currently using private unicode code points to prevent any collisions with existing fonts in case the font is not loaded. Unfortunately this means that in case the font doesn't load (due to windows security settings for example) - no icons are shown.

If we use semantic unicode code points to host our icons, then if the font doesn't load, browsers will fall back to these. Not as pretty, but way better than showing no icons at all.

To start in this direction, the css file that includes the custom font icons can be replaced / enhanced to use those semantic unicode code points. This would concentrate the change to one file in the beginning. Then a remapping of the self generated icon fonts can take place to use those code points as the base. (But this can happen later).

PR #2652, opened 2016-08-30
iconlos durch den tag

der-john:
For browsers that don't download icons or logos.

(Replaces #2651.)

Except for b372d78, these commits can be merged into master...

### DIFF
--- a/src/s1/s1/static/js/Packages/TopLevelState/templates/defaultHeader.html
+++ b/src/s1/s1/static/js/Packages/TopLevelState/templates/defaultHeader.html
@@ -1,7 +1,7 @@
 <header>
     <a class="s1-header-logo-link" data-ng-href="/">
         <abbr title="Participation as a Service">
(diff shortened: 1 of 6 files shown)


src/s1/s1/static/js/Packages/TopLevelState/templates/defaultHeader.html
```diff
@@ -1,7 +1,7 @@
 <header>
     <a class="s1-header-logo-link" data-ng-href="/">
         <abbr title="Participation as a Service">
-            <img class="s1-header-logo" alt="PAAS" src="/static/images/PAAS logo.png" />
+            <h1 title="Participation as a Service"><span style="color: #c3302a; font-weight: bold;">P&#7468;</span><span style="color: #241b70;; font-weight: bold;">&#7468;S</span></h1>
```
dwt: Why replace the image? Images worked at the client, right?

PR #2657, opened 2016-08-31
require password repeat

der-john:
continues #2649.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Edit.html
```diff
@@ -63,8 +63,9 @@
 	        <input
 	            type="submit"
 	            name="submit"
-	            value="{{ 'TR__PUBLISH' | translate }}"
-	            class="button-cta form-footer-button-cta" />
+	            value="{{ 'TR__ACTIVATE' | translate }}"
+	            class="button-cta form-footer-button-cta"
+                data-ng-disabled="data.password.$dirty && data.password !== data.passwordRepeat" />
```
dwt: This also contains the rename from publish to activate -> which is already done in https://github.com/liqd/adhocracy3/pull/2665

der-john: done.

PR #2659, opened 2016-08-31
 change s1 landing page texts

der-john:


src/s1/s1/static/js/Packages/S1/Workbench/Landing.html
```diff
@@ -8,7 +8,17 @@ <h1 class="s1-landing-section-header">{{ processTitle }}</h1>
 
     <section class="s1-landing-section">
         <h1 class="s1-landing-section-header">{{ "TR__S1_INTRODUCTION_HEADER" | translate }}</h1>
-        <p>{{ "TR__S1_INTRODUCTION_TEXT" | translate }}</p>
+        <adh-parse-markdown data-parsetext="introText"></adh-parse-markdown>
```
xi: I would expect that `data-parsetext="introText | translate"` works fine. Did you try it?

PR #2661, opened 2016-08-31
try out non-empty header

der-john:


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Http/Http.ts
```diff
@@ -131,7 +131,7 @@ export class Service {
         }
         if (config.anonymize) {
             _.assign(headers, {
-                "X-Anonymize": ""
+                "X-Anonymize": "X-Anonymize"
             });
```
joka: a comment why this is needed would be helpfull

der-john:
works.

PR #2663, opened 2016-08-31
change name of 'anonymous' in S1

der-john:


joka:
cool, works

joka:
Do we have to update transiflex?

PR #2665, opened 2016-09-01
Reword the save button for user profile

dwt:
because you most likely want to ‚save‘ your password instead of  ‚publish‘-ing it.


der-john:
looks good.

PR #2666, opened 2016-09-01
deactivate messaging in S1

der-john:


xi:
Note: there is already a `ngIf` on the containing `<div>`

PR #2668, opened 2016-09-01
EMU: add sentence with infolink to landing page

der-john:


xi:
- the english version needs to be adapted
- "click here" is bad practice for link texts (see [wcag](https://www.w3.org/TR/WCAG/#navigation-mechanisms-refs) and [nngroup](https://www.nngroup.com/articles/writing-links/) on this topic)

PR #2673, opened 2016-09-06
change date of transition from phase A to B

der-john:
on request of Jana.


xi:
english translation need to be adapted

xi:
Somehow this is exactly the same as #2669

PR #2680, opened 2016-09-09
Make process description dynamic

dwt:
Because we want users / admins to be able to change it via the SDI instead of having to deploy a new version for every change. (This also makes it much easier for them to document when the next process switch will happen).

@joka / @2e2a  Can you please check how we defaulted the short_description/ description? That is quite probably wrong / easy to improve. :)


xi:
I merged master to fix the frontend conflicts. I did not check the desciption text tough.

dwt:
Description text is what it was before it was changed to be Stendal specific - so it should be fine. Even though, it is just a default, to be overridden with something instance specific, like we did with stendal.

When this is deployed, the stendal text will have to be inserted into the description - I can do this on short notice, or in collaboration with someone from liquid immediately by hot patching on the server.

dwt:
The markdown version is available from me on request, as I don't want to post it here if possible. (yes I know, it's currently in code, but it shouldn't be there, so that's why I don't want to post it at even more locations).

2e2a:
Moved the default process description to sheet configuration.

xi:
LGM

PR #2678, opened 2016-09-08
demonstration for a3

der-john:
this creates a new fe package called `demo`.

also, a new logo is installed in the header.

PR #2684, opened 2016-09-12
demonstration for a3

der-john:
(continues #2678.)

this creates a `demo` project and gives it a special header. other projects consist of fe and be package, but demo just has its own fe and uses the meinberlin be.


xi:
I am strongly against duplicating the meinberlin code. We should find a way to share that code instead.

der-john:
I don't like duplicating code, either. But I'm currently in favor of merging this PR and reorganizing/refactoring the code afterwards. In particular, I think, like `IdeaCollection`, the bulk of `Stadtforum` should be in `adhocracy_frontend`, while only a `Module.ts` need remain in `meinberlin` and `demo`.

Bearing this in mind, I don't think we risk comitting many errors through duplicated code.

To illustrate this fact, I made an analysis of what the >2100 extra lines of code in this PR consist of:
- 278 lines: the `Stadtforum` module*,
- 192 lines: various `Module.ts` of `meinberlin` workbenches other than `Stadtforum`**,
- 226 lines: `Adhocracy.ts`**,
- 271 lines: i18n***,
- 661 lines: `LICENSE.txt`**,
- 553 lines: scaffolding python code**.

* is going to (almost) disappear with refactoring.
** nobody is going to touch that code anyways.
*** if somebody touches that, then it's on purpose and only influences `demo`.

der-john:
After considering the matter thoroughly, @xi and I decided to do something different.

der-john:
has been replaced by #2694.

PR #2685, opened 2016-09-12
refactor #2666

xi:
#2666 removed the messaging UI from the frontend for s1. IMHO it would be a better approach to deny messaging in the backend. The frontend will (hopefully) adapt automatically.

@2e2a could you please adjust the permissions in the backend?


der-john:
the permission that @2e2a set isn't correctly checked by the frontend.

der-john:
looks good & works.

PR #2687, opened 2016-09-12
fix/refactor password repeat

xi:
This is a fixup to ab2869c (#2649). See also https://github.com/liqd/adhocracy3/pull/2641#issuecomment-243478907

I fixed the validation of the password repeat in the user settings. I also ported the approach I used to the password repeat in the registration directive. Unfortunately I had to use a scoping hack there because the form was only available in a nested scope.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Edit.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Edit.html
@@ -45,9 +45,8 @@
         <input
             type="password"
             name="password_repeat"
-            data-ng-model="data.passwordRepeat"
-            data-ng-class="{'ng-invalid': (input.password !== input.passwordRepeat) && userEditForm.password_repeat.$dirty}"/>
-        <span class="input-error" data-ng-show="(input.password !== input.passwordRepeat) && userEditForm.password_repeat.$dirty">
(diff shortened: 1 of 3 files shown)


xi:
I just realized that this duplicates #2657. Sorry.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/User/Edit.html
```diff
@@ -45,9 +45,8 @@
         <input
             type="password"
             name="password_repeat"
-            data-ng-model="data.passwordRepeat"
-            data-ng-class="{'ng-invalid': (input.password !== input.passwordRepeat) && userEditForm.password_repeat.$dirty}"/>
-        <span class="input-error" data-ng-show="(input.password !== input.passwordRepeat) && userEditForm.password_repeat.$dirty">
+            data-ng-model="data.passwordRepeat" />
```
der-john: why did you eliminate`ng-invalid`?

xi: No, in this one.  See https://github.com/liqd/adhocracy3/pull/2687/commits/01ca41e54b403efc307e1a84e6edf35d6c0b55e7#diff-38f88590d3a2d5e444e2bd4710f305ffR776

PR #2688, opened 2016-09-13
ideaColl: fix checks

der-john:
the current checks weren't sufficient - some minor fe errors occurred, because even if a process *in principal* could have a `location`, *in fact* it sometimes didn't. (`processProperties.hasLocation` was the only check used.)


xi:
I first thought that not specifying a location was an error. In that case this PR would hide issues instead of fixing them. But I was actually wrong. Nice fix!

PR #2691, opened 2016-09-13
fix process overview

xi:
The process overview introduced in #2027 is currently broken (404). I debugged the issue and found that `AdhUtil.parentPath("http://example.com/")` will return `"http://"`. I fixed it to properly handle URLs.

(Note that most frontend code uses the terms "path" and "url" synonymously. This is in our refactoring backlog.)


der-john:
looks good & works.

PR #2694, opened 2016-09-14
meinberlin split

xi:
This is an alternative approach to #2684: Instead of duplicating code from meinberlin, it is instead moved to a new package meinberlin_lib which is then used in both meinberlin and the new demo package.

This PR adds about the same amount of lines as #2684. However, this is because the scaffolding (including LICENSE) is added twice. Next to no code is duplicated. Note that `Adhocracy.ts` and `a3.scss` are duplicated, but they are very project specific and are expected to diverge.

**Note**: I was **not** super careful when splitting the packages. We may have to move some more code in the next weeks.


der-john:
great job. looks good & works.

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_action_bar.scss
```diff
@@ -90,6 +90,7 @@ An action can trigger showing a modal below the action bar.
     @include rem(font-size, $font-size-small);
     @include rem(padding, 0 $padding);
     @include clearfix;
+    overflow: visible;
```
xi: This actually breaks the clearfix!

PR #2701, opened 2016-09-15
add dynamic info text on top of process overview

der-john:
This PR uses the `description` sheet of the `IRootPool` as an info text on top of the process overview.

The way I see it, there's a huge advantage and a disadvantage to this:

( + ) all of us can manually change the text in SDI
( - ) whatever the current language, the text will remain the same.

@xi and I agreed this disadvantage is bearable, because the content of the processes below the text will be monolingual as well.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Process/Listing.html
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Process/Listing.html
@@ -1,3 +1,6 @@
+<div class="overview-description">
+    <adh-parse-markdown data-parsetext="overviewDesc"></adh-parse-markdown>
+</div>
 <adh-listing data-content-type="{{contentType}}" data-params="params" data-path="/" data-no-create-form="true">
     <adh-process-list-item data-path="{{element}}"></adh-process-list-item>
 </adh-listing>
--- a/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_list_item.scss
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_list_item.scss
@@ -107,3 +107,13 @@ $thumbnail-width: 105px;
 .list-item-badges {
     float: right;
 }
+
+.overview-description {
+    @include container;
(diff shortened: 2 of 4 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Process/Listing.html
```diff
@@ -1,3 +1,6 @@
+<div class="overview-description">
+    <adh-parse-markdown data-parsetext="overviewDesc"></adh-parse-markdown>
+</div>
```
xi: IMHO this should not be part of the listing itself. It should be on top of the listing instead (ResourceArea.html)

src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_list_item.scss
```diff
@@ -107,3 +107,13 @@ $thumbnail-width: 105px;
...
 }
+
+.overview-description {
+    @include container;
+    @include rem(padding, 1.25em $padding);
+    display: block;
+    position: relative;
+    margin-bottom: 1em;
+    background: $color-background-base-introvert;
+}
```
xi: If you create a new component, it must be properly documented.

der-john: it's still changing due or reviews + refactorings. I'll document it later.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Process/Listing.html
der-john: I'll look into it.

der-john: I can't get the resourceArea directive to add a param to its scope.

der-john: @xi and I agree there should be an overview directive that wraps `adh-process-listing` and that the text should be placed there.

PR #2705, opened 2016-09-19
check commentability differently

### LINKED ISSUE #2644: Comment UI doesn't check permissions

der-john:
Like #2648, this fixes #2644, but in a more easy and appropriate way.

See comment on #2648.


xi:
Nice!

PR #2706, opened 2016-09-19
fix map switch

xi:
See https://github.com/liqd/adhocracy3/pull/2694#r79338913

The underlying issue was that the map-filled icon was missing.


der-john:
works. thanks a lot!

PR #2713, opened 2016-09-20
add counter to comment listings

der-john:
needs + extends #2717.

In order to find the "total comment count" generically, I added logic and an http request to `Comment.ts`.

what's bad about this: we add yet another http request.
what's good: the resource may be in the cache anyway + this is generic -> it appears in `pcompass` and `stadtforum`.


der-john:
is being rebased.

PR #2718, opened 2016-09-20
add listing-counter to comments + use totalCount

der-john:
replaces #2713. needs #2717.

In order to find the "total comment count" generically, I added logic and an http request to `Comment.ts`.

what's bad about this: we add yet another http request.
what's good: the resource may be in the cache anyway + this is generic -> it appears in `pcompass` and `stadtforum`.


der-john:
on using `totalCount` (formerly #2721): this is a small fixup of #2717.

`listing-counter`s listed `elements.length`. But if pagination comes into play, `elements.length` is curbed by `currentLimit`.

@xi and I think this change reflects the semantics of the counter better: you want to know how many items of a certain type there are, not how many you are currently seeing on the page.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
```diff
@@ -114,7 +116,8 @@ export class Listing<Container extends ResourcesBase.IResource> {
                 emptyText: "@",
                 // use this to pass custom data to the injected templates
                 custom: "=?",
-                counter: "=?"
+                counter: "=?",
+                counterValue: "@?"
```
xi: This way, `counterValue` will be a string. I think this is confusing.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
```diff
@@ -134,6 +137,8 @@ export class Listing<Container extends ResourcesBase.IResource> {
 
                 $scope.createPath = adhPreliminaryNames.nextPreliminary();
 
+                $scope.noCounterValue = typeof ($scope.counterValue) === "undefined";
```
xi: This is not updated when `$scope.conterValue` changes. What about a function like this:

```js
$scope.getCounterValue = () => {
    return typeof $scope.counterValue !== "undefined" ? $scope.counterValue : $scope.totalCount;
};
```

This would also allow to reduce duplication in the template.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.html
```diff
@@ -34,7 +34,9 @@ <h3 class="sort-header">{{ "TR__SORT_BY" | translate }}</h3>
             </a>
         </div>
     </div>
-    <div class="listing-counter" data-ng-if="counter && elements.length > 0">{{elements.length}} {{contentType | adhResourceName | translate}}</div>
+    <div class="listing-counter" data-ng-if="counter && counterValue > 0">{{counterValue}} {{contentType | adhResourceName | translate}}</div>
+    <div class="listing-counter" data-ng-if="counter && noCounterValue && totalCount == 1">{{totalCount}} {{contentType | adhResourceName | translate}}</div>
+    <div class="listing-counter" data-ng-if="counter && noCounterValue && totalCount > 1">{{totalCount}} {{contentType | adhResourceName:true | translate}}</div>
```
xi: In order to allow automatic pluralization, it would be better to change the adhResourceName filter to accept a number instead of a boolean. Then we could just pass in the count and the filter decides which variant to use.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
der-john: nice idea.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.html
der-john: nice idea.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.ts
der-john: I agree.

xi:
It might be nice if the commentCount would update automatically, just like the count on all other listings. This could be achieved if `counterValue` was a function that is called on each update.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Listing/Listing.html
der-john: > In order to allow automatic pluralization, it would be better to change the adhResourceName filter to accept a number instead of a boolean. Then we could just pass in the count and the filter decides which variant to use.

I'll do that (but in `getNames`, not in `resourceName`).

PR #2709, opened 2016-09-19
integrate ProcessProvider.names into RAProvider.names

der-john:
needs + extends #2702.

PR #2692, opened 2016-09-13
improve process list item

xi:
This is a heavily refactored subset of #2681 that concentrates on improving the list item itself.

Note that 	12a1745 solves the issue of meinberlin-specific code in Process.ts that was mentioned in #2681.

PR #2720, opened 2016-09-20
refactor RA.getName

der-john:
refines #2692. replaces #2709. needs #2717.

as @xi has pointed out, we should prevent implicit circular dependency and move `names` and `getName` to their own service (in a separate PR).


xi:
I think that moving the process listing to `AdhResourceArea` is a bad idea. As I said before, we should move the naming to its own module instead.

Properly reviewing the moving of the process listing is a lot of work, and also unncessary because, as explained before, I think it should be reverted soon anyway. So I propose to rebase this PR instead of merging it and reverting most of it later.

der-john:
All right.

PR #2721, opened 2016-09-20
use totalCount in listing-counter

der-john:
needs #2720. small fixup of #2717.

(`listing-counter`s listed `elements.length`. But if pagination comes into play, `elements.length` is curbed by `currentLimit`.)

@xi and I think this change reflects the semantics of the counter better: you want to know how many items of a certain type there are, not how many you are currently seeing on the page.


der-john:
should be integrated into #2718 to avoid conflicts.

PR #2681, opened 2016-09-09
finally home

der-john:
continues #2675.

> wie kommt man zum home view?

Right now, the home view isn't accessible through the UI yet.

In the near future, in the demo project, the home view will be accessible by doing one of the following:
- clicking on the logo on the top left corner (this is currently being implemented in PR #2678 -> #2694),
- having yourself be redirected, e.g. after account activation (`redirect_url` in the config  #2694),
- breadcrumbs in the process header saying "back to process overview",
- breadcrumbs in the user header saying "back to process overview" (this kind-of fixes #2640).

One more thing: the process overview logic is currently done in the `Process` module - and unfortunately includes some custom `meinberlin` code. @xi and I need to consider where the appropriate place for this logic is.

PR #2722, opened 2016-09-20
Nicer home layout

xi:
Fixup to https://github.com/liqd/adhocracy3/pull/2681#issuecomment-246761946

This also includes moving the home view to its own module and directive.


der-john:
works. looks good except, in my opinion, for 71c579e - I think it looked better before.

PR #2629, opened 2016-08-23
update frontend dependencies

xi:
Mostly minor and patch releases. Grunt had a [major release](http://gruntjs.com/blog/2016-04-04-grunt-1.0.0-released), but it should not affect us.

Things that were not updated:

-   [protractor](https://github.com/angular/protractor/blob/master/CHANGELOG.md)
-   [angular-cache](https://github.com/jmdobry/angular-cache/blob/master/CHANGELOG.md) (see [issue](https://github.com/jmdobry/angular-cache/issues/234))
-   [typescript](https://blogs.msdn.microsoft.com/typescript/2016/07/11/announcing-typescript-2-0-beta/) announced version 2.0-beta some time ago, but the stable version has not yet been released.

PR #2723, opened 2016-09-20
update angular-cache

xi:
I skipped updating angular-cache in #2629 because of an issue. The [upstream bugreport](jmdobry/angular-cache#234) was closed as wontfix, so I am applying a local workaround.


der-john:
works.

PR #2728, opened 2016-09-21
fix meinberlin extract msgs

### LINKED ISSUE #2724: String extraction for meinberlin broken
#2694 broke string extraction / i18n: The translation files are in meinberlin, while the code that contains translatable strings is in meinberlin_lib.

xi:
Fixes #2724 

(the diffs are a bit strange up because I edited very similar files)


der-john:
:( this will have conflicts with #2732.

der-john:
works.

PR #2733, opened 2016-09-21
change / redirect logic

xi:
(see #1965)

- if adhEmbed.initialUrl is set, use that
- if no other option is available, use '/r/' (home, see #2681)

Note that `adhConfig.redirect_url` is set in all current projects, so home is not used in practice yet.


der-john:
definitely an improvement.

PR #2732, opened 2016-09-21
say my name, say my name

der-john:
replaces #2709 & #2720. needs #2717 & #2718.

@xi and I wanted to prevent implicit circular dependency and move `names` and `getName` to their own service. Until now, this logic existed both in `ResourceArea` and `Process`.

PR #2582, opened 2016-08-04
make process header available for idea collection

der-john:
(needs #2573, continues #2577.) Part of the redesign.

! This doesn't do anything unless you set `hasResourceHeader` to `true` in `src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/IdeaCollection/Process/Process.ts`, line 85.

PR #2740, opened 2016-09-22
process headers II

der-john:
continues #2582. extends #2732.

unfortunately, in the debate workbench we now have an empty resource dropdown...


der-john:
this is unnecessarily complicated.

PR #2741, opened 2016-09-22
process headers II

der-john:
continues #2582.

unfortunately, in the debate workbench we now have an empty resource dropdown...


der-john:
I'm merging master and resolving conflicts caused by #2747.

xi:
LGTM.

xi:
I must admit I forgot about the empty resource dropdown. We should find a solution as fast as possible

PR #2747, opened 2016-09-23
Core package

der-john:
@xi and I toyed with the idea of adding all core modules to a package called `Core` (analogous to the `Meinberlin` package). This way, a lot of duplicate code from the numerous `Adhocracy.ts` files could be deleted.

Unfortunately, it's a complicated thing to do. On this branch, I get runtime errors I can't solve. :-1:


xi:
I think this should work now. I did some superficial testing on demo, meinberlin and mercator.

I think it is possible to move a lot more of the code from `Adhocracy.ts` to `Core/Module.ts`. But IMHO we should merge this first step quickly in order to avoid merge conflicts.

PR #2748, opened 2016-09-23
change wording of activation error email

### LINKED ISSUE #971: Activation link semantic is confusing
Hi,

if i create a new user I will receive an activation link via email. If you open that link for the first time you will be logged in. If you open it for the second time the following message will be displayed

```
invalid or expired activation link

To request a new link, please contact our support team, explicitly stating the email address you registered with, via advocate-europe-support@liqd.de.

We will send you a new authentification link within the next 24h. Please check your mail and SPAM meanwhile.

Please excuse the inconvience and make sure you activate your account within the next 6 days.

Thanks!
```

I think this message is very confusing: Why should I now wait for a new activation link, if I just activated my account.

In addition I am also pretty sure that we will never resend the activation link, so it seems we are lying to our user. Also we have a lot of support cases of people claiming that their link was invalid from the beginning. I know from skype chats at least that they issue a requests to all links that are send through them. Maybe there are email providers which do the same? Could that be an issue here?

der-john:
fixes #971


2e2a:
support mail seems a bit out of context with the new wording

2e2a:
Maybe it makes sense to propose the user to try to login, in case the link is used for the second time and the user is already activated.

2e2a:
lgtm

PR #2749, opened 2016-09-26
add s1 resources.json

der-john:
for devs to create a new s1 process easily.


2e2a:
There are already fixtures for s1 processes. See `bin/ad_fixtures`.

PR #2750, opened 2016-09-26
silence linting error

xi:
I do not see the point of adding a docstring here because the module already has one and this is the main function of this module.


2e2a:
ad_checkcode passes on travis.

PR #2751, opened 2016-09-26
Fixup 8f2cea20: fix jsunit tests

xi:
Fixup to 8f2cea20


der-john:
thanks! looks good & works.

PR #2753, opened 2016-09-27
Extend cross window mesg with set and delete token

slomo:
Meant to replace the session extradiction to trusted sites as it has been used for policycompass. Introduces two new cross window messages:

- set auth token and user path (via `setToken`)
- delete auth token (via `deleteToken`)

The embedding site has somehow to get hold of an A3 auth token (eg. creating it using the login api endpoint). It can then set the token to an embedded iFrame.


slomo:
This doesn't yet remove CrossWindowStuff used by pcompass before, because I need both for the transition phase.

slomo:
It would also be interesting to know if either the new or the old api is used / planned to be used by anyother A3 project.

xi:
> It would also be interesting to know if either the new or the old api is used / planned to be used by anyother A3 project.

No, as far as I know no other project uses this API.

xi:
LGTM.

PR #2755, opened 2016-09-27
Add roadmap to documentation

xi:
We have been working on a project roadmap internally and now we want to
publish it.

We also discussed publishing it as github milestones or on a github wiki
page but finally decided that it best fits in our documentation.

For anyone who has seen the previous (internal) draft: I stripped out
anything that I thought should not be publicised and also anything that
is not directly connected to the software but rather to individual
deployments or organizational aspects.

### DIFF
--- a/docs/roadmap.rst
+++ b/docs/roadmap.rst
@@ -5,17 +5,6 @@ Roadmap
 End of September 2016
 ---------------------
 
-Goal
-++++
-
-Polish the meinberlin project.
-
-KPIs
-++++
-
--   project managers can create all currently available process types on
-    meinberlin
-
 Features
 ++++++++
 
@@ -45,14 +34,11 @@ KPIs
 Features
 ++++++++
 
--   process overview can be filtered by region/city
 -   idea collection process with or without a map
 -   private and public processes
--   project managers can invite new users
+-   new users can be invited by liqd
 -   users can edit their own account information
--   project managers can generate embed snippets for a process
--   moderators of processes can download a data file (e.g. CSV) with
-    process contents
+-   embed snippets can be generated for a process
 -   concept for newsfeed / event-stream
 -   concept for notifications (users can follow processes and proposals
     in processes)
@@ -65,15 +51,13 @@ Goal
 ++++
 
 Adhocracy should be optimized to allow processes to be administered by
-initiators to take the workload off of our project managers and
-developers. It also should optimize platform features that allow users
+initiators. It also should optimize platform features that allow users
 and organizations to feel more at home.
 
 KPIs
 ++++
 
 -   we can offer an initial set up of a new process and organization
-    with little effort
 -   At least two persons feel comfortable configuring the adhocracy
     backend (#busfactor)
 -   we have a written strategy for onboarding of new developers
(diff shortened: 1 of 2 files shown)


docs/roadmap.rst
```diff
@@ -0,0 +1,148 @@
...
+++++
+
+-   project managers can create all currently available process types on
+    meinberlin
+
+Features
+++++++++
+
+-   process navigation
+-   SDI for managing processes and users
```
der-john: no feedback / user survey?

2e2a:
I would expect the roadmap points to be in chronological order.

2e2a:
Compiles on readthedocs.

docs/roadmap.rst
xi: not directly related to the software itself

xi:
> I would expect the roadmap points to be in chronological order.

They are in reverse chronological order so that the newest (=most recent) one is at the top.

der-john:
After thinking about the order again, I agree with @2e2a. The roadmap docs are to be updated continuously, I assume, but the plan for the roadmap already goes >6 months into the future. Leaving things as they are, it's less easy for a reader to see what has happened since he last viewed. I think the order that @xi suggested is more useful for changelogs etc than for a roadmap.

der-john:
thanks!

docs/roadmap.rst
```diff
@@ -45,14 +34,11 @@ KPIs
 Features
 ++++++++
 
--   process overview can be filtered by region/city
 -   idea collection process with or without a map
 -   private and public processes
--   project managers can invite new users
+-   new users can be invited by liqd
```
xi: I would avoid mentioning liqd here because it is the main contributor of adhocracy, but not the only one (hopefully)

docs/roadmap.rst
```diff
@@ -45,14 +34,11 @@ KPIs
...
 -   users can edit their own account information
--   project managers can generate embed snippets for a process
--   moderators of processes can download a data file (e.g. CSV) with
-    process contents
+-   embed snippets can be generated for a process
```
xi: This is already possible. The point is to simplify it to a point where project managers can do it independently.

docs/roadmap.rst
```diff
@@ -65,15 +51,13 @@ Goal
...
-initiators to take the workload off of our project managers and
-developers. It also should optimize platform features that allow users
+initiators. It also should optimize platform features that allow users
 and organizations to feel more at home.
 
 KPIs
 ++++
 
 -   we can offer an initial set up of a new process and organization
-    with little effort
```
xi: If you leave this out it is not clear how this is different from the other milestones.

docs/roadmap.rst
```diff
@@ -136,7 +119,7 @@ Features
 -   better discovery of processes (recommended processes, featured
     processes)
 -   search function for comments and contents
--   project managers can edit permissions in all processes
+-   permissions can be added in all processes
```
xi: This is already possible. The point is to simplify it to a point where project managers can do it independently.

der-john: I think two things are misleading about the new wording:
- for LIQD developers, it has been possible to edit permissions all the time. The sentence suggests the contrary.
- permissions can't just be added, they can also be removed, modified etc.

xi:
I am not completely sold on the idea that "initiators" and "project managers" are the same. In my understanding "initiators" are people how initiate processes with the adhocracy UI. Project managers, on the other hand, have a deep knowledge of adhocracy, have access to the SDI and are in contact with developers. Note that this does not require that they are employees of liqd.

Anyway, we can fine-tune this later. It seems like no onw has serious issue with this version, so I am going ahead merging this.

PR #2756, opened 2016-09-27
navigation A: breadcrumbs pointing home

der-john:
(extends #2741.)


der-john:
our team decided that #2757 should be implemented instead of this.

PR #2757, opened 2016-09-27
link called "home"

der-john:
places a link in the `demo` header that says + points to "home".


der-john:
a conflict had been introduced by #2763.

xi:
Note that 	801b42a is a cherry-pick of 	3dc4fe5.

PR #2758, opened 2016-09-28
Send password change notifications

2e2a:
* Send a mail notification when the user password has changed, e.g. via the user settings page
* Ignore changes via the password reset endpoint.


xi:
Are there any more cases that we should ignore? (e.g. register, change by script, ...)

2e2a:
> Are there any more cases that we should ignore? (e.g. register, change by script, ...)

Good point. It is also ignored in this cases.

der-john:
looks great & works.

PR #2528, opened 2016-07-20
use idea collection of core in pcompass

der-john:
continues #2500 and #2489. rebase of #2527.

Unfortunately, the image upload is broken at the moment:

> Error: [$injector:unpr] Unknown provider: flowFactoryProvider <- flowFactory <- adhUploadImageDirective

I'll try fixing it, but I'd be happy about help too.

PR #2759, opened 2016-09-28
fix pcompass

der-john:
this quickly fixes two regressions:

- counter in listing was displayed without "proposals" (regression introduced by #2717)
- "add proposal button" was moved to the header (which is used in #2528), which is not displayed in `pcompass`, because there should be no login.

to display a correct header, the embed option `hideHeader` must be set to false again.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Module.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Module.ts
@@ -11,6 +11,8 @@ import * as AdhTopLevelStateModule from "../../TopLevelState/Module";
 
 import * as Proposal from "./Proposal";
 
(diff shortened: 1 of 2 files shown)


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Module.ts
```diff
@@ -28,6 +30,10 @@ export var register = (angular) => {
             AdhResourceAreaModule.moduleName,
             AdhTopLevelStateModule.moduleName
         ])
+        .config(["adhResourceAreaProvider", (adhResourceAreaProvider) => {
+                adhResourceAreaProvider.names[RIProposal.content_type] = "TR__PROPOSALS";
+                adhResourceAreaProvider.names[RIProposalVersion.content_type] = "TR__PROPOSALS";
+            }])
```
xi: indentation is off

PR #2653, opened 2016-08-30
iconlos durch den tag

der-john:
replaces #2652.

PR #2760, opened 2016-09-28
remap icons

### LINKED ISSUE #2646: IE11 doesn't always render our icon font
Sadly we have no Idea what exact set of preconditions trigger this bug. We cannot reproduce it on our boxes, but we saw it on all machines  at Stendal.

Some screenshots that show the problem are appended.

My analysis: something prevents the custom icon font from loading / showing in some configurations of windows.

Some common reasons would be that external font loading is disabled completely, or in the browser. However, we know that the previous deployment of S1 had working icons. So either they changed that setting recently, or we changed something about how we deploy that font file.
image1
50929cd6-6ba1-11e6-8909-a0455f96aa15
image3
image4
image5
image6

der-john:
attempts to fix #2646.

this PR is mainly supposed to reverse the effects of #2653 on browsers that may download custom fonts while using the utf8 symbols used in #2653 as fallbacks. it aims to do so by using said utf8 symbols as codepoints that point to our icons. (currently one symbol / icon.) if that works, this new mapping should be used in all projects, not just `s1`.

It doesn't work yet. I don't know why and couldn't find docs in fontcustom on how to change codepoints.

additionally to the last commit, I changed `src/adhocracy_frontend/adhocracy_frontend/static/icons/.fontcustom-manifest.json`, line 20, to carry the decimal conversion of the hexadecimal codepoint `\1F3F4`.
```
    "flag": {
      "codepoint": 127988,
      "source": "svg/flag.svg"
    },
```

any help will be much appreciated.


src/adhocracy_frontend/adhocracy_frontend/static/icons/svg/calendar.svg
```diff
@@ -1,31 +0,0 @@
...
-			</g>
-			<g id="g3528">
-				<path id="path3530" fill="#818181" d="M16,10.6h-7.1c-0.5,0-1-0.4-1-1v-8.6c0-0.5,0.4-1,1-1h7.1c0.5,0,1,0.4,1,1v8.7c0,0.5-0.5,0.9-1,0.9zm-6.1-1.9h5.1v-6.8h-5.1v6.8z"/>
-			</g>
-			<g id="g3532">
-				<path id="path3534" fill="#818181" d="m40.9,10.6h-7.1c-0.5,0-1-0.4-1-1v-8.6c0-0.5,0.4-1,1-1h7.1c0.5,0,1,0.4,1,1v8.7c0,0.5-0.4,0.9-1,0.9zm-6.1-1.9h5.2v-6.8h-5.2v6.8z"/>
-			</g>
-		</g>
-	</g>
-</svg>
```
xi: This one is used in mercator

PR #2766, opened 2016-09-29
Additional info columns in sdi

### LINKED ISSUE #2639: SDI: allow filtering users by email and name
The users list in the SDI is very useful - however it becomes considerably less useful as soon as there are many users, because finding the right one without clicking through all users is quite painful.

Therefore, I propose that the filter function is extended to allow filtering by email and name to quickly find the right user to change something.

Chatting with @joka it seems that just showing these additional columns in the SDI will probably do the trick.

2e2a:
* Adds mail and user-name to /principal/users/ view
* Adds Type and Title to organisation/process view, e.g. root

Fixes #2639


der-john:
works.

xi:
The columns work just fine.

However, #2639 suggested that showing the columns will also allow to use them with the filter. That does not work though. It would be really helpful, especially for autonamed resources such as users.

PR #2773, opened 2016-10-05
remap icons

### LINKED ISSUE #2646: IE11 doesn't always render our icon font

xi:
Fixes #2646; Replaces #2760 

This uses `.fontcustom-manifest.json` to define custom unicode codepoints for icons. This way the system fonts can display a reasonable fallback if the iconfont cannot be loaded.

Note that #2760 also removed some unnecessary icons. I did not include that in my pull request because I did not have the time. This should still be done in a separate pull request.


der-john:
I tried to test this with uBlock and got
> GET http://localhost:6551/static/js/Packages/Core/Tracking/Tracking.js net::ERR_BLOCKED_BY_CLIENT

der-john:
I'm having problems testing this.

der-john:
To do:
- [ ] rebuild demo font
- [ ] rebuild mercator font
- [ ] rebuild euth font
- [ ] rebuild pcompass font

xi:
IMHO we do not need to rebuild the other fonts now. This feature is nice, but only really required for s1.

der-john:
works.

PR #2775, opened 2016-10-06
Use german renominate info in s1

2e2a:
The renominate info text appended to the description of a renominated proposal is now in german.


der-john:
nice.

PR #2776, opened 2016-10-06
S1 notification text

2e2a:
Add german texts for the S1 notifications.

To test set the INotification sheet of a user to follow an s1 process and add proposals and comments as another user.


der-john:
nice.

PR #2779, opened 2016-10-06
Fix SDI view permissions

### LINKED ISSUE #2547: Processes in draft state are not listed in SDI

2e2a:
Fixes #2547 

To test create a new processes, which is initially in draft state. It should be listed in the content view.


der-john:
nice.

PR #2781, opened 2016-10-06
Fixup for #2780 (S1 notification text)

2e2a:
Add user name to the activity mail mapping as it is used in the title.


der-john:
good.

PR #2782, opened 2016-10-09
use yaml config file and normalize config options

### LINKED ISSUE #2168: No distinction between frontend_url and canonical_url in backend
The backend uses the config `adhocracy.frontend_url` for two very different things:
-  Point to a place where adhocracy is **embedded**, e.g. for activation links. This is equivalent to what is called `adhocracy.canonical_url` in the frontend config.
-  Point to the frontend **itself**. This is used in `embed_code.html.mako`.

The second usecase was added only recently in #2083. So all deployed configuration assumes the first usecase. I propose to use `adhocracy.canonical_url` for the first usecase to be consistent with the frontend config and use `adhocracy.frontend_url` for the second.

joka:
Frontend / Backend internal changes:
- yaml config instead of ini file, use dictionaries to separate name spaces instead of `.`
- one place for default options (`defaults.yaml`), can be extendend in extension packages (see example in `mercator` package)
- custom config file (absolute path or package file) can be set with the ini file option `yaml.location`
- normalised config option in frontend/backend :
  
  adhocracy.anonymize.enabled -> adhocracy.frontend.anonymize_enabled bool
  adhocracy.custom ->  removed
  adhocracy.custom.\* ->  same,  but json value types now supported,  **Should not be  used** anymore, use adhocracy.frontend.\* instead
  cachebust.enabled -> configurator.cachebust.enabled
  adhocracy_core.caching.http.mode -> adhocracy.caching_mode str
  adhocracy.varnish_url -> adhocracy.caching_proxy str
  adhocracy.thentos_captcha.enabled -> adhocracy.captcha_enabled bool
  adhocracy.thentos_captcha.frontend_url -> adhocracy.captcha_frontend_url str
  adhocracy.thentos_captcha.backend_url -> adhocracy.captcha_backend_url str
  adhocracy.add_default_group -> removed, not used 
  adhocracy.frontend_url -> **changed semantic**: shoud point to the base frontend url
  adhocracy.canonical_url -> same, now also used in the backend with the same semantic, fixes #2168


joka:
We could also remove the namespace "adhocracy". 
We don't use any other namespace and this would improve readability in the backend code.

joka:
Can someone review this?

joka:
LGTM from @dwt

PR #2783, opened 2016-10-10
fix stadtforum

der-john:


xi:
I added a translation for `TR__RESOURCE_POLL`.

PR #2143, opened 2016-03-08
change meinberlin contexts +2

Johnschmidfront:
this is made possible by #2113 i.e. the `CustomHeader` infrastructure.

(This is a minimal version of #2125.)

PR #2784, opened 2016-10-10
fix debate workbench

der-john:
- there is now no empty resource dropdown in the process header (fixup of #2741) 
- the document listing counter now says "1 resource" (unfortunately, the linter complains about `!=`)
- a button saying "add document" appears in the header (i.e. in the `processHeaderSlot`; sort-of continues #2143 and #2113).

I tested `meinberlin`, `demo` and `euth`.

### DIFF
--- a/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Names/Names.ts
+++ b/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Names/Names.ts
@@ -18,6 +18,6 @@ export class Service {
         } else {
             ret = "TR__RESOURCE";
         }
-        return amount != 1 ? ret + "_PLURAL" : ret;
+        return amount !== 1 ? ret + "_PLURAL" : ret;
     }
 }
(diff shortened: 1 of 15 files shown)


der-john:
I just noticed I overlooked `spd`.

der-john:
Done. I haven't tested `spd`, though.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Names/Names.ts
```diff
@@ -18,6 +18,6 @@ export class Service {
         } else {
             ret = "TR__RESOURCE";
         }
-        return amount !== 1 ? ret + "_PLURAL" : ret;
+        return amount != 1 ? ret + "_PLURAL" : ret;
```
xi: While this fixes the symptom, it also hides the underlying issue. My interpretation is the following:

`amount`was not `1`, but `"1"`. This is because of chain of unexpected behavior:

-  The backend returns all values as strings.
-  The frontend converts the values automatically based on information from the meta API
-  The meta API does not contains "search" fields, e.g. `SIPool.count`
-  `SIPool.count` is used as `amount` in the listing

So the proper fix would be to convert the value to number in Listing.ts

der-john: I'm on it.

xi:
LGTM

der-john:
I noticed just now: 2567880 assumes `totalCount` returns strings instead of numbers and we need to fix that. But that can be the case for `commentCount` and other variables as well.

PR #2785, opened 2016-10-10
Fix request to parent resource on subpath rest_url

xi:
Steps to reproduce the issue:

-   have rest API running on a subpath, e.g. /api/
-   visit the root resource (`/r/`) in the browser

This will request `/` and subsequently throw an error because the response can not be parsed as JSON.

This fix works because the root resource is not versionable.


der-john:
lgtm & works. The number of requests goes down.

PR #2787, opened 2016-10-10
Allow to add proposals in sdi

2e2a:
Needed especially to create stadtforum polls.

To test: Create a stadtforum poll in SDI.


xi:
It's a bit strange that I can not set a title on creation, but I guess that is just the way we modelled it.

2e2a:
> It's a bit strange that I can not set a title on creation, but I guess that is just the way we modelled it.

The title is in the versionable. You have to create a proposal an then edit the version0 element.

PR #2789, opened 2016-10-11
Add simple static page resources

2e2a:
When a resource has an IPageService  a "pages/" element is added to the resource pool that can contain IPage elements.

The service is added to the root resource.

To test:
* Create page elements with SDI
* Fetch the pages via the API


xi:
Nice!

joka:
What is the difference to logbook ?

PR #2793, opened 2016-10-11
Add embed sheet and adapter to stadtforum polls

2e2a:
Stadtforum polls now have the default embed code in the IEmbed sheet.


der-john:
lgtm & works.

PR #2794, opened 2016-10-11
Fix admin multi-polygons creation permissions

2e2a:
Adds a missing permission used to add multi-polygons to locations.

When logged in as an admin user in sdi (not god), it should be possible now to add multi-polygons under `/locations.


der-john:
lgtm & works.

PR #2795, opened 2016-10-11
fix noheader option in embed widgets

### LINKED ISSUE #2772: Widget parameter noheader cannot be set to false
When having `data-noheader="false"` in the embed-code the header is not shown.

Only removing the `data-noheader` parameter enables the header.

der-john:
fixes #2772.


xi:
IMHO it is not clear if this is the direction we want to go, so I am putting this on hold. See https://github.com/liqd/adhocracy3/issues/2772#issuecomment-252971662

PR #2798, opened 2016-10-12
Remove some embed widget parameter from the IEmbed sheet if false

2e2a:
See #2772 

The `noheader` and `nocenter` cannot be disabled when setting to false.

As a temporary fix these parameters are deleted from the embed sheet template when set to false.


der-john:
looks good & works.

PR #2799, opened 2016-10-12
Simple static pages frontend

xi:
This adds the frontend code needed to display the simple static pages introduced in #2789. Note that there is no navigation yet.


der-john:
I'm not yet completely convinced that we need resource headers on these pages, but nevertheless: lgtm & works.

PR #2800, opened 2016-10-13
Allow to set empty description and short description

### LINKED ISSUE #2521: Cannot delete IDescription in SDI
Removing an existing description in the SDI and hitting the save button resets the text to the previous state instead of deleting the text.

This is probably also the case in other SDI text / markdown elements, but I haven't tested it.

2e2a:
Fixes #2521 

Allows empty values for `description` and `short_description` of the `IDescription` sheet.

**Important:**
Colander does not differentiate between setting a sheet field to an empty string and not providing the sheet field at all. E.g. Just setting `{description:"Text"}` will set `short_description` to `''`. This means, when changing the description via the API both fields must be provided.

Alternatively, we could also think of allowing the empty value for the text fields in all sheets, but this might break something, if the frontend does not provide all sheet fields.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/sheets/description.py
+++ b/src/adhocracy_core/adhocracy_core/sheets/description.py
@@ -18,8 +18,8 @@ class DescriptionSchema(MappingSchema):
     `description`: a full description
     """
 
-    short_description = Text()
-    description = Text()
+    short_description = Text(missing='')
+    description = Text(missing='')
 
 
 description_meta = sheet_meta._replace(isheet=IDescription,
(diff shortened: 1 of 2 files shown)


xi:
> Colander does not differentiate between setting a sheet field to an empty string and not providing the sheet field at all. E.g. Just setting {description:"Text"} will set short_description to ''. This means, when changing the description via the API both fields must be provided.

Just so I understand correctly: Say I send a PUT request. If I skip a field it will be deleted, but if I skip a complete sheet it will be ignored, right?

2e2a:
>> Colander does not differentiate between setting a sheet field to an empty string and not providing the sheet field at all. E.g. Just setting {description:"Text"} will set short_description to ''. This means, when changing the description via the API both fields must be provided.

> Just so I understand correctly: Say I send a PUT request. If I skip a field it will be deleted, but if I skip a complete sheet it will be ignored, right?

Yes, if we allow empty values for a certain type or sheet field, this will be the case.

That's why we should decide, whether we want to change the sheet behavior in general by allowing empty values in all text fields, or just explicitly allow empty values for certain sheet fields.

This pull request suggests the second to avoid changing to much.

xi:
I talked to @2e2a: Our plan is to switch to this behavior in general, i.e. an empty/missing field will always delete the content instead of being ignored. We will have to flesh out some details, e.g. how this affects default values. A big task is also to ensure that the frontend does not rely on the legacy behavior.

PR #2801, opened 2016-10-17
Make badges and badge-groups addable in sdi

2e2a:
In a `badges/` service (e.g. of an idea collection) it is now possible to add badge-groups and badges via SDI.


der-john:
definitely works.

der-john:
lgtm.

PR #2804, opened 2016-10-18
Better filtering of users in SDI

2e2a:
Allows to filter users by user-name or email.

Substanced uses the `text`index of the system catalog for the filtering functionality, so we extend this index by adding the user-name and email.

To test use the filter field in the SDI users overview.


xi:
Works nice!

PR #2807, opened 2016-10-19
change password security

xi:
We decided that users should be required to enter their password in order to change it. This improves security in cases where the attacker gains access to a user's session, but not their password.

- [X] frontend
- [x] backend

### DIFF
--- a/src/adhocracy_core/adhocracy_core/authentication/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/authentication/__init__.py
@@ -93,24 +93,26 @@ def validate_user_headers(view: callable):
     return wrapped_view
 
 
-def is_header_password(request: IRequest) -> bool:
+def has_password_header(request: IRequest) -> bool:
     """Check if request provided the password in the Password header."""
     return UserPasswordHeader in request.headers
 
 
 def get_header_password(request: IRequest) -> str:
     """Return the password in the Password header."""
-    if is_header_password(request):
-        return request.headers[UserPasswordHeader]
+    return request.headers.get(UserPasswordHeader, None)
 
 
 def validate_password_header(view: callable):
     """Decorator vor :term:`view` to check if the password header may be set.
 
-    :raise pyramid.httpexceptions.HTTPBadRequest: if password is invalid.
+    :raise pyramid.httpexceptions.HTTPBadRequest: if password is invalid or not
+    required. The case that a password is required by a sheet but not set
+    cannot be handled here, as we do not know which sheets are edited by the
+    request.
     """
     def wrapped_view(context, request):
-        password_is_set = is_header_password(request)
+        password_is_set = has_password_header(request)
         error = None
         if password_is_set:
             content = request.registry.content
@@ -121,8 +123,9 @@ def validate_password_header(view: callable):
                 error = error_entry('header',
                                     UserPasswordHeader,
                                     'Invalid password')
-            is_required = content.is_password_required(context, request)
-            if not is_required:
+            is_required_by_some_sheets = \
+                content.is_password_required(context, request)
+            if not is_required_by_some_sheets:
                 error = error_entry('header',
                                     UserPasswordHeader,
                                     'Password not required')
--- a/src/adhocracy_core/adhocracy_core/authentication/__init__.py
+++ b/src/adhocracy_core/adhocracy_core/authentication/__init__.py
@@ -14,6 +14,9 @@ from adhocracy_core.interfaces import error_entry
 UserTokenHeader = 'X-User-Token'
 """The request header parameter to set the authentication token."""
 
+UserPasswordHeader = 'X-User-Password'
+"""The request header parameter to set the user password."""
+
 UserPathHeader = 'X-User-Path'
 """Deprecated: The optional request header to set the userid."""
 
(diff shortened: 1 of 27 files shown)


src/adhocracy_core/adhocracy_core/resources/test_principal.py
```diff
@@ -200,6 +200,9 @@ def test_create(self, meta, registry, principals):
         assert user.tzname == 'UTC'
         assert user.roles == []
         assert user.timezone == timezone(user.tzname)
+        assert user.is_password_valid(registry, '123456') == False
+        assert user.is_password_valid(registry, 'fodThyd2') == True
+
```
xi: To much whitespace

src/adhocracy_core/adhocracy_core/interfaces.py
```diff
@@ -249,6 +249,13 @@ def delete_field_values(fields: [str]):
         """Delete values for every field name in `fields`."""
 
 
+class ISheetRequirePassword(ISheet):
+    """Sheet Interface indicating that a password if required for editing.
```
xi: `s/if/is/`

src/adhocracy_core/adhocracy_core/content/__init__.py
```diff
@@ -326,6 +327,14 @@ def can_delete_anonymized(self, context: object, request: Request) -> bool:
         can_anonymize = _is_anonymized_and_has_permission(context, request)
         return can_anonymize
 
+    def is_password_required(self, context: object, request: Request) -> bool:
+        """Check if some sheets of `context` require a password for editing."""
+        sheets_edit = self.get_sheets_edit(context, request)
+        sheets_require_password = \
+            [s for s in sheets_edit
+             if s.meta.isheet.isOrExtends(ISheetRequirePassword)]
+        return bool(sheets_require_password)
```
xi: `any((s.meta.isheet.isOrExtends(ISheetRequirePassword) for s in sheets_edit))` would be more efficient.

src/adhocracy_core/adhocracy_core/authentication/__init__.py
```diff
@@ -90,6 +93,46 @@ def wrapped_view(context, request):
     return wrapped_view
 
 
+def is_header_password(request: IRequest) -> bool:
```
xi: I am not sure about the conventions, but I would prefer `has_header_password()` or even `has_password_header()`

src/adhocracy_core/adhocracy_core/authentication/__init__.py
```diff
@@ -90,6 +93,46 @@ def wrapped_view(context, request):
...
+    """Check if request provided the password in the Password header."""
+    return UserPasswordHeader in request.headers
+
+
+def get_header_password(request: IRequest) -> str:
```
xi: I am not sure about the conventions, but I would prefer `get_header_password(request, default)` to mirror the behavior of `Dict.get()`

src/adhocracy_core/adhocracy_core/authentication/__init__.py
```diff
@@ -90,6 +93,46 @@ def wrapped_view(context, request):
...
+            is_required = content.is_password_required(context, request)
+            if not is_required:
+                error = error_entry('header',
+                                    UserPasswordHeader,
+                                    'Password not required')
+        if error:
+            request.errors.append(error)
+            raise HTTPBadRequest()
+        return view(context, request)
+    return wrapped_view
```
xi: Correct me if I am wrong, but as far as I can see this is missing the most important error case: `is_required and not password_is_set`

src/adhocracy_core/adhocracy_core/authorization/test_init.py
```diff
@@ -474,4 +474,37 @@ def test_root_acm_extensions_adapter_register(registry, context):
     from . import IRootACMExtension
     root_acm_extension = registry.getAdapter(context, IRootACMExtension)
     assert root_acm_extension == {'principals': [],
+
```
xi: why this whitespace?

src/adhocracy_core/adhocracy_core/interfaces.py
```diff
@@ -249,6 +249,13 @@ def delete_field_values(fields: [str]):
...
+
+    To edit such a Sheet the password needs to be send in the request.
```
xi: This could be more specific, e.g. "… the password needs to be sind in the request header.". I also found this cound this confusing in some other docstrings.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/User/Edit.html
```diff
@@ -33,9 +33,6 @@
             data-ng-model="data.password"
             data-ng-minlength="6"
             data-ng-maxlength="100" />
-        <span class="input-error" data-ng-show="showError(userEditForm, userEditForm.password, 'required')">
-            {{ "TR__ERROR_REQUIRED_PASSWORD" | translate }}
-        </span>
```
der-john: good idea.

der-john:
the frontend code looks good. the whole code works.

the design of the user settings is the only thing I'm not very convinced of anymore. let's not forget to think about that, maybe together with our designer.

xi:
We should clearly document the case when modifications are silently dropped.

src/adhocracy_core/adhocracy_core/authentication/__init__.py
2e2a: `is_required` was not a good name for the variable/function.

At this point we can only tell that some sheets require a password. We do not know if this request wants to edit these sheets and cannot catch that case here.

At a later point during the processing of the sheets, the sheet permissions are checked and also that the password header is required but not set.

src/adhocracy_core/adhocracy_core/authentication/__init__.py
```diff
@@ -93,24 +93,26 @@ def wrapped_view(context, request):
...
+def has_password_header(request: IRequest) -> bool:
     """Check if request provided the password in the Password header."""
     return UserPasswordHeader in request.headers
 
 
 def get_header_password(request: IRequest) -> str:
     """Return the password in the Password header."""
-    if is_header_password(request):
-        return request.headers[UserPasswordHeader]
+    return request.headers.get(UserPasswordHeader, None)
```
xi: the `None` default is implicit

src/adhocracy_core/adhocracy_core/authorization/test_init.py
```diff
@@ -474,10 +474,8 @@ def test_root_acm_extensions_adapter_register(registry, context):
     from . import IRootACMExtension
     root_acm_extension = registry.getAdapter(context, IRootACMExtension)
     assert root_acm_extension == {'principals': [],
-
                                   'permissions': []}
 
-
```
xi: Now you removed too much whitespace

src/adhocracy_core/docs/rest_api.rst
```diff
@@ -440,6 +440,11 @@ attribute in the json tree.)
 
 PUT responses have the same fields as POST responses.
 
+Note, when putting multiple sheets in a request some changes might be currently
+dropped when the request does not have sufficient permissions, e.g. cannot be
+edit by the user or requires an additional header.
```
xi: Should use the [note directive](http://www.sphinx-doc.org/en/1.4.8/markup/para.html?highlight=note#directive-note)

PR #2808, opened 2016-10-19
add picture to home

der-john:
continues #2722 by adding a bg-image to the resource header on the home page.


2e2a:
Short description is not shown in the picture, as it is done in process headers. Instead the long description is shown below the picture.

Is this intended?

xi:
> Short description is not shown in the picture, as it is done in process headers. Instead the long description is shown below the picture.

There were no designs for this. This version is not really intended for production use. But we plan on working on a proper frontpage soon.

xi:
LGTM.

PR #2746, opened 2016-09-23
forum of forums

der-john:
This roughly adds the logic needed to provide `stadtforum` poll listings outside embedded contexts.

*Warning:* The layout is still suboptimal and the logic may be insufficient.

To do:
- [ ] improve layout
- [ ] add i18n for polls (after #2732 is merged)
- [ ] make create and edit work

PR #2810, opened 2016-10-20
integrate Stadtforum into IdeaCollection

der-john:
This continues #2746 by integrating Stadtforum logic into IdeaCollection. It's become pretty long. :(

this is achieved like #2489: meinberlin-specific code stays in `Stadtforum/Module.ts`, writing values into `AdhProcessProvider.processProperties`. These, in turn, are used in `IdeaCollection/Proposal` etc. It adds the following attributes to `processProperties`:
- `hasAuthorInListItem`: a boolean to determine whether proposal's `creator` and `creationDate` are displayed in the `ListItem` template,
- `hasCommentColumn`: a boolean to determine whether comments appear in their own column,
- `hasDescription`: a boolean to determine whether proposals have a description,
- `proposalColumn`: this contains a template url in which a process-specific `detailColumn` - analogous to `processHeaderSlot`.

Along the way, Stadtforum processes are registered in the `mein.berlin.de` context.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Poll/Proposal.ts
```diff
@@ -0,0 +1,59 @@
+/// <reference path="../../../../../lib2/types/angular.d.ts"/>
```
xi: This file should be called `Poll.ts`

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/ListItem.html
```diff
@@ -7,10 +7,11 @@
     ></adh-show-image>
     <div class="list-item-body">
         <div class="list-item-author">
-            {{ "TR__BY" | translate }}
-            <adh-user-meta data-ng-if="data.creator" data-path="{{data.creator}}" data-no-link="true"></adh-user-meta>
-            {{ "TR__ON" | translate }}
-            <adh-time data-datetime="data.creationDate" data-format="L"></adh-time>
+            <div data-ng-if="processProperties.hasAuthorInListItem">{{ "TR__BY" | translate }}
```
xi: the `ngIf` should be on the `list-item-author` element.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Proposal.ts
```diff
@@ -302,6 +306,9 @@ export var listItemDirective = (
                     scope.selectedState = "is-not-selected";
                 }
             }));
+            if (typeof scope.processProperties.hasAuthorInListItem === "undefined") {
+                scope.processProperties.hasAuthorInListItem = true;
```
xi: I am not convinced that setting defaults here is a good idea. It should be easy to find out what the default values are. I don't know of any better place though

xi:
I am not completely convinced of the general premise of this pull request is a good idea.

Sure, the poll workbench currently shares a lot of code with the idea collection workbench. But so does the debate workbench (and every other workbench). So the underlying question here is: How can we reduce code duplication between workbenches?

I am pretty sure that merging all workbenches into one is not the way to go, because that would require `processProperties` to become messy. Instead, I see two strategies:

- Split the existing cases into `the right` set of workbenches.
- Reduce the boilerplate code required to create a workbench

So the question in this pull request is something like this: Should poll stay its own workbench or should it be integrated in either idea collection or debate.

- **Idea collection** is a workbench where users submit ideas (proposals) that can be rated and commented. The different processes that use it differ mostly in the features of the proposal (e.g. map)
- **debate** is a workbench where initiators create texts that can be commented by the community
- **poll** is a workbench where initiators create questions that can be commented and answered (opionion; a lot like rating) by the community

So poll is a lot more like debate, but still significantly different. Most importantly, I think that both poll and debate might change significantly in the future.

der-john:
> So poll is a lot more like debate, but still significantly different.

I don't agree. In my opinion, the difference that polls deal with **questions** and debates deal with **paragraphs** is decisive. From the point of view of the data and the html, the common denominator between Idea Collections and "Poll Collections" is much larger. You're right about the user roles, but I'd like to let backend concerns stay backend concerns.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Proposal.ts
der-john: > I am not convinced that setting defaults here is a good idea. It should be easy to find out what the default values are.

point taken.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/ListItem.html
der-john: you're right.

der-john:
To do:
- [ ] optionally disable comment routes,
- [ ] introduce new parameter `hasDescription` + explicit default,
- ...

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Proposal.ts
```diff
@@ -124,7 +124,8 @@ export var bindPath = (
                     };
 
                     if (!scope.processProperties.proposalColumns) {
-                        scope.data.detail = SIDescription.get(resource).description;
+                        var descriptionSheet : SIDescription.ISheet = SIDescription.get(resource);
```
xi: The type is not required because it can be inferred from `SIDescription.get()`

der-john: ah. thanks.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/ListItem.html
der-john: I still need to do that.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Poll/Proposal.ts
der-john: ok.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Process/Process.ts
```diff
@@ -31,6 +31,7 @@ export interface IProcessProperties {
     hasLocation? : boolean;
     hasLocationText? : boolean;
     hasAuthorInListItem? : boolean;
+    hasDescription? : boolean;
```
der-john: this will be become a mandatory attr in 9be7b1a

src/euth/euth/static/js/Packages/Euth/IdeaCollection/Module.ts
```diff
@@ -32,7 +32,7 @@ export var register = (angular) => {
 
             _.forEach([RIEuthProcess, RIEuthPrivateProcess], (processType) => {
                 var registerRoutes = AdhIdeaCollectionWorkbench.registerRoutesFactory(
-                    processType, RIProposal, RIProposalVersion);
+                    processType, RIProposal, RIProposalVersion, true);
```
der-john: This is not strictly necessary, of course, but I think it's sensible.

src/meinberlin_lib/meinberlin_lib/static/js/Packages/Meinberlin/MeinberlinDe/Module.ts
```diff
@@ -51,13 +54,13 @@ export var register = (angular) => {
...
+                RIBuergerhaushaltProcess, RIBuergerhaushaltProposal, RIBuergerhaushaltProposalVersion, true);
             registerRoutes1("mein.berlin.de")(adhResourceAreaProvider);
             var registerRoutes2 = AdhIdeaCollectionWorkbench.registerRoutesFactory(
-                RIIdeaCollectionProcess, RIGeoProposal, RIGeoProposalVersion);
+                RIIdeaCollectionProcess, RIGeoProposal, RIGeoProposalVersion, true);
             registerRoutes2("mein.berlin.de")(adhResourceAreaProvider);
             var registerRoutes3 = AdhIdeaCollectionWorkbench.registerRoutesFactory(
-                RIKiezkasseProcess, RIKiezkasseProposal, RIKiezkasseProposalVersion);
+                RIKiezkasseProcess, RIKiezkasseProposal, RIKiezkasseProposalVersion, true);
             registerRoutes3("mein.berlin.de")(adhResourceAreaProvider);
```
der-john: Due to rebaseing, `registerRoutes4` (see b82df69) somehow got lost. I fixed it in 2303ecc.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Process/Process.ts
```diff
@@ -32,6 +32,7 @@ export interface IProcessProperties {
     hasLocationText? : boolean;
     hasAuthorInListItem? : boolean;
     hasDescription? : boolean;
+    hasCommentColumn : boolean;
```
der-john: this will be become a mandatory attr in 9be7b1a.

src/euth/euth/static/js/Packages/Euth/IdeaCollection/Module.ts
```diff
@@ -48,6 +48,7 @@ export var register = (angular) => {
                     hasDescription: true,
                     hasImage: true,
                     proposalClass: RIProposal,
+                    proposalDetailColumn: adhConfig.pkg_path + AdhIdeaCollectionWorkbench.pkgLocation + "/ProposalDetailColumn.html",
```
der-john: this will be become a mandatory attr in 9be7b1a.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Proposal.ts
```diff
@@ -123,7 +123,7 @@ export var bindPath = (
                         assignments: assignments
                     };
 
-                    if (!scope.processProperties.proposalColumns) {
+                    if (!scope.processProperties.hasDescription) {
```
der-john: There's a `!` missing here. This gets fixed in 31b7d31.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Proposal.ts
```diff
@@ -184,7 +184,7 @@ var fill = (
     SITitle.set(proposalVersion, {
         title: scope.data.title
     });
-    if (!scope.processProperties.proposalColumns) {
+    if (!scope.processProperties.hasDescription) {
```
der-john: There's a `!` missing here. This gets fixed in 31b7d31.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/Create.html
```diff
@@ -32,7 +32,7 @@
         </label>
 
         <!-- Detail Description -->
-        <label data-ng-if="hasDetailText">
+        <label data-ng-if="processProperties.hasDescription">
```
der-john: The attribute `hasDescription` needs to be added to the `processProperties` interface, of course. (The commit got lost.) This happens in d1b36e4.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Poll/ProposalColumns.html
```diff
@@ -0,0 +1 @@
+<adh-poll-detail-column data-transclusion-id="transclusionId" data-view="view" data-process-properties="processProperties"></adh-poll-detail-column>
```
der-john: One may ask: Why isn't this a detail column already? The answer is: We need to get `proposalUrl` into the scope, and for that we need a separate directive.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Process/Process.ts
```diff
@@ -34,6 +34,7 @@ export interface IProcessProperties {
     maxBudget? : number;
     processClass?;
     proposalClass;
+    proposalColumns?;
```
der-john: this will be become a mandatory attr in 9be7b1a.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Workbench/Workbench.ts
```diff
@@ -145,7 +145,8 @@ export var addProposalButtonDirective = (
 export var registerRoutesFactory = (
     ideaCollection,
     proposalType,
-    proposalVersionType
+    proposalVersionType,
+    hasCommentColumn = true
```
xi: It would be much nicer if this could use the processProperties. But I don't see how this could be done in a nice way.

xi:
LGTM. Feel free to merge.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Process/Process.ts
```diff
@@ -26,17 +26,17 @@ export interface IStateData {
 }
 
 export interface IProcessProperties {
-    hasCommentColumn : boolean;
+    hasCommentColumn? : boolean;
```
der-john: I understand that it's preferable to have all these booleans behave analogously and default to `false`. But why did you call this "improve type"?

PR #2813, opened 2016-10-24
Fix icon font fallbacks

### LINKED ISSUE #2812: demo: icon fallbacks broken
This is a slightly confusing situation:
Icons work all right in demo, but their fallbacks don't.

It's probably related to #2773.

xi:
Fixes #2812 

The issue was that I had some legacy files lying around. Removing them and re-building the icon fonts fixed the issue.


der-john:
looks okay & works.

xi:
One possible place to search for unicode characters is http://unicode-table.com/

PR #2809, opened 2016-10-19
plain resources

xi:
This is basically a raw dump of my continued work on #2206, so it is a bit unstructured.

I wanted to implement the "create resources as plain objects instead of class instances" part. But somehow I got deep into "change mkResources.ts to only generated interfaces and constants", although it is not complete yet. The results:

- `RIFoo` is still a class, but only because it was simpler to leave it that way.
- `SIFoo.Sheet` is now an interface.
- Instead of `resource[SIFoo.nick]` you use `SIFoo.get(resource)` / `SIFoo.set(resource, ...)`. The benefit of this is that the correct sheet type is used, so it is easier to spot errors.
- Size of the generated files is down to 107KB (see https://github.com/liqd/adhocracy3/pull/2806#discussion_r84033504 for previous values). About 50% of that is UMD code that could be stripped.

Still to do:

- Make `ResourcesBase.IResource` generic to allow even better type checking
- Further simplify `mkResources` to improve maintainability and performance

PR #2818, opened 2016-10-25
Mercator: fixup #2809

der-john:
Some types and objects were broken by #2809.

- [x] check euth, pcompass and spd.


src/mercator/mercator/static/stylesheets/scss/components/_mercator_proposal.scss
```diff
@@ -14,12 +14,8 @@ selected. If no proposal is selected, no state is applied.
         border: 0;
     }
 
-    .annotated-section {
-        width: 157.5%;
```
der-john: interesting. why was this there?

der-john:
great job! works.

PR #2821, opened 2016-10-25
stadtforum: show process resource header

der-john:
Shows this resource header whenever a3 is not embedded.

Analogous to #2741.


2e2a:
works

PR #2822, opened 2016-10-25
rm counter from (s1) user activities

### LINKED ISSUE #2817: PAAS: counter is redundant
As visible on the screenshot of the S1 user activity overview below, there are two counters exactly one above the other.

Since the first counter includes proposals, rates and comments, and the second one only proposals, I suggest I remove the second one.

paas-design kaputt

der-john:
Fixes #2817.

### DIFF
--- a/src/s1/s1/static/js/Packages/S1/Proposal/Proposal.ts
+++ b/src/s1/s1/static/js/Packages/S1/Proposal/Proposal.ts
@@ -336,7 +336,8 @@ export var listingDirective = (
             sorts: "=?",
             state: "@?",
             decisionDate: "@?",
-            creator: "@?"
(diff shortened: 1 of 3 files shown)


src/s1/s1/static/js/Packages/S1/Proposal/Proposal.ts
```diff
@@ -355,6 +356,7 @@ export var listingDirective = (
             if (scope.decisionDate) {
                 scope.params.decision_date = scope.decisionDate;
             }
+            scope.counter = typeof scope.counter === "undefined" ? true : scope.counter;
```
xi: this is a bit shaky, but ok

PR #2823, opened 2016-10-25
add image to stadtforum process header

der-john:


der-john:
thanks for reviewing.

PR #2744, opened 2016-09-23
mitmachen verboten

der-john:
Stops displaying the "participate" button in IdeaCollection to non-logged-in users.

The process in US https://tree.taiga.io/project/ggr-meinberlinteam/us/458?milestone=93539 (IdeaColl with fixed proposals) required that users may not create proposals. @2e2a implemented this by taking an `IdeaCollection` and setting its `workflow` to `debate(_private)`.

this is a quickfix to make the "mitmachen" button = `AddProposalButton` disappear: check whether `workflow` is `debate(_private)`.

PR #2825, opened 2016-10-25
mitmachen wieder verboten

der-john:
In #2744, the add-proposal-button was disabled in "debate" workflows by using a quickfix. This PR extends the quickfix in order to disable the button in Stadtforums as well. I'm happy to discuss other ways to do this - probably we should find a cleaner solution for both problems.

Ceterum censeo: We'd have none of the problems if we didn't have the "what's a standard user allowed to do now?" problem ( = #1347).


xi:
LGTM.

PR #2826, opened 2016-10-26
change paas header implementation to be more similar to core

### LINKED ISSUE #2814: PAAS user column on master slightly broken
On the S1 user profile page, the header gets partially covered by adhocracy content. Probably caused by #2805.

paas-design kaputt

xi:
Fixes #2814 

Note that some spacings changed. I did not want to get into the details.

### DIFF
--- a/src/s1/s1/static/stylesheets/scss/components/_s1_header.scss
+++ b/src/s1/s1/static/stylesheets/scss/components/_s1_header.scss
@@ -10,6 +10,11 @@
     }
 }
 
+.s1-header {
+    @include rem(margin-bottom, 10px);
+    position: relative;
+}
+
 .s1-header-menu {
     white-space: nowrap;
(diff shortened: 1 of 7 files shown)


src/s1/s1/static/stylesheets/scss/components/_s1_header.scss
```diff
@@ -13,38 +10,6 @@
     }
 }
 
-.s1-header {
-    margin-top: rem(50px);
-    margin-bottom: rem(10px);
-
-    position: relative;
-}
```
der-john: Why was this removed? It's used in `src/s1/s1/static/js/Packages/S1/Context/header.html`.

der-john:
nice! looks good & works.

xi:
@der-john I think you forgot to revert 	f11c524

PR #2827, opened 2016-10-26
change email address

xi:
This pull request serves as a hub for implementing changing ones email address. This will work a lot like changing the password (#2807): In the user settings, the user can set a new email address which also requires providing the password.

What is different here is that after setting the new email address, it will remain inactive until it has been activated (similar to initial account activation). It is not completely clear yet how much code we can reuse for this activation. Another question is what happens to existing sessions on activation.
- [x] Implement backend to change email address
- [X] Implement UI to change email address
- [x] Implement activation backend
- [x] Implement activation frontend


2e2a:
@xi, @der-john: Just committed the backend part of the email activation.

src/adhocracy_core/adhocracy_core/messaging/__init__.py
```diff
@@ -339,6 +339,27 @@ def send_password_change_mail(self, user=IUser, request: Request=None):
                        request=request,
                        )
 
+    def send_new_email_activation_mail(self, user: IUser, activation_path: str,
+                                       new_email: str, request: Request=None):
+        """Send a avtivation mail to validate the new email of a user."""
```
xi: typo: "activation"

src/adhocracy_core/adhocracy_core/locale/en/LC_MESSAGES/adhocracy.po
```diff
@@ -168,14 +168,52 @@ msgstr ""
...
+"Have fun and be inspired!\n"
+"\n"
+"Your \"${site_name}\" Team\n"
+"\n"
+"----------------------------------------------------------\n"
+"\n"
+"Hallo ${user_name},\n"
+"\n"
+"Diese E-Mail dient der Validierung Deiner neuen Email-Adresse. Bitte\n"
+"nutze den folgenden Link um sie aktivieren.\n"
```
xi: typo: "um sie zu aktivieren"

src/adhocracy_core/adhocracy_core/rest/schemas.py
```diff
@@ -463,16 +465,22 @@ def validate_activation_path(node, value):
...
             autoupdated = False
-            event = ResourceSheetModified(user, IUserBasic, request.registry, {},
+            event = ResourceSheetModified(user, IUserBasic, registry, {},
                                           {}, request, autoupdated)
             registry.notify(event)  # trigger reindex activation_path index
     return validate_activation_path
 
 
+
+
```
xi: Extra whitespace

xi:
Generally this seems to work fine. I think we can leave the activation frontend as it is.

Some questions/ideas:

- It might be a good idea to let the activation link expire after some time.
- I think at one point we discussed sending an information mail to the old email address.

2e2a:
> Generally this seems to work fine. I think we can leave the activation frontend as it is.

We might still consider changing the text shown after the activation to better cover both cases.

> It might be a good idea to let the activation link expire after some time.

This  happens already after a week.

> I think at one point we discussed sending an information mail to the old email address.

As far as I remember we decided not to do this, to avoid sending to much to the user.

2e2a:
Fixed typos

xi:
> > I think at one point we discussed sending an information mail to the old email address.
>
> As far as I remember we decided not to do this, to avoid sending to much to the user.

The way I remember it we decided not to require activation from both accounts. This is why this would only be an information mail to the old account.

PR #2828, opened 2016-10-28
provide new image size 'original'

der-john:
Adds a new image format called 'original' besides 'detail' and 'thumbnail'.

Tests pass as before, but I haven't written a new one for the new logic yet.

### DIFF
--- a/src/adhocracy_core/adhocracy_core/resources/image.py
+++ b/src/adhocracy_core/adhocracy_core/resources/image.py
@@ -47,9 +47,12 @@ class ImageDownload(File, AssetDownload):
         """
         if self._is_resized():
             return self._get_response()
-        elif self.dimensions:
+        elif self.dimensions or self.dimensions is None:
             original = self._get_asset_file_in_lineage(registry)
-            self._upload_crop_and_resize(original)
+            if self.dimensions is None:
+                self._upload(original)
+            else:
+                self._upload_crop_and_resize(original)
             transaction.commit()  # to avoid BlobError: Uncommitted changes
             return self._get_response()
         else:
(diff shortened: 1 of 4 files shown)


joka:
Sidenote: we did not do this in the first place to prevent high server due to big image sizes.

2e2a:
> Sidenote: we did not do this in the first place to prevent high server load due to big image sizes.

The main reason for this pull request ist to keep the original ratio. So we might reduce the the image size.

xi:
> > Sidenote: we did not do this in the first place to prevent high server load due to big image sizes.
>
> The main reason for this pull request ist to keep the original ratio. So we might reduce the the image size.

Absolutely! It's probably best to fix the area, i.e. `width * height == X`.

src/adhocracy_core/adhocracy_core/resources/image.py
```diff
@@ -47,9 +47,12 @@ def get_response(self, registry: Registry=None) -> FileResponse:
         """
         if self._is_resized():
             return self._get_response()
-        elif self.dimensions:
+        elif self.dimensions or self.dimensions is None:
```
2e2a: This is always true. Would need to clearly split the two cases.

src/adhocracy_core/adhocracy_core/resources/image.py
```diff
@@ -65,6 +68,31 @@ def _is_resized(self) -> bool:
...
+                           progressive=True,
+                           quality=80,
+                           optimize=True)
+            else:
+                image.save(bytestream,
+                           image.format)
+            bytestream.seek(0)
+        self.upload(bytestream)
+        self.mimetype = original.mimetype
+
```
2e2a: Avoid code duplication with _upload_crop_and_resize

2e2a: Need to reduce image size. Consider @xi's comment.

src/adhocracy_core/adhocracy_core/resources/image.py
der-john: Oops.

der-john:
doesn't seem to be worth the effort anymore, since demo is not being realized.

PR #2833, opened 2016-11-02
rename + relocate 'proposalColumn'

der-john:
Follow-up of #2810. Enriches #2834.

While working on a bigger refactoring, I noticed that the attribute `proposalColumn` on `IProcessProperties`
- had a name that didn't reflect its semantics: a poll isn't really a proposal;
- hid the fact it was used analogously as `processHeaderSlot`;
- had as a value a template url that was located either in `IdeaCollection/Workbench` or `IdeaCollection/Poll`, but the template in `Workbench` was actually specific to `IdeaCollection/Proposal`.

So I hereby suggest to call the attribute `detailSlot`. This way,
- the function of the template is highlighted: it's a detail view for either proposals, or polls (or documents in the future);
- it becomes clear that it works like `processHeaderSlot`.

Furthermore, I hereby suggest to move the one template from `IdeaCollection/Workbench` to `IdeaCollection/Proposal`.

In #2834/ 1931bae I suggest to rename to more attributes of `IProcessProperties`. I wanted to this here, but there would have been several incompatibilities.


xi:
LGTM

PR #2834, opened 2016-11-02
integrate DebateWorkbench into IdeaCollection

der-john:
Like #2810, but with debate workbenches instead of stadtforums.

One new, optional attribute is added to `IProcessProperties`, a boolean called `document` (04385c9). If it is `true`, the create and edit views are document-focused, i.e. they are the `DocumentCreate` view etc. If not, they are proposal-focused, i.e. `ProposalCreate` etc. See 101fe6e.

- `IdeaCollection`,
- `s1IdeaCollection` (which can be integrated into `IdeaCollection` soon),
- `mercator` (which will probably not be updated).

In my opinion, having just one workbench module can reduce the maintainance effort of the frontend. E.g., there can be one logical place to put `ProcessCreate` and `ProcessEdit` views.

Also, we can simplify some concepts and names. `IdeaCollection/Workbench` can become simply `Workbench`, `IdeaCollection/Proposal` can become `Proposal` - which is analogous to what `Document` is right now -, and `IdeaCollection/Process` can maybe merge into the existing `Process` module (or replace it if we think the existing one should merge into `ResourceArea`).


der-john:
I haven't tested whether the 'add document' button appears as expected yet.

xi:
Could you please rebase?

der-john:
Is being rebased.

PR #2837, opened 2016-11-08
integrate DebateWorkbench into IdeaCollection

der-john:
Like #2810, but with debate workbenches instead of stadtforums. Replaces #2834.

One new, optional attribute is added to `IProcessProperties`, a boolean called `document` (04385c9). If it is true, the create and edit views are document-focused, i.e. they are the `DocumentCreate` view etc. If not, they are proposal-focused, i.e. `ProposalCreate` etc. See 101fe6e.

- `IdeaCollection`,
- `s1IdeaCollection` (which can be integrated into `IdeaCollection` soon),
- `mercator` (which will probably not be updated).
In my opinion, having just one workbench module can reduce the maintainance effort of the frontend. E.g., there can be one logical place to put ProcessCreate and ProcessEdit views.

Also, we can simplify some concepts and names some time soon. `IdeaCollection/Workbench` can become simply `Workbench`, `IdeaCollection/Proposal` can become `Proposal` - which is analogous to what `Document` is right now -, and `IdeaCollection/Process` can maybe merge into the existing `Process` module (or replace it if we think the existing one should merge into `ResourceArea`).


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Workbench/Module.ts
```diff
@@ -27,6 +27,10 @@ export var register = (angular) => {
             AdhResourceAreaModule.moduleName,
             AdhTopLevelStateModule.moduleName
         ])
+        // the (document-focused) IdeaCollection is also registered under the directive name
+        // "adhDebateWorkbench" such as not to break currently running embeds.
+        .directive("adhDebateWorkbench", [
+            "adhTopLevelState", "adhConfig", "adhHttp", Workbench.workbenchDirective])
```
xi: A better alternative may be to register an embed alias (see the definition of `registerDirective()` in Embed.ts)

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Workbench/Workbench.ts
```diff
@@ -226,6 +236,90 @@ export var registerRoutesFactory = (
...
+                    var documentUrl = _.last(_.sortBy(SIParagraph.get(commentable).documents));
+                    return {
+                        commentableUrl: commentable.path,
+                        commentCloseUrl: documentUrl,
+                        documentUrl: documentUrl
+                    };
+                });
+            };
+        }]);
+    }
```
xi: I think it would be clearer to put the proposal/document specific routes into separate functions, maybe even into the respective modules.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Process/Process.ts
```diff
@@ -27,13 +27,14 @@ export interface IStateData {
 
 export interface IProcessProperties {
     detailSlot : string;
+    document? : boolean;
```
xi: I would prefer to continue with the slot concept, so add `editSlot` and `createSlot`.

der-john:
> I think it would be clearer to put the proposal/document specific routes into separate functions, maybe even into the respective modules.

On doing this, it is better to rebase this PR.

PR #2839, opened 2016-11-08
fix stadtforum embed

der-john:
(replaces #2838.) Since #2810 the attribute `processProperties` is required for some directives. Unfortunately, this breaks embedding of these directives.

This branch contains a fix for `meinberlin-stadtforum-proposal-detail`. The following embeds may also be broken:

- `meinberlin-stadtforum-proposal-create`
- `meinberlin-proposal-detail`
- `meinberlin-proposal-list-item`
- `meinberlin-proposal-create`
- `meinberlin-proposal-edit`
- `meinberlin-proposal-list`
It is not clear whether these are used though.


xi:
LGTM.

PR #2840, opened 2016-11-08
rm unused embeddable directives

xi:
Adhocracy generally allows to mark directives as embeddable. This may be done for any of the following reasons:

1. The directive should be embedded on a 3rd party website
2. The directive should be tested on its own through protractor
3. The directive should be tested on its own by developers

In the past we marked many directives as embeddable mainly because of (3): It made development simpler.

When a directive becomes embeddable, the directive, including its interface, become part of the public embedding API which needs to maintain backwards compatibility. This is a strong reason not to mark directives as embeddable by default. (Issues like #2839 may occur)

In an effort to reduce the size of our public API I tried to find out which directives are actually used. Piwik gave me the following results:

- `mein-berlin-bplaene-proposal-embed`
- `meinberlin-stadtforum-proposal-detail`

In addition to that, the following directives are used in acceptance tests:

- `create-or-show-comment-listing`
- `social-share`
- `mercator-2016-proposal-create`

In this pull request I am removing all other embeddable directives.


der-john:
looks good to me.

PR #2841, opened 2016-11-08
der prozess

der-john:
Like #2810, but with debate workbenches instead of stadtforums. Replaces #2834 / #2837.

Three new "slot attributes" are added to `IProcessProperties`: `CreateSlot` (mandatory), `EditSlot` (mandatory) and `ImageSlot` (optional). They work the same way as `DetailSlot` and `ProcessHeaderSlot`. See 101fe6e.

- IdeaCollection,
- s1IdeaCollection (which can be integrated into IdeaCollection soon),
- mercator (which will probably not be updated).

In my opinion, having just one workbench module can reduce the maintainance effort of the frontend. E.g., there can be one logical place to put `ProcessCreate` and `ProcessEdit` views.

Also, we can simplify some concepts and names some time soon. IdeaCollection/Workbench can become simply Workbench, IdeaCollection/Proposal can become Proposal - which is analogous to what Document is right now -, and IdeaCollection/Process can maybe merge into the existing Process module (or replace it if we think the existing one should merge into ResourceArea).


der-john:
guess what: I'm rebasing this.

PR #2842, opened 2016-11-08
der prozess

der-john:
Like #2810, but with debate workbenches instead of stadtforums. Replaces #2834 / #2837 / #2841.

Two new "slot attributes" are added to `IProcessProperties`: `CreateSlot` and `EditSlot`. They work the same way as `DetailSlot` and `ProcessHeaderSlot`. See 101fe6e.

Parallel to these, the `adhIdeaCollectionImageColumn` gets used whenever the existing attribute `processProperties.hasImage` is true.

- IdeaCollection,
- s1IdeaCollection (which can be integrated into IdeaCollection soon),
- mercator (which will probably not be updated).
In my opinion, having just one workbench module can reduce the maintainance effort of the frontend. E.g., there can be one logical place to put ProcessCreate and ProcessEdit views.

Also, we can simplify some concepts and names some time soon. IdeaCollection/Workbench can become simply Workbench, IdeaCollection/Proposal can become Proposal - which is analogous to what Document is right now -, and IdeaCollection/Process can maybe merge into the existing Process module (or replace it if we think the existing one should merge into ResourceArea).


src/spd/spd/static/js/Packages/spd/CollaborativeText/Module.ts
```diff
@@ -0,0 +1,75 @@
...
+            AdhIdeaCollectionModule.moduleName,
+            AdhMovingColumnsModule.moduleName,
+            AdhPermissionsModule.moduleName,
+            AdhResourceActionsModule.moduleName,
+            AdhResourceAreaModule.moduleName,
+            AdhTopLevelStateModule.moduleName
+        ])
+        .config(["adhEmbedProvider", (adhEmbedProvider : AdhEmbed.Provider) => {
+            adhEmbedProvider
+                .registerDirective("debate-workbench");
```
xi: not required. see #2840 (see also https://github.com/liqd/adhocracy3/pull/2842/files#diff-7dfab1596a678b22f54354682fe773faR30)

src/spd/spd/static/js/Packages/spd/CollaborativeText/Module.ts
```diff
@@ -0,0 +1,75 @@
...
+        }])
+        .config(["adhResourceAreaProvider", "adhConfig", (adhResourceAreaProvider: AdhResourceArea.Provider, adhConfig) => {
+            var registerRoutes = (context? : string) => (provider) => {
+                AdhIdeaCollectionWorkbench.registerCommonRoutesFactory(
+                    RIDigitalLebenProcess, RIDocument, RIDocumentVersion)(context)(provider);
+                AdhIdeaCollectionWorkbench.registerDocumentRoutesFactory(
+                    RIDigitalLebenProcess, RIDocument, RIDocumentVersion)(context)(provider);
+            };
+            registerRoutes()(adhResourceAreaProvider);
```
xi: IMHO wrapping this in functions only makes things more complicated in this case.

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Proposal/ListItem.html
```diff
@@ -1,7 +1,7 @@
-<a class="list-item {{selectedState}}" data-ng-class="{ 'm-contains-image': processProperties.hasImage }" href="{{path | adhParentPath | adhResourceUrl }}">
+<a class="list-item {{selectedState}}" data-ng-class="{ 'm-contains-image': processProperties.imageSlot }" href="{{path | adhParentPath | adhResourceUrl }}">
     <adh-show-image
         data-css-class="list-item-image"
-        data-ng-if="processProperties.hasImage"
+        data-ng-if="processProperties.imageSlot"
```
xi: still uses `imageSlot`

src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/IdeaCollection/Workbench/Workbench.ts
```diff
@@ -117,6 +164,26 @@ export var detailColumnDirective = (
...
+        link: (scope) => {
+            scope.$on("$destroy", adhTopLevelState.bind("processUrl", scope));
+            adhPermissions.bindScope(scope, () => scope.processUrl, "processOptions");
+
+            scope.setCameFrom = () => {
+                adhTopLevelState.setCameFrom();
+            };
+        }
+    };
+};
```
xi: This is very similar to `addProposalButtonDirective`. Is it possible to merge them?

xi:
I tested manually and everything seems to work fine.

src/spd/spd/static/js/Packages/spd/CollaborativeText/Module.ts
```diff
@@ -58,8 +48,9 @@ export var register = (angular) => {
                 "<adh-idea-collection-workbench data-process-properties=\"processProperties\">" +
                 "</adh-idea-collection-workbench>";
             adhProcessProvider.setProperties(processType, {
+                createSlot: adhConfig.pkg_path + AdhDocument.pkgLocation + "/CreateSlot.html",
                 detailSlot: adhConfig.pkg_path + AdhDocument.pkgLocation + "/DetailSlot.html",
-                document: true,
+                editSlot: adhConfig.pkg_path + AdhDocument.pkgLocation + "/EditSlot.html",
```
der-john: oops. thank you!

PR #2843, opened 2016-11-09
Service Konto Frontend

xi:
This implements the frontend part of service konto:

- On the login page there is an additional button saying "login with Service-Konto Berlin". Clicking it opens a popup with the URL specified in `adhocracy.frontend.service_konto_login_url`. If this value is not set, no button is shown.
- Once the login in the popup succeeds, it must redirect to `{frontendUrl}/service_konto_finish?token={token}`. This page is empty, but it hands back the token to adhocracy.
- Once adhocracy receives the token, it closes the popup and attempts to login with the token. The success/error handling is identical to that of regular login.

Note that all of this is is implemented in core. You could argue that it is specific to meinberlin, but the steps described above are actually very generic (except for the term "service konto"). It may need some small adjustments to be truly generic, but they can be done when necessary.

One note on implementation: I first thought it would be a good idea to use `adhCrossWindowMessaging`. It later turned out that this was not a good idea (see 	b1c567b), but some changes to that service are still included because I did not bother to rebase.

**In order to test this**:

open `etc/frontend_development.ini` and set

`adhocracy.frontend.service_konto_login_url = https://skbref.verwalt-berlin.de/skb/FVS/Login/Anmeldung.aspx`

`adhocracy.frontend.service_konto_help_url = https://skbref.verwalt-berlin.de/skb/FVP/FV/BasisHilfe/HilfeIndex.aspx`


src/adhocracy_frontend/adhocracy_frontend/static/ServiceKontoFinish.html
```diff
@@ -0,0 +1,23 @@
+<script>
+var parseSearchString = function(s) {
+    var q = {};
+    var items = s.substring(1).split('&');
```
der-john: what is `substring(1)` good for?

der-john:
looks good to me thus far. untested.

src/adhocracy_frontend/adhocracy_frontend/static/ServiceKontoFinish.html
xi: remove the `#`

2e2a:
Reviewed and works on staging.

PR #2844, opened 2016-11-10
improve user settings page

der-john:
While working on #2827, I noticed the user settings UI was getting unorderly. This is still not a perfect UI, but I think it's a (cheap) improvement.


xi:
I am not convinced that this actually improves usability. Giving each setting its own form seems like overkill to me.

I talked to our designer and he said that the biggest issues are wording ("password"/"old password") and the fact that you can not see the current email. I would prefer fixing these two issues instead of overhauling the whole UI.

der-john:
> the biggest issues are wording ("password"/"old password") and the fact that you can not see the current email. I would prefer fixing these two issues instead of overhauling the whole UI.

I'll take care of it.

PR #2849, opened 2016-11-16
Fix S1 decision date handling

### LINKED ISSUE #2786: Inconsistent workflow assignment state_data returned
On the paas staging server the backend returns different values for the IWorkflowAssignment state_date field of a process ([https://aastendal-stage.partou.org/r/aastendal/thementag-21-09-16/](https://aastendal-stage.partou.org/r/aastendal/thementag-21-09-16/)).

When requesting a process directly on the server (bypassing cache) the backend returns different state_date values, sometimes empty, sometimes the current value. During previous tests it seemed also to return an old value.

Also reloading the sheet view  often in SDI shows the problem:
[https://aastendal-stage.partou.org/api/manage/aastendal/thementag-21-09-16/@@properties/IWorkflowAssignment](https://aastendal-stage.partou.org/api/manage/aastendal/thementag-21-09-16/@@properties/IWorkflowAssignment)

@joka Any idea what could cause this?

This has high prio, as it breaks the decision_date mapping.

2e2a:
Fixes #2786 

Note that this fix overwrites any other workflow state-data, e.g. set via SDI. 
This does not affect the current installations, but in case the workflow state-data should be used for other purposes in the future the TODO introduced in this fix must be implemented.

### DIFF
--- a/src/adhocracy_s1/adhocracy_s1/workflows/s1.py
+++ b/src/adhocracy_s1/adhocracy_s1/workflows/s1.py
@@ -29,7 +29,7 @@ def do_transition_to_voteable(context: IPool, request: Request, **kwargs):
 def do_transition_to_result(context: IPool, request: Request, **kwargs):
     """Do various tasks to complete transition to result state."""
     decision_date = datetime.utcnow().replace(tzinfo=UTC)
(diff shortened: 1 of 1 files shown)


der-john:
lgtm.

PR #2850, opened 2016-11-16
die ideen sind frei

der-john:
Continues the conceptual refactoring of #2834:

- `adhIdeaCollectionProcess` gets integrated into `adhProcess`,
- `adhIdeaCollectionPoll`, -`Proposal`, and -`Workbench` become `adhPoll`, `adhProposal` and `adhWorkbench`,
- all directives are renamed accordingly.


src/meinberlin_lib/meinberlin_lib/static/js/Packages/Meinberlin/Alexanderplatz/Workbench/Module.ts
```diff
@@ -30,7 +29,6 @@ export var register = (angular) => {
         .module(moduleName, [
             AdhDocumentModule.moduleName,
             AdhHttpModule.moduleName,
-            AdhIdeaCollectionModule.moduleName,
```
xi: The `AdhProposal` dependency is missing

xi:
LGTM

src/meinberlin_lib/meinberlin_lib/static/js/Packages/Meinberlin/Alexanderplatz/Workbench/Module.ts
xi: Maybe it is not needed after all. Can you explain why you removed it?

PR #2853, opened 2016-11-17
Fix wheel is missing json files 

joka:
That means the fixtures to import data are not available.


joka:
Its a small fix that does not touch programming code. I merge this without further review.

PR #2855, opened 2016-11-21
Fix more icons

xi:
this continues #2773 by fixing two more icons. We got reports that these 2 are especially annoying.


der-john:
looks good to me & works.

PR #2856, opened 2016-11-22
treat all links in markdown as external

xi:
This solves two issues:

- angular seems hijack the routing to some static files if they are under the same domain
- links "oben" in iframe if embedded

I think it is fair to assume that all links in markdown are external. Even if they link to adhocracy content it should be treated as reference rather than navigation.


der-john:
the code looks good to me. but when I test this, the following happens:

1. the markdown produces the following html:
````<a href="www.google.com" rel="external" target="_blank">href</a>```
2. but the browser actually links to `localhost:6551/www.google.com`.

Am I missing something?

xi:
> Am I missing something?

Yes. `www.google.com` is not a vaild URL. Try `http://www.google.com`

der-john:
Looks good & works!

PR #2859, opened 2016-11-22
Fix test coverage

2e2a:


xi:
Coveralls reports that coverage remained the same. I guess it only reports line coverage and this adds coverage for a branch.

PR #2860, opened 2016-11-23
two quickfixes for collaborative text editing

der-john:
#2842 broke a property we weren't conscious of: documents were previously presented alphabetically. This PR restores the old order this way: the process detail view controller just asks if the items to be displayed are documents. If so, it sorts them by title. Otherwise, the usual facets are generated.

Besides that, the image thumbnail is eliminated from the list item.


xi:
Why did you remove `hasImage` from *all* processes?

PR #2865, opened 2016-11-28
Setting to enable/disable email notifications for followed resources

2e2a:


src/adhocracy_core/docs/notification.rst
```diff
@@ -43,8 +43,8 @@ Following
 On default you don`t follow any resource:
 
     >>> resp = app_participant.get(participant_path).json
-    >>> resp['data'][INotification.__identifier__]
-    {'follow_resources': []}
+    >>> resp['data'][INotification.__identifier__].get('follow_resources')
+    []
```
der-john: Is this related? If so, how?

src/adhocracy_core/adhocracy_core/sheets/notification.py
```diff
@@ -37,6 +39,7 @@ class NotificationSchema(MappingSchema):
 
     follow_resources = UniqueReferences(reftype=NotificationFollowReference,
                                         choices_getter=get_follow_choices)
+    email_notification_enabled = Boolean(default=True, missing=drop)
```
2e2a: missing=drop means that if the value is None its not set at all. This allows setting just the required fields of a sheet.

src/adhocracy_core/docs/notification.rst
2e2a: As data now has "follow_resources" and "email_notification_enable" this emphasizes the point above "On default you don`t follow any resource".

der-john: aaaah now I get it.

PR #2858, opened 2016-11-22
improve user settings page II

der-john:
remix of #2844.

TODO:
- [ ] show current email address

I ran into difficulties while working on #2844 already: If I just pre-fill the current email address and send it in the form, I get an error saying "email address is not unique". I tried not sending it, but I didn't succeed yet.

PR #2857, opened 2016-11-22
Service Konto backend

2e2a:
Testing without Frontend:

- in `etc/development.ini`, set `adhocracy.service_konto.enabled = true`.
- Create a Service Konto-Account: [https://skbref.verwalt-berlin.de/skb/FVP/Login/Anmeldung.aspx](https://skbref.verwalt-berlin.de/skb/FVP/Login/Anmeldung.aspx)
- Login to Service-Konto and select mein.berlin Sevice, then "weiter". You will be redirected to mein.berlin.de with a token as a parameter.
- Extract the Token from the network tab of the browser. 
- Login with curl: `curl http://localhost:6541/login_service_konto -H 'Content-Type: application/json;charset=UTF-8' --data-binary '{"token":"TOKEN"}`
- Note that the token is valid only for 20 seconds

PR #2867, opened 2016-11-29
user settings III

der-john:
This PR achieves two things:

1. follow-up of #2858: display current email address to users, so they know they what it is and that they may change it,
2. addition to #2843 / #2857: don't display email address field to Servicekonto users - it's not available,
3. continuation of #2865: if there are "followable" resources, allow the user to enable and disable email notifications for these resources.


src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/User/Edit.html
```diff
@@ -25,7 +25,7 @@
             {{ "TR__ERROR_REQUIRED" | translate }}
         </span>
     </label>
-    <label>
+    <label data-ng-if="data.email">
```
xi: Servicekonto users **do** have an email address

PR #2868, opened 2016-11-29
increase padding to leave room for osx scrollbars

xi:
OSX has the strange habit to draw scrollbars on top of content, blocking its use. To work around that, I increased the general padding.

image


der-john:
looks good & works on OSX.

PR #2883, opened 2017-02-20
Render proposal detail as markdown

xi:
This does not really break old content but allows to have much richer
proposals in the future.


2e2a:
lgtm

PR #2889, opened 2018-03-16
Fix unicode import from json

dwt:
We found that importing json files with umlauts explodes - and this is why.


nidico:
Looks good and works.

PR #2890, opened 2018-03-16
Fix build

dwt:
* Disable rubygems & hologram because of general brokenness
* Update usage of collections.namedtuple to prepare for python 3.6
* Generell Build fixing by switching over to installing and running buildout from the virtualenv, as bootstrap.py has problems with current setuptools versions, which require to be installed by pip.
* Remove bootstrap.py as it can and should no longer be used.


nidico:
Looks good and fixes the S1 development build for me as well. Also, Travis is passing again, yay!

PR #2902, opened 2019-02-26
Use cryptacular from git to make it work with Python 3.6

nidico:
This works around
https://bitbucket.org/dholth/cryptacular/issues/11/not-installing-on-ubuntu-1804


joka:
This pull request adds another source code checkout  without a real need as long as we dont update to python 3.6. So I would wait until this is fixed in the official cryptacular packages.

nidico:
> So I would wait until this is fixed in the official cryptacular packages.

It doesn't look like this would happen anytime soon - the last cryptacular release was from 2012.

> as we dont update to python 3.6

My development machine doesn't have Python 3.5 and our buildout.python seems to be broken, so I tried to go the Python 3.6 way, but of course we can fix our buildout.python instead...

Let's keep this open in the meantime.

PR #2904, opened 2019-02-26
fix regression: broken blob support for S1 package

joka:
The blob storage path for the ZEO client (pyramid app) was set to
var/blobs, the path for the zeo server instead to
var/db/<package>/blobs.

This was causing file not found errors.

Now both path are pointing to var/db/<package>/blobs for the s1 package.
The other packages still need to be fixed.


nidico:
works for me!

